PR #2, opened 2019-05-08
Go structs for Istio 1.2 values.yaml schema

ostromart:
Lean and mean go structs mirroring the istio.io online docs. 
These will be auto-annotated when official documentation is updated. 
Unit tests to follow after these are reviewed for accuracy.

Note: I've refactored this to 1.2 schema.


pkg/apis/istio/v1alpha1/values_types.go
```diff
@@ -0,0 +1,506 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+type Values struct {
+	CertManager     *CertManagerConfig     `json:"certmanager,omitempty"`
+	Galley          *GalleyConfig          `json:"galley,omitempty"`
+	Global          *GlobalConfig          `json:"global,omitempty"`
+	Grafana         map[string]interface{} `json:"grafana,omitempty"`
+	Gateways        *GatewaysConfig        `json:"gateways,omitempty"`
```
f4tq: Shouldn't `gateways` be an slice?  
Or would the intention for those using multiple sets of gateways to create a secondary crd/controller that watches IstioControlPlane.
I'm specifically thinking of the multi-tenant with single control plane use case.

rcernich: agreed, and i think it's a bit more complicated than that.  i'll post a link to work i've done previously to create a go type for values.yaml.

rcernich:
Hey @ostromart, here's a link to the work I had previously done in this area:

https://github.com/Maistra/istio-operator/blob/maistra-0.11/pkg/apis/istio/v1alpha3/controlplane_types.go#L58-L76

Some things of note:

* gateways doesn't map well, so I created a [type](https://github.com/Maistra/istio-operator/blob/maistra-0.11/pkg/apis/istio/v1alpha3/controlplane_types.go#L448-L457) with a specialized [serializer](https://github.com/Maistra/istio-operator/blob/maistra-0.11/pkg/apis/istio/v1alpha3/serializer.go).  From what I could tell, gateways was a general type, with things like ingress, egress, ilb, etc., being predefined, but users could add their own gateways (e.g. see the gateways example).
* I used embedded types for common settings, e.g. most deployments have fields for autoscaling, annotations, env, etc., defined [here](https://github.com/Maistra/istio-operator/blob/maistra-0.11/pkg/apis/istio/v1alpha3/controlplane_types.go#L925-L961) and referenced, e.g.,  [here](https://github.com/Maistra/istio-operator/blob/maistra-0.11/pkg/apis/istio/v1alpha3/controlplane_types.go#L615-L631).

f4tq:
It seems like the GatewayConfigs could be used (as is) to support a separate 'user' centered istio gateway CRD that gets `Watch`ed/`Reconcile`d via the forthcoming controller.  It would need a selector, of course, to support multi-tenancy so it can be applied to the correct IstioControlPlane (in the presence of multiple).

That way, an existing istio installation could incrementally change to add/remove istio gateways needed for user groups clustered in a namespace.   This way we would not need to regenerate, as we do now, the complete helm values for the entire installation and applying to the whole cluster.

Since we use many ingress/egress gw sets now - one per namespace (which come and go) -  it has been challenging to have to touch the monolithic helm values/chart on the existing install just to add a new set of gateways.  

It would be much easier & more granular if there were a CRD representing 'user' istio gateways to support the multi-tenant/shared control plane use case.  Then, when a new pair are needed, just the gw/certs (and rbac/roles for it) would need to be `apply`ed by the admin configuring a tenant/namespace. 

The istio gateway created as described in this PR so far could be dedicated to grpc endpoints (istio concerns) - not user.  In our case, the istio gateways we create all have separate wildcard certs, endpoints and only expose port 443 i.e. none of the grpc.

FWIW, compliance issues can drive the need for separate istio-control planes, multiple control planes  in multiple k8 clusters (api-server access etc) & multiple 'user' istio gateways.  All could conquered with a single stone it seems.

Please consider.

pkg/apis/istio/v1alpha1/values_types.go
ostromart: This is a struct version of:
https://istio.io/docs/reference/config/installation-options/#gateways-options
The intended purpose is to be able to unmarshal a values.yaml into a struct so it can be validated and passed through to helm charts. 
It's not intended to be a new schema for an installer, that's here: https://github.com/istio/operator/pull/3

ostromart:
A couple of comments:
1. We don't need perfect fidelity in these structures. The purpose here is to be able to unmarshal values.yaml into Go structs for further validation. As such, we can leave any fields that are too difficult to define as interface and let the validation deal with them as raw maps if required. 
2. This is indeed a spec for the istio system, not user installed gateways. The proposal for installing multiple user GWs is in https://github.com/istio/operator/pull/3, which does use a slice - PTAL.

pkg/apis/istio/v1alpha1/values_types.go
rcernich: @ostromart, my point was that the schema for gateways is a map, gateway-name->config, and that the gateways you listed are "standard" gateways typically installed with the control plane, but other gateways can be installed in the same manner, e.g. https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/example-values/values-istio-gateways.yaml#L36

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
```diff
@@ -1,9 +1,23 @@
+// Copyright 2019 Istio Authors
```
golangcibot: vendor/k8s.io/client-go/rest/request.go:598:91: too few arguments in call to watch.NewStreamWatcher (from `typecheck`)

pkg/controller/add_istiocontrolplane.go
```diff
@@ -1,7 +1,21 @@
+// Copyright 2019 Istio Authors
```
golangcibot: vendor/k8s.io/client-go/rest/request.go:598:91: too few arguments in call to watch.NewStreamWatcher (from `typecheck`)

cmd/manager/main.go
```diff
@@ -1,123 +1,27 @@
+// Copyright 2019 Istio Authors
```
golangcibot: vendor/k8s.io/client-go/rest/request.go:598:91: too few arguments in call to watch.NewStreamWatcher (from `typecheck`)

pkg/values/values_test.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func readFile(path string) (string, error) {
+	b, err := ioutil.ReadFile(path)
+	return string(b), err
+}
+
+// errToString returns the string representation of err and the empty string if
+// err is nil.
+func errToString(err error) string {
```
golangcibot: `errToString` is unused (from `deadcode`)

pkg/apis/istio/v1alpha1/values_types.go
```diff
@@ -0,0 +1,616 @@
...
+	CPU                              *CPUTargetUtilizationConfig `json:"cpu,omitempty"`
+	CustomService                    *bool                       `json:"customService,omitempty"`
+	Debug                            *string                     `json:"debug,omitempty"`
+	Domain                           *string                     `json:"domain,omitempty"`
+	DrainDuration                    *string                     `json:"drainDuration,omitempty"`
+	Enabled                          *bool                       `json:"enabled,omitempty"`
+	Env                              map[string]string           `json:"env,omitempty"`
+	ExternalIPs                      []string                    `json:"externalIPs,omitempty"`
+	K8sIngress                       *bool                       `json:"k8sIngress,omitempty"`
+	K8sIngressHttps                  *bool                       `json:"k8sIngressHttps,omitempty"`
```
golangcibot: struct field `K8sIngressHttps` should be `K8sIngressHTTPS` (from `golint`)

pkg/apis/istio/v1alpha1/values_types.go
```diff
@@ -0,0 +1,616 @@
...
+// PrometheusMixerAdapterConfig is described in istio.io documentation.
+type PrometheusMixerAdapterConfig struct {
+	Enabled               *bool   `json:"enabled,omitempty"`
+	MetricsExpiryDuration *string `json:"metricsExpiryDuration,omitempty"`
+}
+
+// StdioMixerAdapterConfig is described in istio.io documentation.
+type StdioMixerAdapterConfig struct {
+	Enabled      *bool `json:"enabled,omitempty"`
+	OutputAsJson *bool `json:"outputAsJson,omitempty"`
```
golangcibot: struct field `OutputAsJson` should be `OutputAsJSON` (from `golint`)

ostromart:
Unit tests have been added and are passing for inputs taken from istio/installer chart values.

pkg/apis/istio/v1alpha1/values_types.go
rcernich: Are you saying you updated it to make it generic, if so,`GatewaysConfig` is still a hard-coded list of gateways, which supports some of the known gateways, but is not congruent with the general `values.yaml`.  If you're saying this is something that still needs to be addressed, I'm okay with that.  Let me know.

ostromart: Created https://github.com/istio/istio/issues/14592

PR #3, opened 2019-05-08
Installer CRD proto definition

ostromart:
Initial proto definition which is basically just a wrapper around values.yaml schematized in https://github.com/istio/operator/pull/2.
The only exceptions are:
- kubernetes resources are structured and defined in terms of k8s official protos
- component enablement and namespaces are structured by feature and component, per Istio building blocks and new installer charts.


smawson:
Quick question, why is this going into 1.1 release branch and not master?

ostromart:
Values.yaml v1.1 definitions are stable so first bunch of code going in will be against istio 1.1. I'll cherry pick many of these back into master.

linsun:
Sorry but what is the purpose of having operator release-1.1 branch?   I had thought it is for master and 1.2 only. :)

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,441 @@
...
+}
+
+// InstallerSpec defines the desired state of IstioInstaller.
+message InstallerSpec {
+    // Prefix added to all namespaces for any installed component.
+    string default_namespace_prefix = 5;
+
+    // Selection and configuration of core Istio features.
+    TrafficManagementFeatureSpec traffic_management = 25;
+    PolicyTelemetryFeatureSpec policy_telemetry = 26;
```
linsun: mentioned in the design doc... i prefer to separate the two given it is totally two different features for our users.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,441 @@
...
+    // Selects whether traffic management is installed.
+    // Must be enabled to enable any sub-component.
+    google.protobuf.BoolValue enabled = 1;
+    // Namespace that all traffic management components are installed into.
+    string namespace = 2;
+
+    // Component specific config.
+    PilotComponentSpec pilot = 30;
+    ProxyComponentSpec proxy = 31;
+    SidecarInjectorComponentSpec sidecar_injector = 32;
```
linsun: side car injector isn't tied directly to traffic mgmt feature as you need side car injector for telemetry or mTLS without traffic mgmt.  Is this the right place?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,441 @@
...
+}
+
+// Configuration options for the policy and telemetry feature.
+message PolicyTelemetryFeatureSpec {
+    // Default behavior of the sidecar for handling outbound traffic from the application.
+    enum OutboundPolicy {
```
linsun: why outbound policy for the mesh is configured under policy/telemetry feature spec?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,441 @@
...
+    // Namespace that security components are installed into.
+    string namespace = 2;
+
+    // Enables MTLS for control plane.
+    // Will result in delays starting the pods while secrets are propagated, not recommended for tests.
+    google.protobuf.BoolValue control_plane_mtls = 13;
+
+    // Enables MTLS for service to service traffic.
+    google.protobuf.BoolValue data_plane_mtls = 14;
+
```
linsun: what about mTLS strict?  i feel that is the most important config for security.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,441 @@
...
+    CitadelComponentSpec citadel = 20;
+    CertManagerComponentSpec cert_manager = 21;
+    NodeAgentComponentSpec node_agent = 22;
```
linsun: having these components spec here and the need for user to understand what each component is before they can configure things is hard to use imho and not much difference than what we offer from helm values today.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,441 @@
...
+// InstallerSpec defines the desired state of IstioInstaller.
+message InstallerSpec {
+    // Prefix added to all namespaces for any installed component.
+    string default_namespace_prefix = 5;
+
+    // Selection and configuration of core Istio features.
+    TrafficManagementFeatureSpec traffic_management = 25;
+    PolicyTelemetryFeatureSpec policy_telemetry = 26;
+    SecurityFeatureSpec security = 28;
+    ConfigManagementFeatureSpec config_management = 29;
```
linsun: is this something we need to expose to users?  if galley is required for everything else to run and operate, what is the need to configure this?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,441 @@
...
+message PilotComponentSpec {
+    CommonComponentSpec common = 1;
+
+    // Enables sidecar in the pilot pod.
+    google.protobuf.BoolValue sidecar = 10;
+}
+
+// Configuration options for the proxy.
+message ProxyComponentSpec {
+    CommonComponentSpec common = 1;
```
linsun: what value does this provide on top of side car injector spec?  can't user configure proxy as part of sidecar injector config?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,441 @@
...
+    // k8s ImagePullPolicy setting for all containers.
+    // https://kubernetes.io/docs/concepts/containers/images/#updating-images
+
+    // k8s node affinity.
+    // https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
+    message NodeAffinity {
+        uint32 arch_amd64 = 1;
+        uint32 arch_s390x = 2;
+        uint32 arch_ppc64le = 3;
+    }
```
linsun: we also have rolling update strategy and pod affinity.

linsun: also, cpu/mem limit and requested values

ostromart:
This is 1.1 because we're starting with values.yaml v1.1 "schema". I'll cherry pick this and other PRs into master once v1.1 installation is mostly working.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: The issue is that k8s resources and overlays are necessarily a component rather than feature concept. keep in mind that these are overrides over defaults - most users won't need to worry about this level of detail. 
The initial API did in fact migrate out all the values.yaml configuration into a structured API, but the review feedback  was that the Istio config part was going into component CRDs and therefore shouldn't be a part of this installation API. 
Se the idea here is to keep values.yaml but provide an alternative API just for that parts that will not go into the future CRDs i.e. k8s level stuff.
Having said that, we may want to revisit this path and consider only using values.yaml only under the hood rather than exposing directly, and go back to the initial proposal for a complete API replacement of values.yaml.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Sidecar component config deals with just sidecar resources i.e. the part that's doing the injecting, not proxy.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: afaik galley is optional. This is an override API so there's no requirement to set any of these components if the defaults work.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: I'm ok with splitting them up but I'd like to hear from @smawson as this is something that came out of early reviews.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: It seemed like the least worst place to put it, as it clearly doesn't belong with any other feature. We could create a feature specifically for the injector I suppose.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: So firstly I have to contradict myself a bit about an earlier statement that all non-k8s Istio config would still go through values.yaml. I think even in that case there may be some very coarse "feature selection" knobs that may be good to expose, even though the user could set that directly through values.yaml. These are the sort of knobs that 80% of installs may consider changing and this seems like a good match for that. 
As to why it's PolicyTelemetry rather than just Policy, if these are split again it would go into Policy feature.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: That's exactly what this is. The documentation is misleading here and in the values.yaml description, I'll update it.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: requests/limits are already under Resources above. 
Added k8s Affinity which takes care of both pod and node [anti-]affinity.

ostromart:
After more discussion, istio-installer is probably stable enough to start using now. In that case, going straight to master and we'll start working against 1.2 immediately.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Checked with Sven, he's ok with it, done.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: @ostromart, I agree with @linsun, if we're promoting a biased install (which I think we should, 80/20 rule), then mcp should be default, which means using galley.  That said, there should be a place to allow for whatever customization might be included in the 80%.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,461 @@
...
+    SecurityFeatureSpec security = 28;
+    ConfigManagementFeatureSpec config_management = 29;
+
+    // Ingress/egress gateway installation and configuration.
+    repeated IngressGatewaySpec ingress_gateway = 30;
+    repeated EgressGatewaySpec egress_gateway = 31;
+
+    // List of external operators that will be installed (but not managed)
+    // by the IstioInstaller e.g. prometheus, grafana, kiali.
+    repeated OperatorSpec external_operators = 50;
```
rcernich: I think this is a slippery slope.  I don't think we should be doing anything with external operators, but should instead have some validation that rejects CRs if they would require a CRD supported by an external operator that hasn't been installed.  Using something like OLM would be a better option here.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: Should this be part of the security spec?  Should it apply globally, or is it intended that certain components would override global defaults?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: I think it would be better if we could separate the functional details from resourcing requirements.  While that might be a bit messier from a configuration structure, I think separation is beneficial.  For the work we've been doing with maistra, we've started to use a simple config for features, with a small defaultResources setting, which allows for a better ootb experience on developer machines (things like grafana can be resource hogs).

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,461 @@
...
+
+    CitadelComponentSpec citadel = 20;
+    CertManagerComponentSpec cert_manager = 21;
+    NodeAgentComponentSpec node_agent = 22;
+}
+
+// Configuration options for configuration management feature.
+message ConfigManagementFeatureSpec {
+    // Selects whether security feature is installed. Must be set for any sub-component to be installed.
+    google.protobuf.BoolValue enabled = 1;
```
rcernich: see comments above, i think this should always be enabled for any install.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: is there a reason to separate those concerns?  maybe sidecar injector should be part of proxy config?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,461 @@
...
+    VersionStatus policy_telemetry = 2;
+    VersionStatus security = 3;
+    VersionStatus config_management = 4;
+
+    repeated VersionStatus ingress_gateway = 5;
+    repeated VersionStatus egress_gateway = 6;
+}
+
+// Mirrors k8s.io.api.core.v1.ResourceRequirements for unmarshaling.
+message Resources {
```
rcernich: if it mirrors it, would it not be better to simply embed the upstream type directly?  shallower learning curve if they're the same.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,461 @@
...
+    string externalTrafficPolicy = 11;
+}
+
+// OperatorConfig defines a config used to install an external operator.
+message OperatorSpec {
+    // manifest_path is the URL for the operator install manifest.
+    string manifest_path = 1;
+
+    // namespace is the namespace the operator manifest and CR are installed into.
+    string namespace = 2;
```
rcernich: isn't it more natural to install in the same namespace as the CR?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,461 @@
...
+
+    // spec is the initial CR for the operator.
+    k8sObjectOverlay spec = 100;
+}
+
+// Patch for an existing k8s resource.
+message k8sObjectOverlay {
```
rcernich: This seems like it could really bloat the CR.

Is there a benefit to using an overlay vs customized templates?  If the idea is to save users from modifying the templates, is there a way the overlays could be added to the templates and applied automatically if they exist (e.g. maybe an overlays folder alongside the templates folder in the chart directory).

I'm still having a hard time understanding the value proposition here.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Policy is related to security but is its own thing in Istio and traffic policy is a part of that. It's global in the sense that it applies to all sidecars in the mesh and I don't think it makes sense to override this in the installer at least. Istio does provide config knobs to shape policy with finer granularity.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Resourcing must be tied to components so I'm interpreting the comment as suggesting splitting the components from the features. The reason we probably don't want to do that is that components are logically grouped so they can easily share a namespace (and enablement) and all the policy control associated with the feature/namespace. We want to be able to control all that at feature level, with the ability to override per component.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: This schema is both for the base spec and override spec. We're free to put whatever defaults we want into the base spec and assign base specs to profiles. 
Users would use the same schema to optionally override the base spec values. We can (and do) make galley/MCP on by default in most profiles so unless user explicitly turns it off it will be there.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Yes, they are completely different components. Proxy config deals with the sidecar proxy on user pods (e.g. how much memory to allocate to proxy). Sidecar injector config deals with the resources for the sidecar so the resource settings would apply to the sidecar injector pods.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: What is OLM? This is just an install passthrough, the use case is to just install an external operator (e.g. prometheus for demo) and defer control of it to that sub-operator. There's no interaction with the Istio operator, it's just a convenience so we don't keep maintaining add on config in Istio for demo purposes.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: It would be enabled in most of the base spec profiles.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: The problem is that k8s protos have their own unmarshaler which doesn't extend to just plugging in these types into it. It's copied here so that standard jsonpb unmarshaler works and k8s APIs are preserved.
I was able to link to k8s definitions in some cases but not all due to weirdness in unmarshaling.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: As which CR? This is for both the operator and its CR so I think this is already doing what you're suggesting?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Yes, we want to avoid customizing the templates. I don't follow what you mean by adding overlays to the templates - it seems pretty complex and I think overlays are a user customization so it's easier to apply them to the rendered output than the templates. We as developers will continue to work with the templates, but hopefully less since this mechanism allows users to do customizations without asking us to add parameters to the templates.
Take a look at the overlay example I provided in the header - we're looking for a really simple mechanism to override values that previously required template parameter additions.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
morvencao: So the external operator spec here is just to keep track of sub-operators, Istio operator won't have logic to handle CR of sub-operators.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,585 @@
...
+    repeated OperatorSpec external_operators = 50;
+
+    // Global k8s defaults. Can be overridden per component.
+    KubernetesResourcesSpec k8s_defaults = 80;
+
+    // URL for the base spec. e.g.
+    //   istio-auth.yaml (looks in default dir)
+    //   file://var/istio/install/values/custom/custom-install.yaml
+    // Uses compiled in default install spec by default.
+    string base_spec_path = 100;
```
morvencao: I thought the default spec should be in install package, while the custom spec is generated by the other fields.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,585 @@
...
+}
+
+// Configuration options for the policy enforcement component.
+message PolicyComponentSpec {
+    CommonComponentSpec common = 1;
+}
+
+// Configuration options for the telemetry component.
+message TelemetryComponentSpec {
+    CommonComponentSpec common = 1;
```
morvencao: I'm not sure if we need `loadshedding` in `TelemetryComponentSpec` , it's a telemetry specific parameter.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,585 @@
...
+    google.protobuf.BoolValue data_plane_mtls_strict = 14;
+
+    CitadelComponentSpec citadel = 20;
+    CertManagerComponentSpec cert_manager = 21;
+    NodeAgentComponentSpec node_agent = 22;
+}
+
+// Configuration options for configuration management feature.
+message ConfigManagementFeatureSpec {
+    // Selects whether security feature is installed. Must be set for any sub-component to be installed.
```
morvencao: Nit: should be `config management`, instead of `security`

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
knrc: Operators should not be tracking other operators, that is the job of other tools such as the Operator Lifecycle Manager.  The operator should only be concerned with Custom Resources monitored by any dependent operators.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,585 @@
...
+import "github.com/gogo/protobuf/protobuf/google/protobuf/wrappers.proto";
+
+/*
+IstioInstall is a schema for both defining and overlaying Istio control plane installations.
+The Istio operator uses a compiled-in InstallSpec as default installation values, along with compiled-in install charts.
+Running the operator with an empty user defined InstallSpec results in an installation with default values, using the
+default charts.
+
+The simplest install specialization is to point the user InstallSpec baseValuesPath to a different values file, for
+example an Istio auth installation, which will use the values associated with the auth installation profile for Istio.
```
linsun: Can I assume I can use a service profile here?  Nit to help user easier understand this - I think it makes sense to refer to one of our supported profile like mTLS strictly enforced profile or minimal profile.

linsun: Reading it closer, I think that is your intention, it just was not aligned well with our terminologies:  https://istio.io/docs/setup/kubernetes/additional-setup/config-profiles/

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,585 @@
...
+
+k8s API: this API is a pass through to k8s resource settings for Istio k8s resources. It allows customizing Istio k8s
+resources like Affinity, Resource requests/limits, PodDisruptionBudgetSpec, Selectors etc. in a more consistent and
+k8s specific way compared to values.yaml. See KubernetesResourcesSpec in this file for details.
+
+2. values.yaml
+
+The entirety of values.yaml settings is accessible through InstallSpec. They are included for backward compatibility,
+but also because many of the values will be moved to configuration CRs that are consumed directly by Istio components
+rather than being install time parameters. The only remaining fields will be k8s resource settings which will be
```
linsun: This last sentence is confusing.... I think it is confusing to make "will" statement in the future.  It is not clear why the k8s resource settings will be remaining fields in values.yaml if they are moved to k8s resources API.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,585 @@
...
+EXAMPLES
+
+1. Default Istio install
+
+spec: ~
+
+2. Default remote profile install
+
+spec:
+  baseSpecPath: istio-remote.yaml
```
linsun: you mean values-istio-remote.yaml here?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,585 @@
...
+  config_management:
+    namespace: istio-config-management
+  security:
+    citadel:
+      namespace: istio-citadel
+    cert_manager:
+      namespace: istio-cert-manager
+    node_agent:
+      namespace: istio-node-agent
+
```
linsun: This is nice!

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,585 @@
...
+      - global.proxy.enableCoreDump: true
+      - global.proxy.dnsRefreshRate: 10s
+
+7. Default install with modification to container flag in galley
+
+spec:
+  configuration_management:
+    galley:
+      k8s:
+        overlays:
```
linsun: why overlays here?

linsun: is it for overwrite?

linsun: so is the intention here you can use valuesOverrides to override values that can be configured via helm value, then use yaml style overrides to overwrite yaml blobs?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,585 @@
...
+// InstallSpec defines the desired state of IstioInstaller.
+// The spec is a used to define a customization of the default values that are supplied with an Istio charts release.
+// It is grouped at the top level by feature, where behavior of Istio functional areas is specified.
+// Each feature contains components, where k8s resource level defaults can be overridden.
+// Because the spec is a customization API, specifying an empty InstallSpec results in a default Istio installation.
+message InstallSpec {
+    // Prefix added to all namespaces for any installed component.
+    string default_namespace_prefix = 5;
+
+    // Selection and configuration of core Istio features.
```
linsun: Perhaps add: if you don't specify any feature, the default will be used.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: we may want to leave out external_operator since it is confusing why user needs to specify istio operator isn't managing it.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: I have the same q.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: I think we should move sidecar injector out of traffic management as it is not directly related.  You can use side car injector when you are not using traffic management.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: I think outbound policy should be in traffic management feature, after all configuring outbound policy is for traffic management, unless this outbound policy is specific to the mixer policy pod/service.

costinm:
Afaik the current proposal is to dedup mesh config and values.yaml - so
this should go inside mesh config.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,585 @@
...
+        // Outbound traffic to unknown destinations will be allowed, in case there are no
+        // services or ServiceEntries for the destination port
+        ALLOW_ANY = 0;
+        // Restrict outbound traffic to services defined in the service registry as well
+        // as those defined through ServiceEntries
+        REGISTRY_ONLY = 1;
+    }
+
+    // If set, allows traffic in cases when the mixer policy service cannot be reached.
+    google.protobuf.BoolValue policy_check_fail_open = 10;
```
linsun: can user also configure disable policy check? that seems the most important config for mixer policy.

linsun: for example, in 1.1, we have mixer policy enabled by default but disable policy check enabled by default.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: Thx for the update, makes sense now.  could we use consistent naming? control_plane_mtls -> control_plane_mtls_strict or change data_plane_mtls_strict to data_plane_mtls?   IMHO, strict is our impl detail, people expect mtls to be mtls enforced.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: @rcernich do you have some example yamls for separating the functional details from resourcing requirements?  I think that will help unclutter things and not exposing our impl details to users who only care about function or feature configuration.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,585 @@
...
+
+    // Namespace that security components are installed into.
+    string namespace = 2;
+
+    GalleyComponentSpec galley = 20;
+}
+
+// Configuration options for the pilot component.
+message PilotComponentSpec {
+    CommonComponentSpec common = 1;
```
linsun: why is only pilot using CommonComponentSpec?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: Okay, my comment was directed toward the organization of security related concerns.  If users are more familiar with configuring this in the context of mixer policy, then having it here is probably best.  Given that, would it make more sense to change "security" to certificate management?  Other than certs, it seems security related features are configured in the context of policy.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: I actually don't and based on @ostromart's last comment, maybe what I was thinking is less useful.  I do think it's beneficial to segregate resourcing configuration from feature configuration, but maybe that can be done by simply having something like "ResourceConfig" as part of each section.  The only details within that section would be things like resource limits, node selectors, autoscaling details, etc.

A generic mock up:
```yaml
componentA:
  featureA:
    ...
  operationalConfig:
    resourceRequirements:  # k8s ResourceRequirements
    podDisruptionBudget:  # k8s PodDisruptionBudget
    autoScaling: # things like min/max replicas, target cpu utilization, etc.
    nodeSelector: # map of label->value

operationalConfig: # same as above, but global defaults, applied when no component.operationalConfig is specified.
```

The above is a crude example and assuming one deployment resource per component, but hopefully, it conveys the general idea.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: I guess the question is: should we allow users to disable mcp in a k8s installation?  Unless the answer is, yes, most users disable it, then I think the only thing we need this for is to allow users to set "operationalConfig" for galley.  If they need to disable mcp/galley, then they use the "harder" mechanism of overriding it via raw values.yaml.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: Sorry, what I was implying was that configuration of proxy would include enabling/disabling and configuration of sidecar injector as well, since sidecar injector injects a proxy into user pods.  This would allow all proxy related configuration to be in a single place.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: I think for simplicity, we should just focus on managing the installation of the istio control plane.  Longer term, I suspect downstream users/adopters might create custom operators which manage and installation that includes istio, kiali, etc.

If the control plan has a hard dependency on prometheus, then, as @knrc mentioned, we should rely on something else to manage the operator dependencies, and the istio operator should be installing whatever CR is required by that operator.  If no operator management system is being used, it would fall upon the user to ensure that the required operators are installed.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: Could you please explain: this is for both the operator and its CR?  I suspect I'm missing a critical detail in how this is intended to be used.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: My point here is yes, it's traffic management and yes, it's security, but in this case this is policy setting. Policy feature covers aspects of security and traffic management. 
It does happen to be implemented in mixer policy pod/service, but even conceptually I think this belongs in the policy feature group.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: Here's what I meant regarding overlays and templates.

Take the typical helm structure:
```yaml
base-chart:
  templates: # templates for the main chart
  - resource.yaml # a template for some resource, e.g. deployment maybe
  charts: # directory containing subcharts
    subchart-1: # some subchart
      templates:  # templates for subchart-1
      - resource.yaml # a template for some resource
```

Could we modify that so we have something like:
```yaml
base-chart:
  templates: # templates for the main chart
  - resource.yaml # a template for some resource, e.g. deployment maybe
  overlays: # kustomize overlays for main chart
  - overlay.yaml # some overlay to be applied
  charts: # directory containing subcharts
    subchart-1: # some subchart
      templates:  # templates for subchart-1
      - resource.yaml # a template for some resource
      overlays: # kustomize overlays for subchart-1
      - overlay.yaml # some overlay to be applied
```

This would save users from having to describe pathing, etc. for overrides, etc.  Even if the user just specified a base directory that mirrored the above layout for just the overlays, I think that would be easier to manage.

Another alternative would be to have users layout files in a manner that is identical to kustomize, then just point to that directory.  This would allow you to point users to that documentation, saving them from having to learn the "custom" istio way.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: No, because it's not values.yaml format, but values.yaml wrapped in InstallSpec. The operator profile will include values.yaml but also other settings in the new CRD format in this proto file. Changed this to minimal to be more clear.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Thanks Lin. One of the aims with this API is to be super clean and minimal especially for the 80% - if you have any thoughts along those lines, let me know.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: yes, exactly.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: That applies transitively to every field under InstallSpec, because this is an overlay spec. I stated that in InstallSpec comment above, but maybe you have a suggestion how I can emphasize this/make it more clear?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: It can be done through values.yaml, in fact most of these top level API knobs have corresponding fields in values.yaml.
But, we want to give users the new API and encourage them to use it instead of values.yaml, since that API will be deprecated. 
Yes, currently we do want to allow users to disable MCP and not use galley.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Let me emphasize again, Istio operator will not manage the external operator, only bootstrap it into the cluster and leave it alone. I've changed the text to make that more clear. 
The reason we need this is for simple installs of demos that include things like prometheus/grafana etc. 
Istio is getting out of the business of managing those but we still need a "one-line" install that includes those components. We don't want to require users to install OLM to do that, if they just want to kick the tires of Istio.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: I think we should rename security above then, as it's only configuring certificate management.  That said, like galley, is it required to explicitly enable/disable citadel?  I suppose there are configuration details regarding how it integrates with the other components (e.g. SDS/node agents), but it seems like it's used by default for managing certs on the istio components, and additionally when mTLS is enabled, so does it make sense to have it exposed as its own component (other than for configuring how it's manifested in the install)?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: ptal at the examples at the top, it may make it more clear. 
we start the install with a base profile (default profile is the default if base_spec_path is unset). 
any customizations in the InstallSpec here are ovelaid on top of the selected base InstallSpec. 
so, we could start with istio "demo" spec, but then decide we want to customize that by turning off config management:

spec:
  baseSpecPath: istio-demo.yaml
  configManagement:
    enabled: false

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Ok, done.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: that part is a bit confusing for me. how is installing policy but disabling the check different from not installing policy at all and why do we want to support this scenario?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: i think control_plane_mtls_strict would imply this can be permissive, which is not true. it's either off, or strict. 
otoh, for the data plane, this knob controls whether the setting is strict or permissive. 
as such i think the field names are correct in their current form - wdyt?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: thanks, fixed.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: yes, but how then would you specify k8s settings separately for the proxy and sidecar injector? they are both k8s resources and need their own separate resources settings.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: do you mean we should add it? i guess i don't see which field your comment is talking about.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: yes, it's for both. what this will do in practice is install the raw external operator CR (spec below) into the cluster, get the external operator yaml from the URL and apply that into the cluster. both will be applied to the given namespace. 
after that, the operator ignores this struct (does nor reconcile).

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: If we take the approach I mentioned above (using `operationalConfiguration`), you could have a field for each deployment.

What you're talking about, separate deployments, is exactly what I think we should be avoiding with this new specification.  If it truly is intended to describe conceptual features, then we shouldn't be having a mindset that's focused on the resource layout (e.g. deployments) of the concrete installation.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
morvencao: PTAL: https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/charts/mixer/values.yaml#L33-L37

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: We should investigate supporting kustomize because that's that path that the new installer is taking. In all likelihood, we may end up with two ways to overlay - for simple stuff we could embed that in the CRD here to spare the user from creating and editing files, for heavier customization we can support the full blown kustomize framework. It seems to me like a huge burden to place on the user to create files and directory layouts just to override an environment flag (e.g. pod debug level). See https://github.com/istio/operator/pull/8 for some examples.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: How about something like `Selection and configuration of core Istio features over defaults"

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: If istio installs it, do we expect other operators to manage it?  Or would it leave as unmanaged?

It seems this is only used for demo and quick start?  else, folks would use external operator to install the component so it can be managed.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: ok that makes sense, essentially the base_spec_path can point to any profiles as the base.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,596 @@
...
+message TrafficManagementFeatureSpec {
+    // Selects whether traffic management is installed.
+    // Must be enabled to enable any sub-component.
+    google.protobuf.BoolValue enabled = 1;
+    // Namespace that all traffic management components are installed into.
+    string namespace = 2;
+
+    // Component specific config.
+    PilotComponentSpec pilot = 30;
+    ProxyComponentSpec proxy = 31;
```
linsun: is this proxy for pilot or proxy for the data plane?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: I think i'm strongly opposing to have outbound policy within PolicyFeatureSpec, because I could run Istio with only pilot and without mixer policy (or other policy feature) but wanted to configure my traffic outbound policy.  I think that is where people's heads are today, it'd be confusing to move to be some config under PolicyFeatureSpec.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: ok, these are 2 different install configs.  In 1.1, we wanted to not install policy and disable policy check but we ended up shipping disable policy check but having policy installed by default due to backwards compatible with 1.0.  Installing policy maps to have mixer policy deployment and services installed, disable policy check is a config we expose to user when mixer is installed where user can choose to skip all policy checks.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: are you saying control_plane_mtls can be either permissive or enforced mTLS? i am still concerned with the inconsistent naming.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: Thanks @rcernich - I think this starts with component first and list features inside each component, which is component centric.

I 'd prefer to be feature centric so user doesn't have to understand component if he/she doesn't need to.  I had thought you were proposing seperating the component spec out of the feature spec, e.g. something like this:

```
  traffic_management:
    enabled: true
    ...
  components:
    pilot:
        k8s:
          resources:
            limits:
              cpu: 444m
              memory: 333Mi
            requests:
              cpu: 222m
              memory: 111Mi
          readinessProbe:
            failureThreshold: 44
            initialDelaySeconds: 11
            periodSeconds: 22
            successThreshold: 33
```

ostromart: Keep in mind that all of this API is optional - users would not normally dive down to component level, only if they had to (say) modify k8s resources. Defaults are always used if something is not defined and that would be the case most of the time. 
As far as splitting out the components outside of the feature groups, the disadvantage of doing that is that some settings (e.g. install namespace, enabled) are inherited from the feature level and it makes sense imo to maintain that relationship.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: you are right, that is quite confusing but i see that is what is implemented today now.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: ok SG thanks

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
linsun: That is up to the mixer team and backwards compatible when moving from 1.1.

If mixer v2 lands in 1.3, then we won't need this for sure.

linsun:
Thanks @ostromart - I think I'm good with this as the first pass.

I know @rcernich has some concerns in a few places, so I 'll hold off my lgtm in case it got the PR merged.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: @ostromart, looking at those docs you linked, I think it would be beneficial to go through each of those sections and itemize the settings required for each scenario.  Personally, I think it would be great if the installer CRD schema aligned well with these, especially since the docs are organized conceptually around user tasks.  As I mentioned in another comment, for the new CRD schema, I don't think we should be focused on physical manifestation of these features (i.e. simply exposing the options available for each component/deployment), but should instead be focused on _what_ users want, not how what they want gets done.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: @linsun, your example is close to what I was suggesting, except that I would move `components` under `traffic_management`, and possibly call it `operationalConfig`.  @ostromart, regarding namespaces, IMO, those are operational concerns, so I would place those under `components`/`operationalConfig` too.  That said, I don't know what the technical concern is regarding splitting up the control plane across multiple namespaces (i.e. why you'd want to do that).

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: very crude:
```yaml
trafficManagement:
  enabled: true
  proxy:
    autoInjection:
      enabled: true
      enableNamespacesByDefault: true
  operationalConfig:
    pilot:
      resources: {}
    proxy:
      resources: {}
    sidecarInjector:
      resources: {}
```
HTH

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: Why wouldn't you use the namespace where the CR is installed as the default?  I believe this is the _natural_ way to implement the operator.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: I still think this is a bad idea, and saying it's just for demos, I think makes it worse.  If this is for demos, then I don't think it's too big a hurdle to document that somebody doing a demo will have to install other operators if they need/want third-party features.

I think something like this also emphasizes the use case for some sort of built-in extension mechanism or hook that gives users access to setting non-standard or less common configuration details. This might look something like:

```yaml
trafficManagement:
  ...
policy:
  ...
...
rawValues:
  kiali:
    enabled: true
```

This provides users with a concise set of features to configure, while allowing users to bring their own chart sets and still be able to configure them.  In the case above, say kiali is no longer part of the standard chart set, but is included in some third-party distribution, then that user would have the ability to configure those charts, even though there's not direct support in the CRD itself (well, I guess `rawValues` would be direct support).

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: @ostromart, aside from creating a customized operator, users could mount a volume on the operator which contained the overlays, making it rather simple to customize things (e.g. a git volume).

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: so i think we agree? meaning that the functionality is different and therefore it makes sense to have different names for the fields here. feel free to reopen this thread if we don't, in fact, agree :-)

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Yes, kustomize dir layout could be supported as an alternative to the overlay built into the CRD. I think we should support both approaches and which one is used depends on the use case - for just changing a flag, it would be in the CR, for more extensive overlaying through files and full blown kustomize syntax, it could be the kustomize style structure.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: I think you're missing the point, you install the CR and a control plane is installed around it.  The CR does not need to be collocated with the operator itself.  The operator should be capable of installing multiple control planes (and we currently support this in Maistra, for multitenancy, with some caveats).

To clarify the CR namespace not the operator namespace.  Think of it like installing a deployment, the repica set and pods are created in the same namespace as the deployment resource.  Same paradigm.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: Regarding item two, I think it's pretty simple, the operator would install a CR  managed by another operator.  If the CRD isn't available, the install fails.  Hopefully, the error reporting is good enough so the user sees something like, `install failed: kiali install failed: kiali.io/v1alpha1,Kind=kiali CRD not found`.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: The main difference is that `proxy` and `pilot` are under the `operationalConfig` label, not directly under `trafficManagement`.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: How will the operator find its CR if it's not in the same namespace?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: This may be a dumb question, but if different users are administering different aspects of the control plane and in different namespaces, why would we configure all of them with a single CR?  If they are truly operationally owned by different groups, then they shouldn't be part of the same CR, as any operational owner will have the ability to modify the CR, including those aspects for which they are not responsible.

Furthermore, I think you need to consider the security implications of allowing somebody to specify a namespace that they may not have access to.  (We're working on this problem in Maistra, regarding how we ensure random namespaces can't be joined into a mesh in a multitenant environment.)

It may be important to point out that one of the purposes of using an operator is that the operator is the tool that has the privileges to create all the necessary resources, giving regular users the ability to install a control plane (i.e. joe user doesn't need cluster-admin, he just needs permissions to create an istio control plane cr).

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: I think there's some overlap here and some decisions that need to be made.  The operator automates operation of the control plane, which, in my opinion, includes configuration.  Users should be capable of installing a control plane, configured to their specifications, without having to go through extra steps.  What is the usefulness if we don't do that?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: It can be configured to watch any or all namespaces.  It just depends on what you specify for the watch.  Currently, [operator.yaml is configured to watch its own namespace](https://github.com/istio/operator/blob/master/deploy/operator.yaml#L24-L27), but this can be changed to an empty string, in which case it watches all, or set to `istio-system`, in which case it watches `istio-system`.  There are ways to dynamically configure this when using [OLM](https://github.com/operator-framework/operator-lifecycle-manager).

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: You can create a fully configured control plane by supplying the operator CR and a mesh configmap. The question is whether that configmap is its own thing or whether it's schematized and included in the operator CR and we're going with leaving it as its own thing for now. 
The main reason for that is if you embed it in the operator CRD, the components cannot read or respond to it directly. So to make the API dynamic, the operator would have to restart components whenever changes are made.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: That's a fair question - this is actually another reason for separating config concerns from install concerns. If you lump all that together into the operator CRD and put it in its own namespace, feature/component administrators lose control over things they _should_ be allowed to configure. That's a good argument for moving the config out of the operator. You can still argue that some things like pilot resources should belong to the traffic management domain but in practice that administrative boundaries is probably ok because install parameters tend to be fairly static, so operators would likely be ok with going through a process to change them rather than needing direct access.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: i'm ok with adding operationalConfig, but are there any fields that are directly under that other than the components? if not, it seems it would just be adding an empty level to the tree.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Ugh, I think I may now have to put it back exactly for that reason. We never want the install to fail (even a demo) because the CRD is not available and I think the only way to ensure it is to have the operator create that before setting the CR.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: You could do that but it seems unnecessary - to me it makes sense to have the operator CR and resources in the same namespace because administratively the operator and its config belong to the same domain, whereas the Istio control plane components are in different admin domains. See also my other point about moving the config out of the operator and into mesh configmap or component config crds, and only keeping the install footprint in this crd.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: resolving for now, feel free to ping me or open an issue if there's still a concern.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: actually looking over this thread I'd prefer to call this "components", per @linsun 's suggestion. I suppose we could put the common inheritable namespace label at that level so that it's not an empty level. 
is that acceptable to everyone?

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: Actually, there is a good reason to _not_have the operator and CR collocated.  The operator will be running with a service account that has elevated privileges, which would bleed into the control plane namespace if they were collocated.  This allows users with lower privileges to install and manage a control plane.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
```diff
@@ -0,0 +1,597 @@
...
+        - apiVersion: extensions/v1beta1
+          kind: Deployment
+          name: istio-galley
+          patches:
+          - path: spec.template.spec.containers.[name:galley].command.[--livenessProbeInterval]
+            value: --livenessProbeInterval=123s
+*/
+
+
+message IstioInstall {
```
costinm: I strongly believe we should stop using "install" in any of the new CRDs and in operator. 

Install is not the primary purpose - but operation, i.e. continuous configuration of the components after install.
Yes, it can also install ( or remove ) components as needed - or they may be pre-installed by vendor. 

It is highly confusing to use the term install - unless there is something that can never be changed after install ( and I hope we don't have anything in this category ).

ostromart: I'm not attached to any particular name, do you have anything in mind?
IstioInstallation maybe? An installation implies longer term continuity. 
IstioOperatorConfig?

rcernich: I agree with @costinm.  The operator should be providing automating the operation of Istio.  The type I created, as part of the skeleton, is named `IstioControlPlane`, because the operator is managing the lifecycle of a control plane.  'IstioOperatorConfig` I think is inappropriate because you are not configuring the operator.

Simply put, an operator is a controller that watches specific resources and acts on them, performing tasks that a human operator would do.  A simple of example of this, although not called an operator, is the k8s `Deployment`.  It's controller watches those resources, acting on them as changes are made, e.g. rolling out a new set of pods, increasing replica count, etc.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: While I asked the question, I actually believe that the CR be the place where the configuration is exposed to the user.  If the user modifies the configuration, the operator can make sure the changes are rolled out appropriately.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: I don't think this is really an issue.  As the user will have to install the istio operator itself, those additional operator installations can just be part of that installation process, e.g. https://github.com/istio/operator/tree/master/deploy.  (Also, if using OLM, those dependencies can be defined and OLM will take care of installing those dependencies.)

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: I'm going to resolve this since the discussion is revolving around things that are missing rather than part of this PR. Let's tackle the question of adding more knobs to the operator spec separately - that will be a broader discussion and we need to involve other people.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: That's not the WG consensus (cc @costinm, @linsun). I personally spent a lot of time and energy creating a very complete config schema for the operator, but that's not the direction Istio is going, and I understand the rationale. So I'm ok with the decision to keep just installation footprint definition in the operator and move everything else to what is currently called mesh config. Again, I think this is a broader topic and this is not the right place for that discussion - feel free to put it on the agenda for the WG weekly or raise it at the TOC. 
Closing here in the understanding that we can continue in the other forums.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: Sure, sgtm. I changed it to IstioControlPlane.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: We need a one line demo install option - I think that's an absolutely essential requirement. But I'm going to leave this out of this PR in the interests of moving ahead and have created https://github.com/istio/istio/issues/14534 to track this.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
ostromart: I haven't heard any complaints so I've gone ahead and changed this to add a "components" layer.

pkg/apis/istio/v1alpha1/istioinstaller_types.proto
rcernich: What's the difference between the operator CR and the istio CRs?  My assumption is that the control plane would be installed beside the IstioControlPlane CR.  If there's a requirement to deploy components into separate namespaces, then this would support that.  I don't have a problem with that.

I think one needs to be careful about that though, as it could have an impact on the network configuration.  For example, in OpenShift one can configure the default networking such that NetworkPolicy is being used and individual namespaces are segregated by default.  Deploying components to different namespaces would then require joining the networks of those namespaces so they can talk to each other.  That said, it would be up to the operator to detect the underlying networking, adding correctly configured NetworkPolicy resources so the control plane can talk to itself.  In a multitenant environment, containing multiple control planes, work is also required to join the user/application namespaces into the mesh.  This is something we're working on downstream in [Maistra/istio-operator](https://github.com/Maistra/istio-operator).

ostromart: There are three groups of CRs we've mentioned here. I've put suggested namespaces for each in [brackets].
1. operator CR [istio-operator-config-namespace]
2. Istio control plane CRs [istio-control, istio-security etc.]. Actually there are no CRs here, these are all standard k8s R's. Grouping these by feature (and possibly component) into ns allows separation of admin domains for these. This is a pretty basic feature of the new installer @costinm is working on and there's demand from customers for this type of feature admin domain separation. If you look at those charts, they do include explicit ns config so that the control plane can talk to itself.
3. Istio config CRs [istio-config]. This is user config so likely to be in one ns. 
So that leaves operator ns and that could be either in istio-operator-config-namespace or a separate ns - I don't have an opinion, but it should not be istio-config or any other the Istio control plane namespaces.

rcernich: Okay, so what is the operator CR, what does it describe?

ostromart: Exactly what we are looking at in this PR, meaning the enablement/namespaces of features and components and k8s resources for them (and possibly a few other config knobs like enable mtls that have impact on the installation). All other config items are passed through to mesh config. Right now this still results in charts being rendered with these values but the plan going forward (@costin, @mandarjog) is to gradually remove these items from the helm charts and have Istio control plane read it directly from mesh configmap.  
Note that we would still be able to able to use the operator to create/customize this mesh configmap, it just doesn't fall under the operator API we're defining here, but under a separate API.

rcernich: So this config, along with this operator, will be creating these other CRs as part of the installation process?  After which, human operators (administrators) would modify these other CRs?

ostromart: That's the general idea.

PR #5, opened 2019-05-09
Errors slice type

ostromart:
Used in cases where functions return multiple errors.


rcernich:
FWIW, I've been using simple `[]error' along with https://github.com/kubernetes/apimachinery/blob/master/pkg/util/errors/errors.go#L36-L41

ostromart:
This is basically []errors but with solutions to a couple of annoyances: accumulating nil errors in the list, and the case where append doesn't work correctly in functions where there's a named errors return type. I've seen a few alternatives (multierror etc) but this seemed like the most lightweight solution to these problems.

pkg/util/errs.go
```diff
@@ -0,0 +1,94 @@
...
+	for _, e := range b {
+		m[e.Error()] = true
+	}
+	for _, ea := range a {
+		if !m[ea.Error()] {
+			return false
+		}
+	}
+	return true
+}
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/util/errs_test.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func TestToString(t *testing.T) {
+	if got, want := ToString(testErrs), wantStr; got != want {
+		t.Errorf("got: %s, want: %s", got, want)
+	}
+}
+
+func TestNewErrs(t *testing.T) {
+	var errs Errors
```
golangcibot: S1021: should merge variable declaration with assignment on next line (from `gosimple`)

PR #6, opened 2019-05-09
Utilities package common data types

ostromart:
Common data types used in higher level packages.


pkg/util/common.go
```diff
@@ -0,0 +1,46 @@
+package util
+
+import (
+	"gopkg.in/yaml.v2"
+	"path/filepath"
+	"strings"
+)
+
+const (
+	pathSeparator = "."
```
rcernich: what's the rationale for using this as a separator vs a normal path separator?

ostromart: This package is used for overlays which use a jsonpath style of path.

pkg/util/common.go
```diff
@@ -0,0 +1,46 @@
+package util
+
+import (
+	"gopkg.in/yaml.v2"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

PR #7, opened 2019-05-09
Manifest manipulation package

ostromart:
Utility package for moving between objects and manifest YAMLs.
Used in https://github.com/istio/operator/pull/8.


rcernich:
The types need to be documented.  I'm not sure what the purpose is behind these types, which makes it difficult to review.

ostromart:
Finally coming back to clean this up with docs.

pkg/manifest/objects.go
```diff
@@ -0,0 +1,271 @@
...
+package manifest
+
+import (
+	"bytes"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"istio.io/pkg/log"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/manifest/objects.go
```diff
@@ -0,0 +1,294 @@
...
+func (o *Object) YAML() ([]byte, error) {
+	if o.yaml != nil {
+		return o.yaml, nil
+	}
+	// TODO: there has to be a better way.
+	oj, err := o.JSON()
+	if err != nil {
+		return nil, err
+	}
+	o.json = oj
```
richardwxn: This line is not necessary

pkg/manifest/objects.go
```diff
@@ -0,0 +1,326 @@
...
+		Group:  gvk.Group,
+		Kind:   gvk.Kind,
+		Name:   u.GetName(),
+		json:   json,
+	}, nil
+}
+
+// ParseYAMLToObject parses YAML to an Object.
+func ParseYAMLToObject(yaml []byte) (*Object, error) {
+	r := bytes.NewReader([]byte(yaml))
```
golangcibot: unnecessary conversion (from `unconvert`)

pkg/manifest/objects.go
```diff
@@ -0,0 +1,326 @@
...
+	decoder := k8syaml.NewYAMLOrJSONDecoder(r, 1024)
+
+	out := &unstructured.Unstructured{}
+	err := decoder.Decode(out)
+	if err != nil {
+		return nil, fmt.Errorf("error decoding object: %v", err)
+	}
+	return NewObject(out, nil, []byte(yaml)), nil
```
golangcibot: unnecessary conversion (from `unconvert`)

pkg/manifest/objects.go
```diff
@@ -0,0 +1,294 @@
...
+	b, err := o.object.MarshalJSON()
+	if err != nil {
+		return nil, err
+	}
+	o.json = b
+	return b, nil
+}
+
+// YAML returns a YAML representation of o, using an internal cache.
+func (o *Object) YAML() ([]byte, error) {
```
richardwxn: Why it needs to get JSON representation first?

martonsereg: `JSONToYAML` uses `yaml.Marshal()` inside: https://github.com/ghodss/yaml/blob/master/yaml.go#L101
So I guess a simple `yaml.Marshal` would do it instead of marshalling to JSON and back to YAML.
Afaik apimachinery's `MarshalJSON` doesn't do anything special (other than handling `UnstructuredList` as well, but it isn't needed here), but it would need some testing because I'm not 100% sure.

pkg/manifest/objects.go
```diff
@@ -0,0 +1,326 @@
...
+	o.object.SetLabels(merged)
+	// Invalidate cached json
+	o.json = nil
+	o.yaml = nil
+}
+
+// Objects holds a collection of objects, so that we can filter / sequence them
+type Objects struct {
+	Items []*Object
+}
```
martonsereg: Do we need a struct here? Wouldn't `type Objects []*Object` be enough so it can be iterated instead of getting `.Items` all the time?

pkg/manifest/objects.go
```diff
@@ -0,0 +1,326 @@
...
+
+// ParseObjectsFromYAMLManifest returns an Objects represetation of manifest.
+func ParseObjectsFromYAMLManifest(manifest string) (*Objects, error) {
```
martonsereg: perhaps it would be good to have some tests against this parsing mechanism

pkg/manifest/objects.go
```diff
@@ -0,0 +1,326 @@
...
+	var b bytes.Buffer
+
+	var yamls []string
+	for _, line := range strings.Split(manifest, "\n") {
```
martonsereg: wouldn't it be better to use something like the following instead of splitting the text along `\n`s?
```
scanner := bufio.NewScanner(strings.NewReader(manifest))
for scanner.Scan() {
   line := scanner.Text()
   ...
}
```

pkg/manifest/objects.go
```diff
@@ -0,0 +1,294 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
+
+	"istio.io/pkg/log"
+)
+
+// Object is an in-memory representation of a k8s object, used for moving between different representations
+// (Unstructured, JSON, YAML) with cached rendering.
+type Object struct {
```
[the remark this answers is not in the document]
ostromart: Agreed, but since this is used extensively in other PR's I propose to defer this change to a future PR, tracked here: https://github.com/istio/istio/issues/15099

pkg/manifest/objects.go
```diff
@@ -0,0 +1,294 @@
...
+	o.Group = gvk.Group
+	o.Kind = gvk.Kind
+	o.Name = u.GetName()
+	o.Namespace = u.GetNamespace()
+
+	return o
+}
+
+// Hash returns a unique, insecure hash based on kind, namespace and name.
+func Hash(kind, namespace, name string) string {
```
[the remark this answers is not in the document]
ostromart: This is a bit confusing, probably because the design is not that great. The idea is to create a hash that does not require an object so it can be used outside the package to store objects in a map. The missing link is that the two hashes are the same.

pkg/manifest/objects.go
```diff
@@ -0,0 +1,294 @@
...
+	o.json = oj
+	y, err := yaml.JSONToYAML(oj)
+	if err != nil {
+		return nil, err
+	}
+	o.yaml = y
+	return y, nil
+}
+
+// YAML returns a YAML representation of o, or an error string if the object cannot be rendered to YAML.
```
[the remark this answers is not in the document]
ostromart: This one doesn't return an error.

pkg/manifest/objects.go
```diff
@@ -0,0 +1,326 @@
...
+	u, ok := o.(*unstructured.Unstructured)
+	if !ok {
+		return nil, fmt.Errorf("parsed unexpected type %T", o)
+	}
+
+	return &Object{
+		object: u,
+		Group:  gvk.Group,
+		Kind:   gvk.Kind,
+		Name:   u.GetName(),
```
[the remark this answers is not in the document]
ostromart: Yes, it is better - done.

pkg/manifest/objects.go
ostromart: Are you sure? :-) It looks necessary but I may be missing something.

pkg/manifest/objects.go
ostromart: Yes, good point. This ended up this way because there were other members which are gone now, it's a good time to clean up.

pkg/manifest/objects.go
ostromart: Yes, definitely. I think we have coverage here because of inclusion in other tests but explicit unit tests are required too: https://github.com/istio/istio/issues/15101

pkg/manifest/objects.go
ostromart: Good idea - done.

pkg/manifest/objects.go
ostromart: @richardwxn, see http://ghodss.com/2014/the-right-way-to-handle-yaml-in-golang/

pkg/manifest/objects.go
richardwxn: @ostromart I think when you call o.JSON(), it already do this for you at line131?

pkg/manifest/objects.go
```diff
@@ -0,0 +1,343 @@
...
+				return nil, err
+			}
+		}
+	}
+	yamls = append(yamls, b.String())
+
+	var objects Objects
+
+	for _, yaml := range yamls {
+		// We need this so we don't error on a file that is commented out
```
waynz0r: Couldn't we use `ParseYAMLToObject` here and also dealing with commented manifest snippets by checking the parsed object for `Kind`?

```
o, err := ParseYAMLToObject([]byte(yaml))
if err != nil {
	log.Infof("error decoding object: %s\n%s\n", err, yaml)
	continue
}

if o.GroupKind().Group == "" {
	continue
}

objects = append(objects, o)
```

pkg/manifest/objects.go
ostromart: Yes, good point. Done.

pkg/manifest/objects.go
ostromart: Yes, that's a great suggestion, thanks.

PR #8, opened 2019-05-09
Package for patching kubernetes resources

ostromart:
This code contains the guts of kustomize style overlays for the generated manifest. 
Please see code and tests for documentation and examples. 
Depends on https://github.com/istio/operator/pull/34 for Go structs compiled from proto.


rcernich:
General comment, if this is supposed to represent "kustomize" type capabilities, would it be better to try to use the actual kustomize code, e.g. like we do with helm?

For example, use functionality from any of these packages: https://github.com/kubernetes-sigs/kustomize/tree/master/pkg

rcernich:
Okay, I don't think I understood what was going on here initially.  Are you creating a schema for defining patches in the CR?  If so, I think that has real potential to bloat the CR, with all of this patching.

If that is the case, then maybe, instead of defining your own schema for patching, it might be better to utilize json merge patch or strategic merge patch descriptions, as these are standard and supported by things like kubectl patch.  Users would at least have a head start in understanding the format.

ostromart:
The problem is neither json merge nor strategic patch support value based path selection. 
What this means in practice is that you cannot modify just a flag in one of the containers (since this is a list item) as shown in the unit tests here. You have to re-specify the entire container. I started off thinking I could just use json merging but I just don't think it's workable, since customizing container flags is going to be a very common use case for this overlay mechanism.

rcernich:
I believe strategic merge patch supports ordering and removal.  Here's a link to the docs:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md

ostromart:
It does, but what I'm saying is list entry selection through key is not supported, so you have to replace/remove the entire list item. For example, how would you remove the following line with json/strategic merge?
https://gist.github.com/chanwit/4e1e4f951cf3357c1c3d7804f8162312#file-istio-demo-yaml-L2815
(say you want to remove --webhook-config-file).

rcernich:
According to the jsonpatch doc, while accessing a list item by key is not supported (what key would you use?), you can access them by index.  For example, from http://jsonpatch.com/

```json
{
  "op": "replace",
  "path": "/biscuits/0/name",
  "value": "Chocolate Digestive"
}
```

You can find more k8s specific examples if you look (e.g. changing the image name on specific container within a deployment).

ostromart:
Right, jsonpatch only supports indexing which is too brittle to use in practice, since we cannot depend on item index in lists remaining constant over time.

waynz0r:
The base functionality looks fine to me to move forward with it for now, but we should remove the dependencies to the github.com/ostromart/operator repo before merging this in.

ostromart:
Removed the dependencies (these were just left over from testing in my local ws) and added some minor fix. If there are no comments this can be approved, anyway it can not go in until the dependencies are merged.

pkg/patch/patch.go
```diff
@@ -0,0 +1,401 @@
...
+	by, err := base.YAML()
+	if err != nil {
+		return nil, util.NewErrs(err)
+	}
+	err = yaml.Unmarshal(by, bo)
+	if err != nil {
+		return nil, util.NewErrs(err)
+	}
+	for _, p := range patches {
+		dbgPrint("applying path=%s, value=%s\n", p.Path, p.Value)
```
golangcibot: p.Value undefined (type *v1alpha2.K8SObjectOverlay_PathValue has no field or method Value) (from `typecheck`)

pkg/patch/patch.go
```diff
@@ -0,0 +1,401 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// objectOverrideMap converts oos, a slice of object overlays, into a map of the same overlays where the key is the
+// object manifest.Hash.
+func objectOverrideMap(oos []*v1alpha2.K8SObjectOverlay, namespace string) (map[string][]*v1alpha2.K8SObjectOverlay_PathValue, error) {
```
golangcibot: objectOverrideMap - result 1 (error) is always nil (from `unparam`)

pkg/patch/patch.go
```diff
@@ -0,0 +1,401 @@
...
+				return getNode(nn, fullPath, remainPath[1:])
+			}
+		}
+		return nil, fmt.Errorf("path %s: element %s not found", fullPath, pe)
+	}
+
+	dbgPrint("interior node")
+	// non-list node.
+	if nnt, ok := nc.node.(map[interface{}]interface{}); ok {
+		var nn interface{}
```
golangcibot: S1021: should merge variable declaration with assignment on next line (from `gosimple`)

PR #11, opened 2019-05-10
Recursive directory watcher

ostromart:
I haven't seen a simple one with debounce so this is a quick implementation.


rcernich:
Where would this be used?

ostromart:
The plan was to use this to trigger reconciliation if the chart path changes or new charts appear. But this functionality is a ways off so let me withdraw this review for now.

PR #14, opened 2019-05-13
Flesh out basic helm-based reconciler

rcernich:


pkg/controller/istiocontrolplane/inputs.go
```diff
@@ -0,0 +1,91 @@
...
+	"istio/charts/galley",
+	"istio/charts/prometheus",
+	"istio/charts/mixer",
+	"istio/charts/pilot",
+	"istio/charts/gateways",
+	"istio/charts/sidecarInjectorWebhook",
+	"istio/charts/grafana",
+	"istio/charts/tracing",
+	"istio/charts/kiali",
+}
```
ostromart: We should be using istio/installer charts here, these ones are deprecated.

pkg/controller/istiocontrolplane/inputs.go
```diff
@@ -0,0 +1,91 @@
...
+// IstioRenderingInput is a RenderingInput specific to an IstioControlPlane instance.
+type IstioRenderingInput struct {
+	instance *v1alpha1.IstioControlPlane
+}
+
+var _ helmreconciler.RenderingInput = &IstioRenderingInput{}
+
+// NewRenderingInput creates a new IstioRenderiongInput for the specified instance.
+func (f *IstioInputFactory) NewRenderingInput(instance runtime.Object) (helmreconciler.RenderingInput, error) {
+	istioControlPlane, ok := instance.(*v1alpha1.IstioControlPlane)
```
ostromart: NewFoo functions should generally avoid errors if possible. You can avoid it by taking IstioControlPlane as input here.

pkg/controller/istiocontrolplane/inputs.go
```diff
@@ -0,0 +1,91 @@
...
+
+// GetChartPath returns the absolute path locating the charts to be rendered.
+func (i *IstioRenderingInput) GetChartPath() string {
+	path := i.instance.Spec.ChartPath
+	if len(path) == 0 {
+		return filepath.Join(controllerConfiguration.BaseChartPath, controllerConfiguration.DefaultChartPath)
+	} else if filepath.IsAbs(path) {
+		return i.instance.Spec.ChartPath
+	}
+	return filepath.Join(controllerConfiguration.BaseChartPath, path)
```
ostromart: Seems like it might be better to bind the path value when you create the renderer, or even the input to it. That way you don't need to do work every time to get the path. Also, IstioRenderingInput should probably not concern itself with logic about how the path is constructed.

pkg/controller/istiocontrolplane/inputs.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// GetTargetNamespace returns the namespace within which rendered namespaced resources should be generated
+// (i.e. Release.Namespace)
+func (i *IstioRenderingInput) GetTargetNamespace() string {
+	return i.instance.Namespace
+}
+
+// GetProcessingOrder returns the order in which the rendered charts should be processed.
+func (i *IstioRenderingInput) GetProcessingOrder(manifests map[string][]manifest.Manifest) ([]string, error) {
```
ostromart: map[string][]manifest.Manifest seems worthy of its own type if it gets passed around much.

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
```diff
@@ -35,7 +36,14 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileIstioControlPlane{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	factory := &helmreconciler.Factory{
+		CustomizerFactory: &helmreconciler.CustomizerFactory{
+			InputFactory:    &IstioInputFactory{},
+			MarkingsFactory: &IstioMarkingsFactory{},
+			ListenerFactory: &IstioListenerFactory{},
+		},
```
ostromart: The factory terminology here is a bit confusing. Why do we need to pass in factories?

pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -0,0 +1,264 @@
...
+
+// IstioRenderingListener is a RenderingListener specific to IstioControlPlane resources
+type IstioRenderingListener struct {
+	*helmreconciler.CompositeRenderingListener
+}
+
+// NewRenderingListener returns a new IstioRenderingListener, which is a composite that includes IstioStatusUpdater
+// and IstioChartCustomizerListener.
+func (f *IstioListenerFactory) NewRenderingListener(instance runtime.Object) (helmreconciler.RenderingListener, error) {
+	if istioControlPlane, ok := instance.(*v1alpha1.IstioControlPlane); ok {
```
ostromart: As above, recommend requiring the caller to pass in *v1alpha1.IstioControlPlane

pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -0,0 +1,264 @@
...
+// IstioStatusUpdater is a RenderingListener that updates the status field on the IstioControlPlane
+// instance based on the results of the Reconcile operation.
+type IstioStatusUpdater struct {
+	*helmreconciler.DefaultRenderingListener
+	instance   *v1alpha1.IstioControlPlane
+	reconciler *helmreconciler.HelmReconciler
+}
+
+var _ helmreconciler.RenderingListener = &IstioStatusUpdater{}
+var _ helmreconciler.ReconcilerListener = &IstioStatusUpdater{}
```
ostromart: What's the purpose of these?

pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -0,0 +1,264 @@
...
+		DefaultRenderingListener: &helmreconciler.DefaultRenderingListener{},
+		instance:                 instance,
+	}
+}
+
+// EndReconcile updates the status field on the IstioControlPlane instance based on the resulting err parameter.
+func (u *IstioStatusUpdater) EndReconcile(_ runtime.Object, err error) error {
+	status := u.instance.Status
+	if err == nil {
+		if condition := status.GetCondition(v1alpha1.ConditionTypeInstalled); condition.Status != v1alpha1.ConditionStatusTrue {
```
ostromart: This could be greatly simplified if we move to a simpler enum based state machine.

pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -0,0 +1,264 @@
...
+		}
+		logger.Error(err, "unexpected error occurred waiting for deployment to become ready", gvk.Kind, name)
+		return false, err
+	})
+	if err != nil {
+		logger.Error(nil, "deployment failed to become ready in a timely manner", gvk.Kind, name)
+	}
+}
+
+// KialiChartCustomizer is a ChartCustomizer for the kiali chart
```
ostromart: This should probably be in a separate file.

pkg/controller/istiocontrolplane/markings.go
```diff
@@ -0,0 +1,107 @@
...
+	// OwnerGroupKey represents the group of the owner to which the resource relates
+	OwnerGroupKey = MetadataNamespace + "/owner-group"
+
+	// OwnerGenerationKey represents the generation to which the resource was last reconciled
+	OwnerGenerationKey = MetadataNamespace + "/owner-generation"
+)
+
+var (
+	// XXX: move this into a ConfigMap so users can override things if they add new types in customized charts
+	// ordered by which types should be deleted, first to last
```
ostromart: Seems like an overkill to allow users to add types through ConfigMap. If they go to such lengths, they probably forked Istio and should fork the operator too.

pkg/controller/istiocontrolplane/markings.go
```diff
@@ -0,0 +1,107 @@
...
+	namespacedResources = []schema.GroupVersionKind{
+		{Group: "autoscaling", Version: "v2beta1", Kind: "HorizontalPodAutoscaler"},
+		{Group: "policy", Version: "v1beta1", Kind: "PodDisruptionBudget"},
+		{Group: "route.openshift.io", Version: "v1", Kind: "Route"},
+		{Group: "apps.openshift.io", Version: "v1", Kind: "DeploymentConfig"},
```
ostromart: Let's stick to istio APIs - external APIs should be plugins.

pkg/controller/istiocontrolplane/markings.go
```diff
@@ -0,0 +1,107 @@
...
+		{Group: "networking.istio.io", Version: "v1alpha3", Kind: "Gateway"},
+		{Group: "networking.istio.io", Version: "v1alpha3", Kind: "VirtualService"},
+	}
+
+	// ordered by which types should be deleted, first to last
+	nonNamespacedResources = []schema.GroupVersionKind{
+		{Group: "admissionregistration.k8s.io", Version: "v1beta1", Kind: "MutatingWebhookConfiguration"},
+		{Group: "admissionregistration.k8s.io", Version: "v1beta1", Kind: "ValidatingWebhookConfiguration"},
+		{Group: "certmanager.k8s.io", Version: "v1alpha1", Kind: "ClusterIssuer"},
+		{Group: "oauth.openshift.io", Version: "v1", Kind: "OAuthClient"},
```
ostromart: Ditto.

pkg/controller/common/util.go
```diff
@@ -0,0 +1,117 @@
...
+		// XXX: panic?
+		return
+	}
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	labels[label] = value
+	_ = accessor.SetLabels(resource, labels)
+}
+
```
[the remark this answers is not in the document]
rcernich: agreed, but i was copying this stuff everywhere, so i moved it out to a helper.  this guy probably should return an error if it's not an object.

pkg/controller/istiocontrolplane/inputs.go
rcernich: at this point, i'd like to leave that as a future task, as what i have working in maistra is based on the charts in istio/istio.

pkg/controller/istiocontrolplane/inputs.go
rcernich: agreed, although it's only called once.

as for it being part of the rendering input, it is part of the input for rendering, so i would argue the logic belongs here.  that said, i could see logic being part of args.go, as that is what's driving the logic.  either way, i don't really care.  i'll move it to args.go.

pkg/controller/istiocontrolplane/inputs.go
rcernich: good idea.

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
rcernich: yes, i suppose this could be a single factory with three methods.  i started out by defining the interfaces for input, markings, and hooks (which became a listener/callback interface) and kept the separation through the factories.  i'll look at consolidating these into a single factory interface.

pkg/controller/istiocontrolplane/listeners.go
rcernich: maybe this is possible after consolidating the factory interfaces into one.  that said, the helmreconciler bit is intended to be completely generic (hence my comment above about templates/generics).

pkg/controller/istiocontrolplane/listeners.go
rcernich: this is simple type checking, although their worth is suspect since DefaultRenderingListener implements those interfaces already.  (they provide a compile error if the interface changes.)

pkg/controller/istiocontrolplane/listeners.go
rcernich: fine with me.  i can look into that.

pkg/controller/istiocontrolplane/listeners.go
rcernich: yeah, they can be split up, but right now, there's not really any customization besides the status field.  these wait functions could be generic.  i don't have an opinion one way or the other.

pkg/controller/istiocontrolplane/markings.go
rcernich: yeah, i don't know.  this is really just for the pruning logic and there's probably not much else that could be added anyway.  i'll delete the comment.  thanks for the feedback.

pkg/controller/istiocontrolplane/markings.go
rcernich: whoops, good catch.  i thought i pruned out all the openshift specific types.

pkg/apis/istio/v1alpha1/istiocontrolplane_types.go
```diff
@@ -44,6 +60,115 @@ type IstioControlPlaneList struct {
...
+const (
+	// ConditionReasonInstallSuccessful ...
+	ConditionReasonInstallSuccessful ConditionReason = "InstallSuccessful"
+	// ConditionReasonInstallError ...
+	ConditionReasonInstallError ConditionReason = "InstallError"
+	// ConditionReasonReconcileSuccessful ...
+	ConditionReasonReconcileSuccessful ConditionReason = "ReconcileSuccessful"
+	// ConditionReasonReconcileError ...
+	ConditionReasonReconcileError ConditionReason = "ReconcileError"
+)
```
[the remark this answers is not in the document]
rcernich: i was trying to follow the pattern used by other types, e.g. appsv1.Deployment.Status.Conditions.

while a simple status value would be...simple, i think there's value into having conditions.  by having separate conditions for installation and reconciliation, one can determine whether the initial installation failed or if there was a failure processing later updates.  i think this would be useful information to have, especially because processing updates to the CR could be problematic.  do you mind if we leave it for now?

pkg/helmreconciler/listeners.go
```diff
@@ -0,0 +1,472 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/helm/pkg/manifest"
+
+	"istio.io/operator/pkg/controller/common"
+)
+
+// CompositeRenderingListener is an implementation of RenderingListener which is composed of an array of listeners.
```
ostromart: better to separate these out into files.

pkg/helmreconciler/listeners.go
```diff
@@ -0,0 +1,472 @@
...
+// All methods are delegated to each element in the Listeners array.  For completion events (e.g. EndResource()), the
+// delegates are invoked last to first.
+type CompositeRenderingListener struct {
+	// Listeners represents a list of Listeners to which this object will delegate calls.
+	Listeners []RenderingListener
+}
+
+var _ RenderingListener = &CompositeRenderingListener{}
+var _ ReconcilerListener = &CompositeRenderingListener{}
```
ostromart: remove

pkg/helmreconciler/listeners.go
```diff
@@ -0,0 +1,472 @@
...
+		}
+	}
+	return utilerrors.NewAggregate(allErrors)
+}
+
+// ResourceError delegates ResourceError to the Listeners in first to last order.
+func (l *CompositeRenderingListener) ResourceError(obj runtime.Object, err error) error {
+	// reverse order for completions
+	allErrors := make([]error, 0, len(l.Listeners))
+	for index := len(l.Listeners) - 1; index > -1; index-- {
```
ostromart: for l, idx := range l.Listeners
here and elsewhere

pkg/helmreconciler/listeners.go
```diff
@@ -0,0 +1,472 @@
...
+			allErrors = append(allErrors, listenerErr)
+		}
+	}
+	return utilerrors.NewAggregate(allErrors)
+}
+
+// EndResource delegates EndResource to the Listeners in last to first order.
+func (l *CompositeRenderingListener) EndResource(obj runtime.Object) error {
+	// reverse order for completions
+	allErrors := make([]error, 0, len(l.Listeners))
```
ostromart: you don't have to make slices for appending (you can append to nil slice).
use var errs []error here instead.

pkg/helmreconciler/patching.go
```diff
@@ -0,0 +1,158 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CreatePatch creates a patch based on the current and updated versions of an object
+func (h *HelmReconciler) CreatePatch(current, updated runtime.Object) (Patch, error) {
+	var retVal Patch
+	patch := &basicPatch{client: h.client}
+	currentAccessor, err := meta.Accessor(current)
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("cannot create object accessor for current object:\n%v", current))
```
ostromart: any reason not to just use return nil, fmt.Errorf(), here and elsewhere?

pkg/controller/common/util.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	_, ok := annotations[annotation]
+	return ok
+}
+
+// DeleteAnnotation is a helper function which deletes the specified annotation from the specified object.
+func DeleteAnnotation(resource runtime.Object, annotation string) {
+	annotations, err := meta.NewAccessor().Annotations(resource)
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
rcernich: If there are no annotations, e.g. because the object doesn't support them, who cares.  The object has no annotations, so nothing to delete; no different than the case where the annotation doesn't exist.

pkg/controller/common/util.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	if annotations == nil {
+		return
+	}
+	delete(annotations, annotation)
+	_ = meta.NewAccessor().SetAnnotations(resource, annotations)
```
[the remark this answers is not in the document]
rcernich: That's why I added that, so it would clear the linter check in CI.  We know it won't err, because we've already succeeded with GetAnnotations().  That said, let me change the code so that i preserve the accessor, that way, method signature we're calling doesn't return an error type.

pkg/controller/common/util.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// GetAnnotation is a helper function which returns the value of the specified annotation on the specified object.
+// returns ok=false if the annotation was not found on the object.
+func GetAnnotation(resource runtime.Object, annotation string) (value string, ok bool) {
+	annotations, err := meta.NewAccessor().Annotations(resource)
+	if err != nil {
+		return
```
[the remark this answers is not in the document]
rcernich: if the object doesn't have annotations, there's nothing to return.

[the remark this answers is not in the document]
rcernich: Also, by default, this will return, "", false.  I don't think an error is useful here and it clutters the api.

pkg/helmreconciler/listeners.go
rcernich: are you suggesting each type be separated out into individual files?

pkg/helmreconciler/listeners.go
rcernich: Those provide type checking, ensuring that CompositeRenderingListener properly implements those interfaces.

pkg/helmreconciler/listeners.go
rcernich: this is reverse order iteration.

pkg/helmreconciler/listeners.go
rcernich: This simply prevents resizing the slice.  At most, there will be len(l.Listeners) errors, so we presize the slice to avoid allocating memory during append.

pkg/helmreconciler/patching.go
rcernich: I wanted to collect the upstream error, which is often useful for debugging and support.

pkg/apis/istio/v1alpha1/istiocontrolplane_types.go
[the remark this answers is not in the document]
rcernich: we can address this in a future pr

pkg/controller/common/util.go
```diff
@@ -0,0 +1,117 @@
...
+func IndexOf(l []string, s string) int {
+	for i, elem := range l {
+		if elem == s {
+			return i
+		}
+	}
+	return -1
+}
+
+var accessor = meta.NewAccessor()
```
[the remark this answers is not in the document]
rcernich: this has been removed.

pkg/controller/istiocontrolplane/listeners.go
rcernich: this can be done in a future pr.

pkg/controller/istiocontrolplane/markings.go
rcernich: removed.

pkg/controller/istiocontrolplane/markings.go
rcernich: resolved.

pkg/controller/istiocontrolplane/markings.go
rcernich: removed.

pkg/controller/common/util.go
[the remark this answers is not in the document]
rcernich: added comments, which hopefully clarify why there's no need to propagate errors for get/has/delete.

pkg/controller/common/util.go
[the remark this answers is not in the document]
rcernich: modified the way this is implemented to avoid `_ =`

pkg/helmreconciler/interfaces.go
```diff
@@ -0,0 +1,185 @@
...
+	// Input returns the RenderingInput that should be used to rendering the charts.
+	Input() RenderingInput
+	// PruningDetails returns the details required to support pruning of obsolete resources.
+	PruningDetails() PruningDetails
+	// Listener returns the RenderingListener that should handle the rendering callbacks.
+	Listener() RenderingListener
+}
+
+type RenderingCustomizerFactory interface {
+	// NewCustomizer returns a new RenderingCustomizer for the specified object.
```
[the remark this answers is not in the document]
rcernich: So customizations can be injected per instance being reconciled.  We new up a HelmReconciler for every instance coming through the Reconcile() callback.  This is primarily to encapsulate all state associated with the reconciliation of a specific object.

pkg/helmreconciler/interfaces.go
```diff
@@ -0,0 +1,185 @@
...
+	NewCustomizer(obj runtime.Object) (RenderingCustomizer, error)
+}
+
+// PruningDetails define the labels and annotations used to mark resources managed by the operator, as well as the
```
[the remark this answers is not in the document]
rcernich: Yeah, I don't know if that's any better.  What really throws the naming off is the inclusion of `ResourceTypes`.  I thought `Markings` worked great for the labels/annotations, but ... ResourceTypes, which specifically applies to the pruning logic (well it all does really, as we use the labels and annotations to list specific types of resources for pruning).

pkg/helmreconciler/listeners.go
```diff
@@ -0,0 +1,472 @@
...
+	l.reconciler.logger = l.reconciler.logger.WithValues("chart", chart)
+	l.reconciler.logger.V(l.Level).Info("begin updating resources for chart")
+	return manifests, nil
+}
+
+// BeginResource logs the event and updates the logger to log with values resource=name, kind=kind, apiVersion=api-version
+func (l *LoggingRenderingListener) BeginResource(obj runtime.Object) (runtime.Object, error) {
+	l.loggerStack = append(l.loggerStack, l.reconciler.logger)
+	accessor := meta.NewAccessor()
+	kind, _ := accessor.Kind(obj)
```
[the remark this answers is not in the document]
rcernich: This is just setting up the logger for the call with value pairs.  If this errors, it should not result in an error for the install.  This is why the error is ignored.  It doesn't make sense to return an error if it doesn't result in an error that should be handled.  That is unless you think reconciliation should fail because we couldn't add metadata to the log entries?  That's a different discussion.

pkg/helmreconciler/listeners.go
rcernich: this can be done in a subsequent pr, if necessary.

rcernich:
This is ready to be merged.  Any subsequent feedback/discussion should be done through an issue or follow-on PR.

pkg/helmreconciler/listeners.go
ostromart: If it doesn't implement it then you won't be able to pass it through an interface during compile time. If you don't pass it through an interface, then you don't need the interface definition (YAGNI). 
I looked through the k8s codebase and the only reason I can see for having these is where an implementation is provided/generated which is not used in the same compilation unit. That's not the case here so all this is doing is adding dead code and making things less readable. 
But I may be misunderstanding/misreading so I'm totally open to soliciting outside opinions on this.

pkg/helmreconciler/listeners.go
ostromart: But what if you have no errors at all (most common case)? Then you end up allocating and freeing memory every time unnecessarily.

pkg/helmreconciler/listeners.go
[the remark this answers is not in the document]
ostromart: Keep in mind that one of the goals is for other developers to be able to easily reason about the code. What you just explained above is not obvious from reading the code and leads to unnecessary head scratching for future contributors. If you have to spend a paragraph explaining why an error is ok to ignore, isn't it easier just to handle it?

pkg/apis/istio/v1alpha1/istiocontrolplane_types.go
```diff
@@ -21,6 +32,11 @@ type IstioControlPlaneStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// ObservedGeneration represents the last generation processed by the controller.  This is used to short-circuit reconcilations.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+	// Conditions represent the condition of the control plane
+	Conditions []Condition `json:"conditions,omitempty"`
```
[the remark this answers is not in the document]
rcernich: I thought we agreed to defer this to a follow on PR.

pkg/helmreconciler/listeners.go
rcernich: A fair point, although I don't see why you didn't mention this in the first place.  Still, this shouldn't be holding up the PR.

rcernich:
Rebased again and ready for merging.

rcernich:
Bump.  Can we get this merged now?

pkg/helmreconciler/listeners.go
ostromart: I didn't mention it because I didn't look at the behavior in detail till you explained why you were pre-sizing. 
You really want to check this in rather than just fix it?

pkg/apis/istio/v1alpha1/istiocontrolplane_types.go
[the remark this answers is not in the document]
ostromart: I made this comment 11 days ago and 8 days ago you commented "I propose we defer this to a follow on pr". I don't think I explicitly agreed at that time but sure, we can defer it.

pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -0,0 +1,256 @@
...
+		logger.Error(err, fmt.Sprintf("could not get object accessor for %s", gvk.Kind))
+		return
+	}
+	name := objectAccessor.GetName()
+	namespace := objectAccessor.GetNamespace()
+	deployment := &unstructured.Unstructured{}
+	deployment.SetGroupVersionKind(gvk)
+	// wait for deployment replicas >= 1
+	logger.Info("waiting for deployment to become ready", gvk.Kind, name)
+	err = wait.ExponentialBackoff(wait.Backoff{Duration: 6 * time.Second, Steps: 10, Factor: 1.1}, func() (bool, error) {
```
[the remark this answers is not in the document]
ostromart: ping

pkg/helmreconciler/listeners.go
ostromart: Please don't resolve conversations that are not resolved. If we agree to disagree that's one thing and the path forward for that is for a TL to step in to decide.

ostromart:
There are still outstanding issues before this can be merged:
1. move hard coded consts
2. fix slice pre-allocation
3. respond to my comment about unneeded interface compilation checks

rcernich:
Hey @ostromart,

1.  consts have been moved to a `const` block.
2. I don't think the preallocation is a problem that needs to be fixed.  I think it's a best practice.  (Although, given that this is not performance sensitive, probably overkill.)
3. I believe the interface checks are required, as they serve a variety of purposes:
   1. They validate that the type actually implements the interface, without having to use it.
   2. If the type is used in a switch or dynamic type cast, there is no way to verify that it implements the interface without these statements.
   3. Because these statements are immediately following the type definition, they help document the interfaces implemented by the type, which improves readability (as go has no `implements` clause).

HTH

ostromart:
Hey Rob, I had a chat with a couple of very experienced Go programmers regarding 3. They agree that this construct serves no purpose except in the context of a library, which this code isn't. However, they recall having a similar debate years ago (with someone who was more used to Java) and they decided to agree to let the code in just for the sake on moving on. So I'll follow their example, but these constructs are completely redundant and therefore just reduce readability:
i. this only happens in a library
ii. in order for this to happen you'd have to create an interface type in a function and then try to cast it to the type which doesn't make sense. the moment you pass the interface implementation through a function call, it is checked by the compiler. 
iii. go has a great documentation system and the generally accepted way to document that something implements an interface is to put that in the comment. 
If you don't believe me, I don't have any problem with you posting this discussion to golang-nuts to convince yourself. 
That leaves 2 and I disagree that this is best practice and is rather a logic error. What you are doing is always preallocating a slice of a certain size in a situation where the common case is that the slice size should be zero (and in almost all other cases will be smaller than your preallocated slice size). There are valid scenarios where preallocating slices makes sense but this is not such a case. Preallocating is to increase efficiency - here you're actually reducing efficiency.
Please correct that case and I'm otherwise ok to approve.

rcernich:
Removed the array allocation.

Here's an example where not asserting the interface type breaks down, per 3.ii. in my earlier comment:

https://github.com/istio/operator/pull/14/files#diff-5908b9bed8a59172195218904b0c996eR42

Note that this would result in a silent failure that may be difficult to track down (why isn't my callback being invoked?  Whoops, I fat fingered the method declaration).

Anyway, as you said, we'll have to agree to disagree on this one.

ostromart:
The code you pointed me to actually looks wrong - what you have there is two parallel level interfaces that have nothing to do with each other in the Go structural sense (Rendering and ReconcilerListener are two separate interfaces with no common base interface). Inside of CompositeRenderingListener you have a member RenderingListener that in a function you are casting to a ReconcilerListener. What you should be doing (if I understand your intent correctly) is to create a common base interface (say Listener) and inheriting that into both interfaces. CompositeRenderingListener should have a slice of Listener. And inside of its functions you should not be downcasting to a particular type of listener - doing so indicates an interface design problem. 
I'm happy to give more of this type of high level review feedback, but please make the reviews smaller and include design docs for anything that involves interface design.

rcernich:
I would say the naming is confusing then.  RenderingListener handles callbacks during rendering, while ReconcilerListener allows objects to accept the current HelmReconciler.

Regarding the interface design, I spent over two hours on a call with you going over the design in great detail.  I thought we were okay after that call, with a few minor tweaks, but I guess I was mistaken.

That said, thanks for merging this.

ostromart:
It's not about naming being confusing. You have two interfaces A and B that have no common parent and you have a type C that has a slice of A that is cast to B in one of the functions. That's not how interfaces are used in Go. The correct way would be create a parent Z which A and B inherit and C only uses the interface defined in Z if it needs to store A and Bs in a slice of Z's. 
Design docs are not just for me but for future contributors trying to understand the code. If there was a design doc we'd probably spend less time going over it live, although that was a good walkthrough and I found it very helpful.

PR #23, opened 2019-05-25
Add local codecov check script

richardwxn:
Forked from codecov scripts from istio repo and did some minor changes. Currently there is no test available so the report generated would still be empty.


ostromart:
Thanks for looking into this - I suggest doing this in two steps: copy the files without modification and add any changes in a followup PR. There's code in the repo so you should get some results now.

richardwxn:
> Thanks for looking into this - I suggest doing this in two steps: copy the files without modification and add any changes in a followup PR. There's code in the repo so you should get some results now.

yes, now we are able to get some results already. Actually I already exclude some changes from this PR, so it only contains necessary change such as path, filenames etc.

ostromart:
sg. would you mind splitting this up into two PR's? first would be just to copy scripts unmodified from istio. second would be to patch in any changes. this will create a record of any changes we've had to make.

richardwxn:
> sg. would you mind splitting this up into two PR's? first would be just to copy scripts unmodified from istio. second would be to patch in any changes. this will create a record of any changes we've had to make.

create another PR to just copy scripts unmodified #32

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/jstemmer/go-junit-report v0.0.0-20190106144839-af01ea7f8024 // indirect
 	github.com/kr/pretty v0.1.0
+	github.com/jstemmer/go-junit-report v0.0.0-20190106144839-af01ea7f8024 // indirect
```
ostromart: remove

go.sum
```diff
@@ -263,6 +263,8 @@ github.com/wadey/gocovmerge v0.0.0-20160331181800-b5bfa59ec0ad h1:W0LEBv82YCGEtc
 github.com/wadey/gocovmerge v0.0.0-20160331181800-b5bfa59ec0ad/go.mod h1:Hy8o65+MXnS6EwGElrSRjUzQDLXreJlzYLlWiHtt8hM=
 github.com/ugorji/go/codec v0.0.0-20181204163529-d75b2dcb6bc8/go.mod h1:VFNgLljTbGfSG7qAOspJ7OScBnGdDN/yBr0sguwnwf0=
 github.com/xordataexchange/crypt v0.0.3-0.20170626215501-b2862e3d0a77/go.mod h1:aYKd//L2LvnjZzWKhF00oedf4jCCReLcmhLdhm1A27Q=
+github.com/wadey/gocovmerge v0.0.0-20160331181800-b5bfa59ec0ad h1:W0LEBv82YCGEtcmPA3uNZBI33/qF//HAAs3MawDjRa0=
+github.com/wadey/gocovmerge v0.0.0-20160331181800-b5bfa59ec0ad/go.mod h1:Hy8o65+MXnS6EwGElrSRjUzQDLXreJlzYLlWiHtt8hM=
```
ostromart: duplicated?

scripts/codecov.sh
```diff
@@ -35,22 +35,14 @@ GO_TOP=$(cd "$(dirname "$0")"/../../../..; pwd)
 
 export GOPATH=${GOPATH:-$GO_TOP}
 
-if [ "${1:-}" != "" ]; then
+if [[ "${1:-}" != "" ]]; then
```
ostromart: should this be fixed also in istio/istio? 
that's what happens when you touch something ;-)

richardwxn: thanks for reminder, I would fix related problems in istio/istio

go.mod
richardwxn: done

PR #25, opened 2019-05-25
Add support for pkg/collateral to integrate with the doc pipeline.

geeknoid:
Also updates common files. This caused me to need to 
add missing copyright notices due to the new linter check.


BUGS-AND-FEATURE-REQUESTS.md
```diff
@@ -1,6 +1,6 @@
 # Bugs and Feature Requests
 
-You can report bugs and feature requests to the Istio team is one of three places:
+You can report bugs and feature requests to the Istio team in one of three places:
 
 - [Product Bugs and Feature Requests](https://github.com/istio/istio/issues)
```
ostromart: Should this be istio/operator rather than istio/istio?

geeknoid: We aggregate issues in istio/istio in general. That's why istio/operator doesn't even let you open issues at all.

The BUGS-AND-FEATURE-REQUESTS.md file is part of the common files from istio/common-files. It'll be at the root of every repo to tell casual readers where to go to file bugs.

PR #29, opened 2019-06-07
Initial implementation of helm renderer

ostromart:
Contains functions for a component based helm renderer. Eventually these functions can probably be consolidated under common utilities.
Requires https://github.com/istio/operator/pull/33 and https://github.com/istio/operator/pull/7.

PR #33, opened 2019-06-07
Recursive file watcher

ostromart:
Builds on fsnotify by adding recursive dir watching with debounce. 
Used in https://github.com/istio/operator/pull/29.


pkg/util/fswatch/fswatch.go
```diff
@@ -0,0 +1,129 @@
...
+func WatchDirRecursively(dir string) (<-chan struct{}, error) {
+	subdirs, err := recursiveSubdirs(dir)
+	if err != nil {
+		return nil, err
+	}
+
+	// Raw notifies that must be debounced.
+	notifyRaw := make(chan struct{})
+	for _, d := range subdirs {
+		watchDir(d, notifyRaw)
```
golangcibot: Error return value of `watchDir` is not checked (from `errcheck`)

pkg/util/fswatch/fswatch.go
```diff
@@ -0,0 +1,129 @@
...
+			case <-timer.C:
+				notify <- struct{}{}
+			}
+		}
+	}()
+
+	return notify, nil
+}
+
+func watchDir(dir string, notify chan<- struct{}) error {
```
golangcibot: watchDir - result 0 (error) is always nil (from `unparam`)

pkg/util/fswatch/fswatch.go
```diff
@@ -0,0 +1,129 @@
...
+							default:
+							}
+						}
+					case err, ok := <-watcher.Errors:
+						if !ok {
+							log.Printf("watcher channel closed for %s", dir)
+							done <- struct{}{}
+							return
+						}
+						log.Printf("error for watcher on %s:", dir, err)
```
golangcibot: printf: Printf call needs 1 arg but has 2 args (from `govet`)

pkg/util/fswatch/fswatch.go
```diff
@@ -0,0 +1,146 @@
...
+						if !ok {
+							log.Printf("watcher channel closed for %s", dir)
+							done <- struct{}{}
+							return
+						}
+						if event.Op&fsnotify.Write == fsnotify.Write {
+							log.Println("modified file:", event.Name)
+							select {
+							case notify <- struct{}{}:
+							default:
```
martonsereg: does this `select` block exist because sometimes multiple `WRITE` events can arrive for a single file change?
If yes, than a comment line explaining it would be great here because it's far from trivial.

ostromart: This is just avoiding blocking on sending the notify - added comment.

ostromart:
Thanks for the review and even more so testing it out and discovering the problem. 
What surprises me about your experience is that this should be a dir (rather than file) watcher so in principle if a new file is created, we should see fsnotify.Write event. I wonder if you could just make a simple mod on line 96 of also checking for delete and create events? Maybe the dir is just as OR over all file events under it. If so it would not be intuitive but at least I guess it will solve the problem you mentioned. 
If that doesn't fix it I think we can create an issue - it's perfect for someone else in the community to take on.

martonsereg:
It seems that an `fsnotify` watch on a dir is simply a watch on all of the files in that subtree.
So if a file is added it's not a `WRITE` on the directory, just a `CREATE` on the file itself. Same with `REMOVE`.
If we want to be notified about file creates and removes, than we should add those events as well.

pkg/util/fswatch/fswatch.go
```diff
@@ -0,0 +1,147 @@
...
+			go func() {
+				for {
+					select {
+					case event, ok := <-watcher.Events:
+						if !ok {
+							log.Printf("watcher channel closed for %s", dir)
+							done <- struct{}{}
+							return
+						}
+						if event.Op&fsnotify.Write == fsnotify.Write {
```
martonsereg:
```suggestion
						if event.Op&fsnotify.Write == fsnotify.Write || event.Op&fsnotify.Create == fsnotify.Create || event.Op&fsnotify.Remove == fsnotify.Remove {
```

PR #35, opened 2019-06-11
Initial implementation of fetcher of remote charts url

richardwxn:
1. implementation of url fetcher to download and verify charts, so it can point to remote/local charts during installation instead of relying on complied in ones.
2. dependent URL watcher/ Release channel integration
3. Add tests


pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,68 @@
...
+	return nil
+}
+
+// Run the polling job with target directory at specific interval
+func Run(udir string, interval uint64) {
+	po := &poller{
+		url:         udir,
+		minInterval: interval,
+	}
+	destDir, err := ioutil.TempDir("", "charts-")
```
ostromart: these sorts of string consts should be exposed in a const() block at the top with a name a comment.

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		log.Fatal("failed to create temp directory for charts")
+		return
+	}
+	uf := &URLFetcher{
+		url:        po.url + "/istio-installer.tar.gz",
+		verifyUrl:  po.url + "/istio-installer.tar.gz.sha256",
```
ostromart: same comment about string constants.

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,68 @@
...
+		untar:      true,
+		untarDir:   "untar",
+		verify:     true,
+		destDir:    destDir,
+		downloader: newFileDownloader(),
+	}
+	gocron.Every(interval).Minutes().Do(po.checkUpdate, uf)
+	<-gocron.Start()
+
+	defer os.RemoveAll(destDir)
```
ostromart: this doesn't look right. defer is used at the start of the function because it's called just before function returns for all paths. here, you may as well just call the function without defer.

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,155 @@
...
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"istio.io/istio/pkg/log"
```
ostromart: you should set up goimports to order these for you. generally it's
built-in packages

external packages

istio packages

with blank line between.

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,155 @@
...
+	"k8s.io/helm/pkg/chartutil"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+type FileDownloader struct {
+	// HTTP/HTTPS Client
```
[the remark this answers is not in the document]
richardwxn: thanks for the suggestions, I rephrased the comments based on the guideline

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,155 @@
...
+		if fi, err := os.Stat(ud); err != nil {
+			if err := os.MkdirAll(ud, 0755); err != nil {
+				return fmt.Errorf("failed to untar (mkdir): %s", err)
+			}
+
+		} else if !fi.IsDir() {
+			return fmt.Errorf("failed to untar: %s is not a directory", ud)
+		}
+
+		return chartutil.ExpandFile(ud, saved)
```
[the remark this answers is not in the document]
richardwxn: I would just do a normal untar operation then. Actually I think it may depend on what format we expect when we render manifest based on local charts.

pkg/helm/urlfetcher.go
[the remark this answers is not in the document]
richardwxn: I addressed your other comments directly and also replaced the gocron job with a normal goroutine work

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,155 @@
...
+	"istio.io/istio/pkg/log"
+	"k8s.io/helm/pkg/chartutil"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+type FileDownloader struct {
```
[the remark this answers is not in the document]
richardwxn: Added comments

pkg/helm/urlwatcher.go
richardwxn: removed defer

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+	}
+}
+
+// Run the polling job with target directory at specific interval
+func Run(dirURL string, interval time.Duration) error {
+	po := &poller{
+		url:         dirURL,
+		minInterval: interval,
+		ticker: time.NewTicker(time.Minute * interval),
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
		ticker:      time.NewTicker(time.Minute * interval),
```

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,154 @@
...
+	// client is a HTTP/HTTPS client.
+	client *http.Client
+}
+
+// URLFetcher is used to fetch and manipulate charts from remote url
+type URLFetcher struct {
+	// url is url to download the charts
+	url string
+	// verifyUrl is url to download the verification file
+	verifyUrl string
```
golangcibot: struct field `verifyUrl` should be `verifyURL` (from `golint`)

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,154 @@
...
+// Send GET Request
+func (c *FileDownloader) Get(href string) (*bytes.Buffer, error) {
+	buf := bytes.NewBuffer(nil)
+
+	req, err := http.NewRequest("GET", href, nil)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := c.client.Do(req)
+	defer resp.Body.Close()
```
golangcibot: httpresponse: using resp before checking for errors (from `govet`)

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,94 @@
...
+		p.existingHash = newHash
+		err := uf.fetchChart(shaF)
+		log.Errorf("Error fetching charts and verify: %v", err)
+		return err
+	}
+	return nil
+}
+
+func (p *poller) poll(uf *URLFetcher) {
+	for {
```
golangcibot: S1000: should use for range instead of for { select {} } (from `gosimple`)

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,154 @@
...
+	// url is url to download the charts
+	url string
+	// verifyURL is url to download the verification file
+	verifyURL string
+	// verify indicates whether the downloaded tar should be verified
+	verify bool
+	// destDir is path of charts downloaded to, empty as default to temp dir
+	destDir string
+	// untarDir is destination of untar
+	untarDir string
```
ostromart: Since users will probably not care where the charts actually are, I think we can combine the untar and dest dirs for simplicity.

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,154 @@
...
+		if !strings.EqualFold(hex.EncodeToString(sum), hash) {
+			return errors.New("checksum does not match")
+		}
+	}
+
+	// After verification, untar the chart into the requested directory.
+	return archiver.Unarchive(saved, f.untarDir)
+}
+
+// fetch sha file
```
ostromart: Optional: lint will complain about the comment format if the symbol is exported, but I recommend following the same pattern for most unexported symbols, especially functions. It leads to a more consistent style and is more idiomatic.

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,154 @@
...
+		return "", fmt.Errorf("SHA file url is empty")
+	}
+	shaF, err := f.downloader.DownloadTo(f.verifyURL, f.destDir)
+	if err != nil {
+		return "", err
+	}
+	return shaF, nil
+}
+
+func newFileDownloader() *FileDownloader {
```
ostromart: Seems unusual that FileDownloader is public but the new is not. Maybe FileDownloader should be private?

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,154 @@
...
+		client: &http.Client{
+			Transport: &http.Transport{
+				DisableCompression: true,
+				Proxy:              http.ProxyFromEnvironment,
+			}},
+	}
+}
+
+// Send GET Request
+func (c *FileDownloader) Get(href string) (*bytes.Buffer, error) {
```
ostromart: SendGET would be more meaningful to readers when this function is used outside the package.

pkg/helm/urlfetcher_test.go
```diff
@@ -0,0 +1,140 @@
...
+			chartsFileName: InstallationChartsFileName,
+			expectFile:     "./istio-installer.tar.gz",
+			verify:         false,
+		},
+		{
+			name:           "Charts download and verify",
+			chartsFileName: InstallationChartsFileName,
+			shaFileName:    InstallationShaFileName,
+			expectFile:     "./istio-installer.tar.gz",
+			verify:         true,
```
ostromart: Should add verify failure case also.

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,154 @@
...
+	// client is a HTTP/HTTPS client.
+	client *http.Client
+}
+
+// URLFetcher is used to fetch and manipulate charts from remote url
+type URLFetcher struct {
+	// url is url to download the charts
+	url string
+	// verifyURL is url to download the verification file
+	verifyURL string
```
ostromart: This may also be supplied from the caller through the API (from IstioControlPlane CR) so we should add that as an option, maybe when the fetcher is created.

pkg/helm/urlfetcher.go
richardwxn: done

pkg/helm/urlfetcher.go
richardwxn: done

pkg/helm/urlfetcher.go
richardwxn: done

pkg/helm/urlfetcher.go
richardwxn: done

pkg/helm/urlfetcher_test.go
richardwxn: added

pkg/helm/urlfetcher.go
richardwxn: Can you clarify this point?

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	shaF, err := f.downloader.DownloadTo(f.verifyURL, f.destDir)
+	if err != nil {
+		return "", err
+	}
+	return shaF, nil
+}
+
+// NewFileDownloader create a wrapper for download files
+func NewFileDownloader() *fileDownloader {
```
golangcibot: exported func NewFileDownloader returns unexported type *helm.fileDownloader, which can be annoying to use (from `golint`)

pkg/helm/urlfetcher.go
ostromart: I don't see a public API, only private, so how would a caller create a fetcher for a particular URL?

pkg/manifest/objects.go
```diff
@@ -59,7 +59,7 @@ func NewK8sObject(u *unstructured.Unstructured, json, yaml []byte) *K8sObject {
 		json:   json,
 		yaml:   yaml,
 	}
-
+	helm.R
```
golangcibot: undeclared name: `helm` (from `typecheck`)

ostromart:
Per offline conversation dir structure should be [VERSION.txt, charts/, profiles/] where VERSION.txt will contain YAML that specified the charts version, and the compatible versions of the operator binary specified in a range format e.g. https://github.com/hashicorp/go-version

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,165 @@
...
+	downloader *FileDownloader
+}
+
+// NewURLFetcher creates an URLFetcher pointing to urls and destination
+func NewURLFetcher(repoURL string, destDir string) *URLFetcher {
+	uf := &URLFetcher{
+		url:        repoURL + "/" + InstallationChartsFileName,
+		verifyURL:  repoURL + "/" + InstallationShaFileName,
+		verify:     true,
+		destDir:    destDir,
```
ostromart: We want this to default to a platform specific temp dir - however, not one that's randomly generated, since this will lead to downloading the same package multiple times.

richardwxn: If we run the fetcher by poller, during the poller running time, the destDir does not change, which is under platform default temp dir. Do you mean that even across different poller the destDir name should be fixed?

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,203 @@
...
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"istio.io/operator/pkg/util"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,101 @@
...
+func (p *poller) poll(uf *URLFetcher) {
+	for t := range p.ticker.C {
+		// When the ticker fires
+		fmt.Println("Tick at", t)
+		err := p.checkUpdate(uf)
+		log.Errorf("Error polling charts: %v", err)
+	}
+}
+
+// PollURL runs the polling job with target directory at specific interval
```
ostromart: PollURL continuously polls the given url, which points to a directory containing an installation package at the given interval and fetches a new copy if it is updated.

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,101 @@
...
+func PollURL(dirURL string, interval time.Duration) error {
+	po := &poller{
+		url:         dirURL,
+		minInterval: interval,
+		ticker:      time.NewTicker(time.Minute * interval),
+	}
+	destDir, err := ioutil.TempDir("", ChartsTempFilePrefix)
+	if err != nil {
+		log.Fatal("failed to create temp directory for charts")
```
ostromart: Don't log.Fatal - this will cause program to exit.

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,203 @@
...
+	return &FileDownloader{
+		client: &http.Client{
+			Transport: &http.Transport{
+				DisableCompression: true,
+				Proxy:              http.ProxyFromEnvironment,
+			}},
+	}
+}
+
+// SendGet makes http GET request to url
```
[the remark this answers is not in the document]
richardwxn: href is actually the url passed in, renamed it to url to avoid confusion

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,101 @@
...
+
+const (
+	// InstallationChartsFileName is sample Chart filename to fetch
+	InstallationChartsFileName = "istio-installer.tar.gz"
+	// InstallationShaFileName is Sha filename to verify
+	InstallationShaFileName = "istio-installer.tar.gz.sha256"
+	// VersionsFileName is version file's name
+	VersionsFileName = "VERSIONS.yaml"
+	// ChartsTempFilePrefix is temporary Files prefix
+	ChartsTempFilePrefix = "charts-"
```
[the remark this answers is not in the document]
richardwxn: done

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,101 @@
...
+		return fmt.Errorf("failed to read sha file: %s", err)
+	}
+	// Original sha file name is formatted with "HashValue filename"
+	newHash := strings.Fields(string(hashAll))[0]
+	// compare with existing hash
+	// if sha file updated then fetch latest charts
+	if !strings.EqualFold(newHash, p.existingHash) {
+		p.existingHash = newHash
+		err := uf.fetchChart(shaF)
+		log.Errorf("Error fetching charts and verify: %v", err)
```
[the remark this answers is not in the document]
richardwxn: done

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,101 @@
...
+		return err
+	}
+	return nil
+}
+
+func (p *poller) poll(uf *URLFetcher) {
```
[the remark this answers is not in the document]
richardwxn: make it an independent function to be more readable and also we might have more sophisticated polling logic in the future

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,101 @@
...
+	for t := range p.ticker.C {
+		// When the ticker fires
+		fmt.Println("Tick at", t)
+		err := p.checkUpdate(uf)
+		log.Errorf("Error polling charts: %v", err)
```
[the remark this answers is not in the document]
richardwxn: done

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"istio.io/pkg/log"
+)
+
+// poller is used to poll files from remote url at specific internal
+type poller struct {
+	// url is remote target url to poll from
+	url string
+	// minInterval is time intervals in minutes for polling
+	minInterval time.Duration
```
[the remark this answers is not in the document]
richardwxn: removed the minInterval as it is not necessary, just to keep a ticker is enough

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,101 @@
...
+	InstallationShaFileName = "istio-installer.tar.gz.sha256"
+	// VersionsFileName is version file's name
+	VersionsFileName = "VERSIONS.yaml"
+	// ChartsTempFilePrefix is temporary Files prefix
+	ChartsTempFilePrefix = "charts-"
+)
+
+// checkUpdate checks whether sha file being updated or not
+// fetch latest charts if yes
+func (p *poller) checkUpdate(uf *URLFetcher) error {
```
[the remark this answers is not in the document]
richardwxn: originally I put that as part of poller, but later I make them more independent because I think poller can be used with some other logics as well, not necessarily the URLfetcher

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,98 @@
...
+	for t := range p.ticker.C {
+		// When the ticker fires
+		log.Debugf("Tick at", t)
+		if err := p.checkUpdate(uf); err != nil {
+			log.Errorf("Error polling charts: %v", err)
+		}
+	}
+}
+
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

pkg/helm/urlfetcher.go
ostromart: For a poller we are ok, since the process stays alive. The problem is the non-poll mode - if a user runs a CLI command multiple times that targets URL, each time the command will download a fresh copy of the same install package.

pkg/helm/urlwatcher.go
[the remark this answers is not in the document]
ostromart: then, you could redesign the poller New function to take the new fetcher. otherwise you're forcing the user to maintain two separate objects that belong together.
also, unless you have a specific scenario in mind that looks likely it's best to follow YAGNI.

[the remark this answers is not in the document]
richardwxn: done

pkg/helm/urlfetcher.go
richardwxn: Actually the NewURLFetcher accept destDir passed in, so the caller can specify whether it is a random path or a fix path. Right now the poller would pass in a random path. To make it more clear, add a check in NewURLFetcher to use a fix path under platform specific temp dir when destDir passed in is empty.

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,105 @@
...
+	for t := range p.ticker.C {
+		// When the ticker fires
+		log.Debugf("Tick at: %s", t)
+		if err := p.checkUpdate(); err != nil {
+			log.Errorf("Error polling charts: %v", err)
+		}
+	}
+}
+
+func NewPoller(dirURL string, destDir string, interval time.Duration) *poller{
```
golangcibot: exported func NewPoller returns unexported type *helm.poller, which can be annoying to use (from `golint`)

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,105 @@
...
+	uf := NewURLFetcher(dirURL, destDir, InstallationChartsFileName, InstallationShaFileName)
+	return &poller{
+		url:         dirURL,
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/helm/urlfetcher.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	name := filepath.Base(u.Path)
+	destFile := filepath.Join(dest, name)
+	if err := ioutil.WriteFile(destFile, data.Bytes(), 0644); err != nil {
+		return destFile, err
+	}
+
+	return destFile, nil
+}
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
}
```

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,106 @@
...
+		log.Error("failed to create temp directory for charts")
+		return err
+	}
+
+	po := NewPoller(dirURL, destDir, interval)
+	go po.poll()
+
+	os.RemoveAll(destDir)
+	return nil
+}
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
}
```

pkg/helm/urlfetcher_test.go
```diff
@@ -0,0 +1,162 @@
...
+				return
+			}
+		}
+		ef := filepath.Join(fq.destDir, test.chartsFileName)
+		if _, err := os.Stat(ef); err != nil {
+			t.Error(err)
+			return
+		}
+	}
+}
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
}
```

pkg/helm/urlwatcher.go
```diff
@@ -0,0 +1,106 @@
...
+	destDir, err := ioutil.TempDir("", ChartsTempFilePrefix)
+	if err != nil {
+		log.Error("failed to create temp directory for charts")
+		return err
+	}
+
+	po := NewPoller(dirURL, destDir, interval)
+	go po.poll()
+
+	os.RemoveAll(destDir)
```
ostromart: I think you have a problem here - destDir will be removed while poller is still running. Either add a waitgroup or make life easier and just put the poller in a sync loop here. There's really no need for this to  be a goroutine.

richardwxn: good catch, I will just move it to main thread

ostromart (approved):

pkg/manifest/objects.go
```diff
@@ -60,7 +60,7 @@ func NewK8sObject(u *unstructured.Unstructured, json, yaml []byte) *K8sObject {
 		json:   json,
 		yaml:   yaml,
 	}
-
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

PR #42, opened 2019-06-21
Refactoring utils v2

martonsereg:
Cleanup of utils with various functions moved out of other packages.

In addition to @ostromart's changes:
- util functions that were declared in multiple files were cleaned up
- copyright and licence headers were added
- deleted `common.go` and moved path related util functions to `path.go`, others to `util.go` because both `util.go` and `common.go` was used to collect uncategorised util functions
- added some missing comments on util functions


pkg/util/util.go
```diff
@@ -0,0 +1,159 @@
...
+	bjb, err := yaml.YAMLToJSON([]byte(b))
+	if err != nil {
+		dbgPrint("bad YAML in isYAMLEqual:\n%s", b)
+		return false
+	}
+
+	return string(ajb) == string(bjb)
+}
+
+// YAMLDiff returns the diff between two YAML string representations.
```
richardwxn: In IsYAMLEqual function it do comparison by converting them to JSON first but in YAMLDiff it unmarshall to map[string]interface{} first and marshal then compare. What is the reason for difference behind them?

PR #49, opened 2019-06-26
Make README.md more inviting to contributors

sdake:
The README.md previously stated non CODEOWNERS could not contribute. This
is in violation of the spirit of the Istio project. Correct that problem by
making the README.md more inviting.

Our goal is to build a team. The code is a byproduct of that effort.

The MVP was a little difficult to parse and used an ordered list. Simplify the
wording of the MVP such that it is a little more vague, but a little less wordy.
Finally remove the ordering of the MVP feature set since all requirements are
of equivalent value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,8 @@
 
 The main purpose of this repository is to consume the micro-manifests from the
 [istio/installer](https://github.com/istio/installer) repository and generate validated
-Kubernetes manifests. The micro-manifests contain various steps or logical groupings of
-`gotpl` templated manifests. The operator's main job is to order and validate the generation
-or rollout of these micro-manifests into a functional working deployment of Istio.
+Kubernetes manifests that are used internally to present a daemonized mode for the operator
+or a command line interface for the operator.
 
 This repository is under construction and may generate frustration for developers or SREs
 consuming or developing this code base. If you find a problem or require an enhancement, please
@@ -23,14 +22,17 @@ Even better, get involved in Operator development work and
 
 ## Initial Goals
 
-The MVP for the operator is:
+The initial goals/MVP for the operator are:
 
-- Provide a mode of operation that uses a controller to install generated manifests.
-- Provide a mode of operation that generates a Kubernetes manifest.
-- Provide a structured API to eventually replace the de-facto `values.yaml` API.
-- Provide an API aligned with an Istio a-la-carte building blocks philosophy.
-- Provide a functional structure.
-- Provide a mechanism for manifests to be customized that survive upgrades.
-- Provide in-place upgrades.
-- Provide canary rollout upgrades.
-- Provide validation of all parameters.
+1.   Replace helm tiller as the recommended install upgrade controller. We will continue to use helm templates and
+the helm library to generate manifests under the hood, but will take ownership of the logic behind the install API.
+The operator will have modes of operation analogous to tiller - controller in and out of cluster and one-shot manifest
+generation for use with external toolchains.
+1.   Provide a structured API to eventually replace the de-facto values.yaml API. This API will be partially moved
+to CRDs used to configure Istio components directly, hence the operator API will only contain the portions that will 
+remain in the operator CRD, and continue to use values.yaml for the parameters that will be moved to component CRDs.
+The API will be aligned with Istio a-la-carte/building blocks philosophy and have a functional structure.
+1.   Add an overlay mechanism to allow users to customize the rendered manifest in a way that survives upgrades. 
+1.   Support upgrade, both in-place and dual control plane version rollover. The latter will be supported through the 
+use of the [istio/installer](https://github.com/istio/installer) charts.
+1.   Provide validation of all parameters.
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 # Istio Operator
 
-## Oerview
+## Overview
 
 The main purpose of this repository is to consume the micro-manifests from the
 [istio/installer](https://github.com/istio/installer) repository and generate validated
--- a/README.md
+++ b/README.md
@@ -5,23 +5,32 @@
 
 # Istio Operator
 
-Update 4/22/19
+## Oerview
 
-This repo is for the initial, alpha version of the official Istio install operator. It's currently only open to the core 
-design team for commits, while we merge prototypes that evolved in parallel. Once we have
-a basic functional skeleton in place we'll open the repo up for community contributions.
+The main purpose of this repository is to consume the micro-manifests from the
+[istio/installer](https://github.com/istio/installer) repository and generate validated
+Kubernetes manifests. The micro-manifests contain various steps or logical groupings of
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,23 +5,32 @@
...
-generation for use with external toolchains.
-1.   Provide a structured API to eventually replace the de-facto values.yaml API. This API will be partially moved
-to CRDs used to configure Istio components directly, hence the operator API will only contain the portions that will 
-remain in the operator CRD, and continue to use values.yaml for the parameters that will be moved to component CRDs.
-The API will be aligned with Istio a-la-carte/building blocks philosophy and have a functional structure.
-1.   Add an overlay mechanism to allow users to customize the rendered manifest in a way that survives upgrades. 
-1.   Support upgrade, both in-place and dual control plane version rollover. The latter will be supported through the 
-use of the [istio/installer](https://github.com/istio/installer) charts.
-1.   Provide validation of all parameters.
+## Oerview
```
brian-avery: Overview

README.md
```diff
@@ -5,23 +5,32 @@
...
+
+- Provide a mode of operation that uses a controller to install generated manifests.
+- Provide a mode of operation that generates a Kubernetes manifest.
+- Provide a structured API to eventually replace the de-facto `values.yaml` API.
+- Provide an API aligned with an Istio a-la-carte building blocks philosophy.
+- Provide a functional structure.
+- Provide a mechanism for manifests to be customized that survive upgrades.
+- Provide in-place upgrades.
+- Provide canary rollout upgrades.
+- Provide validation of all parameters.
```
linsun: - Provide various validated profile configurations through operators, including simple get started configuration.

README.md
```diff
@@ -5,23 +5,32 @@
...
+The MVP for the operator is:
+
+- Provide a mode of operation that uses a controller to install generated manifests.
+- Provide a mode of operation that generates a Kubernetes manifest.
+- Provide a structured API to eventually replace the de-facto `values.yaml` API.
+- Provide an API aligned with an Istio a-la-carte building blocks philosophy.
+- Provide a functional structure.
+- Provide a mechanism for manifests to be customized that survive upgrades.
+- Provide in-place upgrades.
+- Provide canary rollout upgrades.
```
linsun: Is the operator supporting multiple control planes?

README.md
```diff
@@ -5,23 +5,32 @@
...
+[file an issue](https://github.com/istio/operator/blob/master/BUGS-AND-FEATURE-REQUESTS.md)
+in Istio's centralized Issue trackers.
+
+Even better, get involved in Operator development work and
+[contribute](https://github.com/istio/operator/blob/master/CONTRIBUTING.md)!
+
+## Initial Goals
+
+The MVP for the operator is:
+
```
linsun: Could we write the MVP from user's perspective, ideally with roles associated with the user?

sdake: I think that makes sense as a followup.  I simply wanted to clean up the existing `README.md`.  I was hoping for a quick merge without changing the terms of the MVP - and leave the changing of the terms of the MVP (if there are any) to further environments wg debate.

Cheers
-steve

README.md
sdake: that was in the original MVP terms sheet.  in-place and canary.  This PR does not change the spirit or objectives of the MVP.  Instead the language is clarified in simpler and more concise wording.  Happy to iterate over multiple PRs.

README.md
sdake: this would be an extension of the MVP, and I agree highly valued.  Something we should discuss in an Environments workgroup meeting.  Rather than change the MVP goal-posts in this PR, I'd prefer community discussion on the topic.

README.md
sdake: thanks for the catch.

README.md
```diff
@@ -5,23 +5,32 @@
...
-use of the [istio/installer](https://github.com/istio/installer) charts.
-1.   Provide validation of all parameters.
+## Overview
+
+The main purpose of this repository is to consume the micro-manifests from the
+[istio/installer](https://github.com/istio/installer) repository and generate validated
+Kubernetes manifests. The micro-manifests contain various steps or logical groupings of
+`gotpl` templated manifests. The operator's main job is to order and validate the generation
+or rollout of these micro-manifests into a functional working deployment of Istio.
+
```
rcernich: I think this is too detailed and also over simplifies what the operator should be doing.  The operator should be fully managing an Istio control plane instance.  Rendering of templates and installation of manifests is really an implementation detail and, in my opinion is not helpful to anybody other than somebody wanting to know what's happening behind the scenes.

README.md
rcernich: I think we need to be careful about supporting multiple control planes.  I don't believe Istio itself supports multiple installs, as some resources are cluster wide, implying that multiple installs would share those resources (e.g. MeshPolicy and ClusterRbacConfig).  Furthermore, the watchers are cluster wide, so...

README.md
rcernich: Should we consider API and validation as part of the MVP at all?  I think having a simpler install experience, provided by the operator is a good start and doesn't preclude this in the future.

README.md
sdake: ok - I was simply keeping content in the old document and rewording so it was more clear.  I agree the old README.md was too detailed - and this emerged from that.

sdake:
I agree re the MVP - the goals are not clear.  I simply rewrote the existing goals in a more concise format.  We should have further discussion about what the MVP is, but I didn't change anything that wasn't already in the doc :)

ostromart:
Thanks for taking a stab at this @sdake. A couple of things:
- could we hold this off for a couple more days while I wrap up the outstanding PR's? They are getting close and I was able to merge several today. I think having those in there will create a much more meaningful starting point for other contributors.
- I'm not sure if there's much value anymore in even mentioning the MVP. We can give an overview of the operator goals and architecture and leave it at that? I'm not sure we have to publicly announce whatever decision we reach in the WG.

sdake:
> Thanks for taking a stab at this @sdake. A couple of things:
> 
> * could we hold this off for a couple more days while I wrap up the outstanding PR's? They are getting close and I was able to merge several today. I think having those in there will create a much more meaningful starting point for other contributors.
> * I'm not sure if there's much value anymore in even mentioning the MVP. We can give an overview of the operator goals and architecture and leave it at that? I'm not sure we have to publicly announce whatever decision we reach in the WG.

sounds reasonable.  I think it makes sense to let the folks that know the current arch and implementation provide the architecture overview and feature set.  I'll hold off on the PR until sometime early next week if that is sufficient time.

Cheers
-steve

rcernich:
> I agree re the MVP - the goals are not clear. I simply rewrote the existing goals in a more concise format. We should have further discussion about what the MVP is, but I didn't change anything that wasn't already in the doc :)

makes sense

ostromart:
PTAL at https://github.com/istio/operator/pull/61. 
If we're happy with that, let's close this.

PR #50, opened 2019-06-26
Add prow infrastructure

sdake:
This initial take on the prow infrastructure is targeted at
linting to begin with.

### DIFF
--- a/prow/lib.sh
+++ b/prow/lib.sh
@@ -1,0 +1,161 @@
+#!/bin/bash
+
+# Copyright 2018 Istio Authors
+#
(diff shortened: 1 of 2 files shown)


prow/lib.sh
```diff
@@ -0,0 +1,161 @@
+#!/bin/bash
```
howardjohn: in the future we should probably put this in common repo. Lets get the prow update done first though in case it changes

sdake: i was thinking of doing that before proceeding with this work but had same reaction that should go in after the prow update because it will probqably end in a rewrite of this library.

PR #60, opened 2019-06-28
move util to common pkg.

morvencao:


pkg/util/label.go
```diff
@@ -0,0 +1,79 @@
...
+func GetLabel(resource runtime.Object, label string) (value string, ok bool) {
+	labels, err := meta.NewAccessor().Labels(resource)
+	if err != nil {
+		// if we can't access labels, then it doesn't have this label
+		return
+	}
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	value, ok = labels[label]
```
ostromart: This doesn't seem right - we create a new labels and then check to see if our label matches? Suggest returning false if labels doesn't exist. You could just call HasLabel first to avoid code duplication.

pkg/util/label.go
```diff
@@ -0,0 +1,79 @@
...
+	if labels == nil {
+		return false
+	}
+	_, ok := labels[label]
+	return ok
+}
+
+// DeleteLabel is a helper function which deletes the specified label from the specified object.
+func DeleteLabel(resource runtime.Object, label string) {
+	resourceAccessor, err := meta.Accessor(resource)
```
ostromart: I think this could be the same as in HasLabel for consistency.

pkg/util/label.go
```diff
@@ -0,0 +1,79 @@
...
+package util
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// HasLabel is a helper function returning true if the specified object contains the specified label.
+func HasLabel(resource runtime.Object, label string) bool {
+	labels, err := meta.NewAccessor().Labels(resource)
```
ostromart: NewAccessor is cheap but still it's probably better to define it as a local package var and call that.

pkg/util/label.go
```diff
@@ -0,0 +1,79 @@
...
+		delete(labels, label)
+		resourceAccessor.SetLabels(labels)
+	}
+}
+
+// GetLabel is a helper function which returns the value of the specified label on the specified object.
+// returns ok=false if the label was not found on the object.
+func GetLabel(resource runtime.Object, label string) (value string, ok bool) {
+	if HasLabel(resource, label) {
+		labels := resourceAccessor.Labels(resource)
```
golangcibot: cannot initialize 1 variables with 2 values (from `typecheck`)

pkg/util/label.go
```diff
@@ -0,0 +1,79 @@
...
+// DeleteLabel is a helper function which deletes the specified label from the specified object.
+func DeleteLabel(resource runtime.Object, label string) {
+	resourceAccessor, err := meta.Accessor(resource)
+	if err != nil {
+		// if we can't access labels, then it doesn't have this label, so nothing to delete
+		return
+	}
+
+	if HasLabel(resource, label) {
+		labels := resourceAccessor.Labels(resource)
```
golangcibot: resourceAccessor.Labels undefined (type v1.Object has no field or method Labels) (from `typecheck`)

pkg/util/label.go
```diff
@@ -0,0 +1,79 @@
...
+// limitations under the License.
+
+package util
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var	(
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
var (
```

pkg/util/label.go
```diff
@@ -0,0 +1,79 @@
...
+		delete(labels, label)
+		resourceAccessor.SetLabels(labels)
+	}
+}
+
+// GetLabel is a helper function which returns the value of the specified label on the specified object.
+// returns ok=false if the label was not found on the object.
+func GetLabel(resource runtime.Object, label string) (value string, ok bool) {
+	if HasLabel(resource, label) {
+		labels := accessor.Labels(resource)
```
golangcibot: cannot initialize 1 variables with 2 values (from `typecheck`)

pkg/util/label.go
```diff
@@ -0,0 +1,79 @@
...
+
+	if HasLabel(resource, label) {
+		labels, _ := accessor.Labels(resource)
+		delete(labels, label)
+		resourceAccessor.SetLabels(labels)
+	}
+}
+
+// GetLabel is a helper function which returns the value of the specified label on the specified object.
+// returns ok=false if the label was not found on the object.
```
ostromart: It returns false...

pkg/util/label.go
```diff
@@ -0,0 +1,79 @@
...
+func GetLabel(resource runtime.Object, label string) (value string, ok bool) {
+	if HasLabel(resource, label) {
+		labels, _ := accessor.Labels(resource)
+		value, ok = labels[label]
+	}
+	return
```
ostromart: Nit: labeling return values and doing a naked return is best avoided if possible. I suggest writing this as follows:
func GetLabel(resource runtime.Object, label string) (string, bool) {
	if HasLabel(resource, label) {
		labels, _ := accessor.Labels(resource)
		return labels[label]
	}
	return "", false

pkg/util/label.go
```diff
@@ -0,0 +1,79 @@
...
+		resourceAccessor.SetLabels(labels)
+	}
+}
+
+// GetLabel is a helper function which returns the value of the specified label on the specified object.
+// returns "" and false if the label was not found on the object.
+func GetLabel(resource runtime.Object, label string) (string, bool) {
+	if HasLabel(resource, label) {
+		labels, _ := accessor.Labels(resource)
+		return labels[label]
```
golangcibot: wrong number of return values (want 2, got 1) (from `typecheck`)

pkg/util/label.go
```diff
@@ -0,0 +1,80 @@
...
+		resourceAccessor.SetLabels(labels)
+	}
+}
+
+// GetLabel is a helper function which returns the value of the specified label on the specified object.
+// returns "", false if the label was not found on the object.
+func GetLabel(resource runtime.Object, label string) (string, bool) {
+	if HasLabel(resource, label) {
+		labels, _ := accessor.Labels(resource)
+		value, ok = labels[label]
```
golangcibot: undeclared name: `value` (from `typecheck`)

pkg/util/label.go
morvencao: Yep, got your point.

PR #66, opened 2019-07-03
Fix iop install

ostromart:
Fixes a number of issues with iop install:
- error condition was causing goroutines to block
- default namespace setting was not being verified
- output formatting has been cleaned up


pkg/name/name.go
```diff
@@ -67,87 +67,78 @@ const (
...
-		log.Error(err.Error())
-		return ""
+		return "", fmt.Errorf("error in Namepsace for feature=%s, component=%s: %s", featureName, componentName, err)
 
 	}
 	defaultNamespace, ok := defaultNamespaceI.(string)
 	if !ok {
-		log.Errorf("CustomPackagePath has bad type %T, expect string", defaultNamespaceI)
-		return ""
+		return "", fmt.Errorf("DefaultNamespacePrefix has bad type %T, expect string", defaultNamespaceI)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

cmd/iop/shared.go
```diff
@@ -70,51 +71,59 @@ func genManifests(args *rootArgs) (name.ManifestMap, error) {
...
 	}
-	if errs := validate.CheckIstioControlPlaneSpec(mergedcps); len(errs) != 0 {
-		log.Fatalf(errs.ToError().Error())
+	if errs := validate.CheckIstioControlPlaneSpec(mergedcps, true); len(errs) != 0 {
+		return nil, errs.ToError()
 	}
 
 	if yd := util.YAMLDiff(mergedYAML, util.ToYAMLWithJSONPB(mergedcps)); yd != "" {
-		log.Fatalf("Validated YAML differs from input: \n%s", yd)
+		return nil, fmt.Errorf("Validated YAML differs from input: \n%s", yd)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

pkg/manifest/installer.go
```diff
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes/scheme"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
golangcibot: a blank import should be only in a main or test package, or have a comment justifying it (from `golint`)

pkg/validate/validate.go
```diff
@@ -20,6 +20,7 @@ import (
 	"reflect"
 	"strings"
 
+	"istio.io/fortio/log"
```
golangcibot: could not import istio.io/fortio/log (invalid package name: "") (from `typecheck`)

cmd/iop/shared.go
```diff
@@ -70,51 +71,59 @@ func genManifests(args *rootArgs) (name.ManifestMap, error) {
...
-		log.Fatalf(err.Error())
+		return nil, err
 	}
 
 	manifests, errs := cp.RenderManifest()
 
 	return manifests, errs.ToError()
 }
+
+func logAndPrintf(args *rootArgs, v ...interface{}) {
```
ayj: This capability should really be rolled into istio.io/pkg/log if its useful.

cmd/iop/install.go
```diff
@@ -46,10 +46,33 @@ func installManifests(args *rootArgs) {
...
+			logAndPrintf(args, "Error strings:\n%s\n", out.Stderr[cn])
+		}
+		if strings.TrimSpace(out.Stdout[cn]) != "" {
+			logAndPrintf(args, "Command output:\n%s\n", out.Stdout[cn])
+		}
 	}
 }
+
+// underline generates an underline with the underlineChar of length s.
+func underline(s, underlineChar string) string {
```
ayj: I'm not sure this warrants its own function. `strings.Repeat("=", len(cs))` is equally readable IMO.

data/charts/crds/templates/crds.yaml
```diff
@@ -1,3 +1,12 @@
+# To prevent accidental injection into istio-system.
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ .Release.Namespace }}
+  labels:
+    istio-operator-managed: Reconcile
+    istio-injection: disabled
+---
```
ayj: We should start namespaceing new labels, e.g. 

```yaml
operator.istio.io/managed: Reconcile
```

We'll need to migrate existing labels over time. xref https://github.com/istio/istio/issues/14777

pkg/manifest/installer.go
ayj: Maybe `import k8s.io/client-go/plugin/pkg/client/auth` to support all known client auth plugins? e.g. azure, oidc.

pkg/name/name.go
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

cmd/iop/shared.go
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

pkg/name/name.go
```diff
@@ -67,87 +67,78 @@ const (
...
+// IsComponentEnabledInSpec reports whether the given feature and component are enabled in the given spec. The logic is, in
 // order of evaluation:
 // 1. if the feature is not defined, the component is disabled, else
 // 2. if the feature is disabled, the component is disabled, else
 // 3. if the component is not defined, it is reported disabled, else
 // 4. if the component disabled, it is reported disabled, else
 // 5. the component is enabled.
 // This follows the logic description in IstioControlPlane proto.
-func IsComponentEnabled(featureName FeatureName, componentName ComponentName, installSpec *v1alpha2.IstioControlPlaneSpec) bool {
+func IsComponentEnabledInSpec(featureName FeatureName, componentName ComponentName, installSpec *v1alpha2.IstioControlPlaneSpec) (bool, error) {
```
ayj: `IsComponentEnabledInSpec` and `Namespace` functions mix user input validation and runtime behavior. Long term it would be good to separate these into distinct phases so we can catch bad input earlier in the pipeline.

data/charts/crds/templates/crds.yaml
ostromart: Done.

pkg/name/name.go
ostromart: Agreed and I had exactly the same feeling as I was working on this. I do feel uneasy about not double checking, but if the contract is clear the extra check can be removed. I've put a TODO here for now because validation is far from perfect right now so more work is needed before we can be sure the inputs are good.

pkg/manifest/installer.go
```diff
@@ -46,16 +49,38 @@ const (
...
+
+var (
+	// operatorLabelStr indicates Istio operator is managing this resource.
+	operatorLabelStr = name.ToOperatorNamespace("istio-operator-managed")
 	// istioComponentLabelStr indicates which Istio component a resource belongs to.
-	istioComponentLabelStr = "istio-component"
+	istioComponentLabelStr = name.ToOperatorNamespace("istio-component")
 	// istioVersionLabelStr indicates which Istio version
-	istioVersionLabelStr = "istio-version"
+	istioVersionLabelStr = name.ToOperatorNamespace("istio-version")
```
ayj: It took me a minute to understand what was going on here. This might be more readable to define the well-known labels a const, e.g.

```golang
OperatorLabelOperatorManaged = "operator.istio.io/managed" // valid values are ...
OperatorLabelOperatorComponent = "operator.istio.io/component" 
OperatorLabelOperatorVersion = "operator.istio.io/version" 
```

@geeknoid created some comment packages to catalog annotations (see https://github.com/istio/pkg/blob/master/annotations/annotation.go). It might be worth doing the same for labels so that are more easily discover-able via docs.

PR #68, opened 2019-07-04
add yaml separator for generated manifest.

morvencao:
Resolve: https://github.com/istio/istio/issues/15370

Currently if you generate the yaml manifests with default profile:

```
#./iop manifest > istio.yaml
```

You will find the yaml is invalid because there are some missing yaml separator:

```
# kubectl apply -f istio.yaml.old
error: error validating "istio.yaml.old": error validating data: [ValidationError(ClusterRole): unknown field "roleRef" in io.k8s.api.rbac.v1.ClusterRole, ValidationError(ClusterRole): unknown field "spec" in io.k8s.api.rbac.v1.ClusterRole, ValidationError(ClusterRole): unknown field "subjects" in io.k8s.api.rbac.v1.ClusterRole]; if you choose to ignore these errors, turn validation off with --validate=false
```

Check the generated yaml manifests:
```
...
          - weight: 2
            preference:
              matchExpressions:
              - key: beta.kubernetes.io/arch
                operator: In
                values:
                - s390x
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: istio-nodeagent-istio-system
  labels:
    app: istio-nodeagent
    release: istio
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: istio-nodeagent-istio-system
subjects:
  - kind: ServiceAccount
    name: istio-nodeagent-service-account
    namespace: istio-system
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: istio-nodeagent-istio-system
  labels:
    app: istio-nodeagent
    release: istio
rules:
- apiGroups: [""]
  resources: ["configmaps"]
  verbs: ["get"]
...
```

This PR is to add yaml separator for each yaml file rendered by the helm engine.

### DIFF
--- a/pkg/helm/helm.go
+++ b/pkg/helm/helm.go
@@ -119,6 +119,10 @@ func renderChart(namespace, baseValues, overlayValues string, chrt *chart.Chart)
 
 	var sb strings.Builder
 	for _, f := range files {
+		// add yaml separater if the rendered file doesn't have one at the end
+		if !strings.HasSuffix(strings.TrimSpace(f)+"\n", YAMLSeparator) {
+			f += YAMLSeparator
+		}
 		_, err := sb.WriteString(f + YAMLSeparator)
 		if err != nil {
 			return "", err
--- a/pkg/helm/helm.go
+++ b/pkg/helm/helm.go
@@ -119,7 +119,7 @@ func renderChart(namespace, baseValues, overlayValues string, chrt *chart.Chart)
 
 	var sb strings.Builder
 	for _, f := range files {
-		_, err := sb.WriteString(f)
+		_, err := sb.WriteString(f + YAMLSeparator)
 		if err != nil {
 			return "", err
 		}


pkg/helm/helm.go
```diff
@@ -115,7 +115,7 @@ func renderChart(namespace, baseValues, overlayValues string, chrt *chart.Chart)
 
 	var sb strings.Builder
 	for _, f := range files {
-		_, err := sb.WriteString(f)
+		_, err := sb.WriteString(f + YAMLSeparator)
```
howardjohn: is it ok if we get 2 separators?

morvencao: Have checked the default behavior of `helm template`, it will add additional YAMLSeparator between two resources, no matter whether the original template contains YAMLSeparator or not.

ostromart:
If we go down this route we should remove file ending separators which are in most of the helm templates. But really, I think it's easier just to add the missing separator to the offending template, since inserting these extras everywhere right now will lead to lots of null objects in the manifest.

pkg/helm/helm.go
```diff
@@ -115,7 +115,11 @@ func renderChart(namespace, baseValues, overlayValues string, chrt *chart.Chart)
 
 	var sb strings.Builder
 	for _, f := range files {
-		_, err := sb.WriteString(f)
+		// add yaml separater if the rendered file doesn't have one at the end
+		if !strings.HasSuffix(strings.TrimSpace(f) + "\n", YAMLSeparator) {
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
		if !strings.HasSuffix(strings.TrimSpace(f)+"\n", YAMLSeparator) {
```

morvencao:
Only add yaml separator if the rendered chart manifest doesn't have yaml separator at the end.

morvencao:
@ayj Sure

pkg/helm/helm.go
sdake: @howardjohn if K8s receives an empty blob in a manifest (triggerered by 2 seperators) it ignores it.  Its not tidy, and something that could be optimized, but not really worth the trouble at this time.

Cheers
-steve

howardjohn: it's not just about what kubernetes expects, we need to make sure all usages of this output handle it ok as well. I don't know where this is actually used, just wanted to point it out since we had issues in the test framework with extra separators

sdake: interesting re test infra.

howardjohn: yeah if I remember right it would try to do `kubectl apply nil` or something. Nothing that can't be fixed just need to he careful

morvencao:
Closing this for this is included in https://github.com/istio/operator/pull/128.

PR #70, opened 2019-07-06
add feature and components for gateway.

morvencao:
Resolve: https://github.com/istio/istio/issues/15376

- add gateway feature
- add ingressgateway and egressgateway components
- enable gateway feature in default profile


ostromart:
Thanks for the contribution, Morven. If you're interested in taking on larger tasks in the operator, or just want to listen in, we have a weekly meeting - I can send you an invite.

morvencao:
@ostromart Sure, thanks. But I 'm not sure if I can attend the meeting or not because of the timezone issue.

data/profiles/default.yaml
```diff
@@ -233,6 +233,54 @@ autoInjection:
           rewriteAppHTTPProbe: false
           selfSigned: false
 
+# Istio Gateway feature
+gateways:
+  enabled: true
+  components:
+    ingressGateway:
+    - gateway:
+        common:
```
ayj: Do the default gateway ports need to be added here? 

* https://github.com/istio/installer/blob/67b0877a322bca62cbd1ebd9e48a5f748b534507/gateways/istio-ingress/values.yaml#L22-L50
* https://github.com/istio/installer/blob/67b0877a322bca62cbd1ebd9e48a5f748b534507/gateways/istio-ingress/values.yaml#L98-L108
* https://github.com/istio/installer/blob/67b0877a322bca62cbd1ebd9e48a5f748b534507/gateways/istio-ingress/values.yaml#L123-L147

Also, should this include other advanced options like `ISTIO_META_ROUTER_MODE: "sni-dnat"`, zvpn?

morvencao: @ayj thanks, I have raised another PR: https://github.com/istio/operator/pull/76 to migrate other profiles(demo, demo-auth, minimal...) to operator and have applied your comments into that PR.
And in this PR, I will only add feature and components for gateways.

PR #71, opened 2019-07-07
To demo use of values_types.go as source of truth for value.yaml doc

richardwxn:
For [Create a single source of truth for values.yaml docs](https://github.com/istio/istio/issues/15249)
1. Extract the globalConfig part from the original Values_types.go file and added comment based on existing documentation as sample_values.types.go file

2. Use K8S code-generator tool https://github.com/kubernetes/code-generator/tree/master/cmd/go-to-protobuf to convert to generated.proto

3. Use [protoc-gen-docs](https://github.com/istio/tools/tree/master/protoc-gen-docs) to generate the viewable html doc from the generated.proto

Limitation:
The existing generator tools cannot handle map[string]interface{} type properly so have to manually patch those fields to proto.


ayj:
PR titles become part of git history. Can you re-title to match the functional/behavior change described in https://github.com/istio/istio/issues/15249? 

It looks like you ran several commands plus some manual fix-ups to generate these files. If these steps need to be re-run in the future it would be good to document them in a README adjacent to sample_values_types.go.

richardwxn:
Updated the PR title and added README

ostromart:
Could we see an example of the docs that are generated out of the proto? How will this look for examples and more lengthy descriptions?

geeknoid:
Sorry, I might be a broken record here, but  I have to push...

What's the benefit of using Go as the source of truth, rather than protos like we explicitly chose to do in the rest of Istio?

If you use protos as the source of truth, you  can use a marker type to represent interface{}. Depending on the use, then a simple SED replacement can turn that marker type into interface{} in the Go code  (similar to what we did in Mixer).

I'm continue to be worried that diverging from what we do in the rest of the platform will have a long-term conceptual and practical cost.

ostromart:
@geeknoid my reason for preferring Go struct over proto is that proto doesn't really represent the truth i.e. it would not be possible to unmarshal a YAML into a the proto due to the interface{} issue. I also feel a bit dirty using hacks to fix up the differences. With the go structs as the source of truth, the proto would not be published and I feel more justified in hacking to generate it, since the proto would be just for generating the docs and wouldn't be publicly visible (it's here in the PR just to show what it looks like, the PR not be merged.)
Also from a developer experience pov, it's easier to work with the structs because that's what's used in the code (renames can be done without recompiling for example.). We'd only involve the proto format temporarily when generating site docs.

ayj:
Is there a material difference between the Istio API (platform agnostic) and the APIs to manage the k8s specific manifestation of the control plane? My understanding was that the istio/operator was inherently k8s specific and the therefore the use of k8s-ism, e.g. 'go struct' schemas, kubebuilder, etc. was reasonable and self-consistent.

geeknoid:
@ostromart I understand the benefits you bring up, but I don't think that warrants the difference from everything else in Istio. We have folks that take dependencies on our protos already and I expect for more of that to happen given MCP. We're also building infra to do various static analysis on our CRDs, based on proto annotations.

@ayj I'm not interested in the implementation details here, but in how we define the customer-visible CRD and produce the customer-visible documentation. The rest of Istio CRDs are defined in proto form and everything derives from there. Unless there is a concrete practical reason that this doesn't work, I think we should be aligning things.

ostromart:
All right, apart from the issues I mentioned I don't have any objections to using proto as the source of truth. I've discussed with @richardwxn and we'll do the following: 
1. Transform the existing go structs to a proto.
2. Add compilation to go structs from that to the proto make target. 
3. Patch on any fixes we need as part of the make.
We can then proceed with the documentation work using the proto.

geeknoid:
Thank you.

And as I mentioned, we're also adding support for OpenAPI spec output for
our config protos. This will provide us with the ability to introduce
constraints which can be validated automatically by the API server. And Oz
will also be introducing cross-resource validation support in Galley  so
that we can ensure referential integrity. So being on the proto plan helps
with these things too.

ayj:
Is Galley planning to consume the operator CRD as part of cross-resource validation?

geeknoid:
There is a chicken-and-egg problem here of course, but Galley probably
should. The operator CRD is going to be fairly complex, so it stands to
reason that validation will be a good thing.

richardwxn:
Create a new PR #79 to use proto as source of truth based on discussion above

ayj:
100% agreed that operator API validation is important. I have strong reservations about Galley being responsible for that validation. We should probably follow-up in a separate doc/issue :)

ostromart:
Closing this, let's continue in https://github.com/istio/operator/pull/79.

PR #76, opened 2019-07-09
[WIP]migrate common profiles.

morvencao:
Resolve: https://github.com/istio/istio/issues/15330

Migrate real sds profile, and also minimal, demo, demo-auth profiles.

depends on https://github.com/istio/operator/pull/70 since we need to enable gateway for some compiled-in profiles. 

Also applied comments of https://github.com/istio/operator/pull/70#discussion_r301287667


data/profiles/demo-auth.yaml
```diff
@@ -0,0 +1,474 @@
...
+          resources:
+            requests:
+              cpu: 10m
+              memory: 100Mi
+          env:
+            - name: GODEBUG
+              value: "gctrace=1"
+        values:
+          image: mixer
+          autoscaleEnabled: false
```
ostromart: this looks incompatible with the hpaSpec above. Probably set it to true.

data/profiles/demo.yaml
```diff
@@ -0,0 +1,474 @@
...
+            periodSeconds: 30
+            timeoutSeconds: 5
+          resources:
+            requests:
+              cpu: 10m
+              memory: 100Mi
+
+        values:
+          image: pilot
+          autoscaleEnabled: fase
```
ostromart: typo, also should probably be true.

data/profiles/demo-auth.yaml
morvencao: Currently, autoscaler or hpa is disabled for demo and demo-auth profiles.
However, we can't specify whether to enable autoscaler in the new platform level installation API, so I have to add `autoscaleEnabled: false` to disable autoscaler for demo profile.

I thought we should remove `hpaSpec` in k8s section.

data/profiles/demo.yaml
morvencao: Fixed typo.

data/profiles/demo-auth.yaml
ostromart: Yes, in that case please remove hpaSpec.

data/profiles/demo-auth.yaml
```diff
@@ -0,0 +1,413 @@
...
+            istio-sidecar-injector-service-account.istio-remote: istio-sidecar-injector.istio-remote.svc
+            istio-sidecar-injector-service-account.istio-pilot11: istio-sidecar-injector.istio-pilot11.svc
+            istio-sidecar-injector-service-account.istio-control: istio-sidecar-injector.istio-control.svc
+            istio-sidecar-injector-service-account.istio-master: istio-sidecar-injector.istio-master.svc
+            istio-sidecar-injector-service-account.istio-control-master: istio-sidecar-injector.istio-control-master.svc
+            istio-galley-service-account.istio-pilot11: istio-galley.istio-pilot11.svc
+            istio-galley-service-account.istio-control: istio-galley.istio-control.svc
+            istio-galley-service-account.istio-master: istio-galley.istio-master.svc
+            istio-galley-service-account.istio-control-master: istio-galley.istio-control-master.svc
+            istio-galley-service-account.istio-config: istio-galley.istio-config.svc
```
ayj: These cert names correspond to the dual-control plane topology. I think Citadel includes defaults for the basic case already. We _might_ need `istio-pilot-service-account.istio-system:istio-pilot.istio-system` for parity with the istio/istio manifests.

data/profiles/demo.yaml
```diff
@@ -0,0 +1,413 @@
...
+    pilot:
+      common:
+        k8s:
+          env:
+          - name: GODEBUG
+            value: "gctrace=1"
+          replicaCount: 1
+          readinessProbe:
+            handler: {}
+# FIME: what's the problem here?
```
ayj: 1) s/FIME/TODO referencing user or github tracking issue.

2) Is this required or simplify copied from some existing file?

data/profiles/demo-auth.yaml
morvencao: Sure.

data/profiles/demo.yaml
morvencao: Done, this is related to https://github.com/istio/istio/issues/15366, still investigating on the solution.

data/profiles/demo-auth.yaml
morvencao: Done

ayj: This doesn't look right. For the default/demo installs in istio-system namespace I don't think any of these are needed. IIUC, these additional certs are required for the dual control plane install.

morvencao: Yep, removed other items and only added `istio-pilot-service-account.istio-system:istio-pilot.istio-system`.

PR #77, opened 2019-07-09
fix error sample manifests in readme.

morvencao:


README.md
```diff
@@ -237,9 +237,8 @@ profile. Here's an example of overriding some global level default values:
 ```yaml
 profile: sds
 values:
-  global:
-    logging:
-      level: "default:warning" # override from info
+  logging:
+    level: "default:warning" # override from info
```
richardwxn: I think this part was correct before?

ostromart: Yes, you're right. Thanks.

ostromart: I mean that Morven is right i.e. there should be no global.

PR #80, opened 2019-07-11
Remove vendor entirely

richardwxn:
As part of go module migration, clean up the vendor entirely. Tested locally. 
TODO: clean up makefile as some targets depend on vendor path.


ostromart:
lgtm, but please fix the affected makefile targets because this breaks proto generation.

ostromart:
Looks like circleci and prow environments also need to be updated.

richardwxn:
updated the prow and circleci part. will update the makefile in another PR

PR #89, opened 2019-07-16
fix broken probe in generated manifests.

morvencao:
Fix: https://github.com/istio/istio/issues/15366

Context: https://github.com/istio/istio/issues/15366#issuecomment-511711356


pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```diff
@@ -19,6 +19,7 @@ package v1alpha2;
 import "k8s.io/api/core/v1/generated.proto";
 import "k8s.io/api/autoscaling/v2beta1/generated.proto";
 import "k8s.io/apimachinery/pkg/apis/meta/v1/generated.proto";
+//import "k8s.io/apimachinery/pkg/util/intstr/generated.proto";
```
ostromart: Comment or remove.

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```diff
@@ -618,6 +620,46 @@ message Resources {
...
+    HTTPGetAction httpGet = 2;
+    TCPSocketAction tcpSocket = 3;
+    int32 initial_delaySeconds = 4;
+    int32 timeout_seconds = 5;
+    int32 period_seconds = 6;
+    int32 success_threshold = 7;
+    int32 failure_threshold = 8;
+}
+
+message ExecAction {
```
ostromart: Comment these, otherwise Go lint will complain later.

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```diff
@@ -618,6 +620,46 @@ message Resources {
...
+    repeated HTTPHeader httpHeaders = 5;
+}
+
+message HTTPHeader {
+    string name = 1;
+    string value = 2;
+}
+
+message TCPSocketAction {
+    //k8s.io.apimachinery.pkg.util.intstr.IntOrString port = 1;
```
ostromart: Per my comment in the issue, I think we should patch this in the structs manually. We really want to avoid any deviations from k8s schema.

morvencao: Sure, will do it.

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
morvencao: Done

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
morvencao: Added comments.

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
morvencao: Have extended the k8s `IntOrString` type and generated patch to the generated go code, also add some unit test cases in `validation_test.go` to make sure both the `port number` and `port name` can be marshal/unmarshal properly.

morvencao: Please check the updated patch file here: https://github.com/morvencao/operator/blob/br_fix_probe/pkg/apis/istio/v1alpha2/fixup_go_structs.patch

PR #93, opened 2019-07-18
Fix file path error of fix_values_structs patch.

morvencao:
The PR does the following two things:
- update makefile to generated the original go code without patch
- add generated go code to repo

---------------------------------------------
The workflows for proto to go are different based on the changes:

1) Only update proto file, without patch change:
```
make proto_with_setup # generate the go code and apply the patch
```

2) U pdate proto file and the patch for generated go code:
```
make proto_orig_with_setup # generate the go code and copy to orig
# make change to the generated go code
make gen_patch # update the patch
```


morvencao:
@ostromart Thanks for reminder, already based on the latest code.

morvencao:
@richardwxn It's used to generate the new patch if someone wants to update the patch.

sdake:
please force rebase your patch to remove circleci from the testingl.

richardwxn:
I updated the Makefile adding support for values_types.proto, can you update your change for that as well

morvencao:
@richardwxn Have rebased code, and fixed file path error of `fix_values_structs.patch`.

morvencao:
@richardwxn Have rebased code, and fixed file path error of `fix_values_structs.patch`.

PR #94, opened 2019-07-18
Remove circleci

sdake:
This test is redundant and serves no purpose.


ayj:
scripts/codecov_diff.sh also references circleci (see https://github.com/istio/operator/blob/d40f9e6fc1223cc35a8271c994d3e97385499cf6/scripts/codecov_diff.sh#L39-L71)

sdake:
thanks @ayj

richardwxn:
@sdake why we need to remove all code coverage scripts as well? I thought only the one supposed to run  on circleci need to be removed, the other one can be run locally

elfinhe:
+1 We'd better keep the coverage script, which can be run locally. And the coverage script is referenced by Makefile.

PR #98, opened 2019-07-18
Sync charts with istio/installer

ostromart:
Two way sync with istio/installer charts. After this it should be possible to reference the istio/installer charts directly without duplicating in this repo.
See https://github.com/istio/installer/pull/284 for sync in the other direction.


ostromart:
@morvencao no, the plan is to start pointing to release branches directly in the istio/installer repo.

PR #100, opened 2019-07-19
Work around lint problem

sdake:
The PATH is not properly setup by prow. Until that problem is solved
workaround the problem by setting the path directly in the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,7 @@ clean-proto:
 	rm -f $(api_pb_gos)
 
 lint:
+	export PATH=${PATH}:${GOPATH}/bin
 	@scripts/check_license.sh
 	@scripts/run_golangci.sh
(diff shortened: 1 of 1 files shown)


morvencao:
As a workaround, it makes sense.

IMO, the further fix should be in prow job template,  maybe we can add 
```
export PATH=$GOPATH/bin:$PATH && make lint
```
to `operator-lint-master ` job command here: https://github.com/istio/test-infra/blob/master/prow/cluster/jobs/istio/operator/istio.operator.master.yaml#L31-L34

sdake:
IMO it makes more sense to fix it in the groovy templates of prow.  I am not clear this groovy code is being executed.  I don't even know how these groovy templates are run.

This is simply to unblock the gate.

PR #104, opened 2019-07-20
Cleanup and changes based on CLI review

ostromart:


pkg/util/common.go
```diff
@@ -44,11 +39,5 @@ func (t Tree) String() string {
 
 // IsFilePath reports whether the given URL is a local file path.
 func IsFilePath(path string) bool {
```
richardwxn: Can you help to clarify? When we call helm.NewHelmRenderer, we pass in installation package path, which is a HTTP URL, then in NewHelmRender we determine to return render base on compiled in, file or HTTP. But there we have only two render, one is for local file one is for complied in. I assume we are going to integrate with the URLFetcher change there. If that's the case, the check here is not enough to make sure it must be a local file path, right?

pkg/validate/validate.go
```diff
@@ -157,20 +156,5 @@ func validateInstallPackagePath(path util.Path, val interface{}) util.Errors {
...
-			return nil
-		}
-	}
-
-	errStr := fmt.Sprintf("Invalid path prefix for %s: %s. \nMust be one of the following:\n", path, valStr)
-	for _, vp := range validPrefixes {
-		errStr += fmt.Sprintf("%-15s %s", vp, validPrefixDesc[vp])
-	}
-	return util.NewErrs(fmt.Errorf(errStr))
+	return nil
```
richardwxn: Not sure why we remove the validation for local file path? Do we still support local installation path

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```diff
@@ -143,20 +141,22 @@ spec:
...
+// The spec is a used to define a customization of the default profile values that are supplied with each Istio release.
 // It is grouped at the top level by feature, where behavior of Istio functional areas is specified.
 // Each feature contains components, where k8s resource level defaults can be overridden.
-// Because the spec is a customization API, specifying an empty InstallSpec results in a default Istio controlplane.
+// Because the spec is a customization API, specifying an empty InstallSpec results in a default Istio control plane.
 message IstioControlPlaneSpec {
-    // Prefix added to all namespaces for any installed component.
-    string default_namespace_prefix = 5;
+    // Default namespace if feature or component namespaces are not set.
+    string default_namespace = 5;
```
ayj: This is going back to the single namespace control plane?

morvencao:
Change sets need to rebase to pass the `operator-mandiff-master`

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
ostromart: No, this is just a default. I think the prefix thing doesn't make sense based on the names I've seen.

pkg/util/common.go
ostromart: Yes, that's correct. We have to use priority ordering at the caller with URL first, since there's really no pattern we can look for to distinguish a file path. One could even argue that what we have here is not strictly correct because you could have a file path that doesn't match.

pkg/validate/validate.go
ostromart: Yes, but we've removed the file:/// prefix.

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```diff
@@ -382,7 +384,7 @@ message GalleyComponentSpec {
 message CommonComponentSpec {
     // Selects whether this component is installed.
     google.protobuf.BoolValue enabled = 1;
-    // Namespace that all resources for this component are installed into. Override for parent feature namespace.
+    // Namespace that all resources for this component are installed into. Overrides for parent feature namespace.
```
ayj: re: Overrides, this is a single override for the component or multiple overrides?

ostromart: That was not clearly written. I think the new version should be more clear.

PR #105, opened 2019-07-20
Update mesh command layout per CLI review

ostromart:
This PR updates the mesh command layout per
https://docs.google.com/document/d/16SWW14zrwFKc9OzH6biMQD1u9INllzMa56jde9vgQkM/edit?ts=5d30fc6c#heading=h.z0zx9zymre83


cmd/mesh/manifest-migrate.go
```diff
@@ -0,0 +1,37 @@
...
+		Short: "Migrates a file containing Helm values to IstioControlPlane format.",
+		Long:  "The migrate subcommand is used to migrate a configuration in Helm values format to IstioControlPlane format.",
+		Args:  cobra.ExactArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			manifestMigrate(rootArgs)
+		}}
+
+}
+
+func manifestMigrate(args *rootArgs) {
```
golangcibot: `manifestMigrate` - `args` is unused (from `unparam`)

cmd/mesh/manifest.go
```diff
@@ -32,36 +17,42 @@ import (
...
+	mc.AddCommand(mgc)
+	mc.AddCommand(mdc)
+	mc.AddCommand(mac)
+	mc.AddCommand(mmc)
+
+	return mc
 }
 
-func genManifests(args *rootArgs) (name.ManifestMap, error) {
+func genManifests(args *rootArgs, inFilename string) (name.ManifestMap, error) {
```
golangcibot: `genManifests` - `args` is unused (from `unparam`)

cmd/mesh/profile-list.go
```diff
@@ -0,0 +1,48 @@
...
+		Short: "Lists available Istio configuration profiles.",
+		Long:  "The list subcommand is used to list available Istio configuration profiles.",
+		Args:  cobra.ExactArgs(0),
+		Run: func(cmd *cobra.Command, args []string) {
+			profileList(rootArgs, plArgs)
+		}}
+
+}
+
+func profileList(args *rootArgs, plArgs *profileListArgs) {
```
golangcibot: `profileList` - `plArgs` is unused (from `unparam`)

sdake:
from linter prow job:
```
/home/prow/go/src/istio.io/operator/cmd/mesh/profile.go missing license
/home/prow/go/src/istio.io/operator/cmd/mesh/profile.go missing Copyright
/home/prow/go/src/istio.io/operator/cmd/mesh/manifest.go missing license
/home/prow/go/src/istio.io/operator/cmd/mesh/manifest.go missing Copyright
```

cmd/mesh/profile-dump.go
```diff
@@ -0,0 +1,161 @@
...
+	// rootConfigNode sets the root node for the subtree to display the config for.
+	rootConfigNode string
+	// set is a string with element format "path=value" where path is an IstioControlPlane path and the value is a
+	// value to set the node at that path to.
+	set string
+}
+
+func addProfileDumpFlags(cmd *cobra.Command, args *profileDumpArgs) {
+	cmd.PersistentFlags().StringVarP(&args.inFilename, "filename", "f", "", filenameFlagHelpStr)
+	cmd.PersistentFlags().StringVarP(&args.rootConfigNode, "root-config-node", "r", "",
```
richardwxn: this naming is more from internal structure perspective, can we give it a more understandable name to user?

cmd/mesh/profile-dump.go
```diff
@@ -0,0 +1,161 @@
...
+		logAndFatalf(args, err.Error())
+	}
+
+	t, err := translate.NewTranslator(version.NewMinorVersion(1, 2))
+	if err != nil {
+		logAndFatalf(args, "%s", err)
+	}
+
+	if pdArgs.helmValues {
+		mergedYAML, err = component.CompileHelmValues(mergedcps, t, "")
```
richardwxn: although this one was merged already, but I think would it be better to rename that as TranslateToHelmValues

cmd/mesh/profile-list.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+type profileListArgs struct {
+	// set is a string with element format "path=value" where path is an IstioControlPlane path and the value is a
+	// value to set the node at that path to.
+	set string
```
richardwxn: This command is not yet implemented, just curious where will this flag be used? From the description, it looks like it would set the node at specific path with given value, then why would it be part of "list" command?I thought list command here is doing something immutable

ayj: Agreed that it is surprising to see a mutation flag on a list operation which is typically read-only.

cmd/mesh/profile-dump.go
ostromart: yeah, it's not the greatest but i had a hard time thinking of something better. any ideas?

cmd/mesh/manifest.go
```diff
@@ -0,0 +1,188 @@
...
+		overlayFilenameLog = "[Empty Filename]"
+	}
+
+	// Start with unmarshaling and validating the user CR (which is an overlay on the base profile).
+	overlayICPS := &v1alpha2.IstioControlPlaneSpec{}
+	if err := util.UnmarshalWithJSONPB(overlayYAML, overlayICPS); err != nil {
+		return nil, fmt.Errorf("could not unmarshal the overlay YAML from file: %s", overlayFilenameLog)
+	}
+	if errs := validate.CheckIstioControlPlaneSpec(overlayICPS, false); len(errs) != 0 {
+		return nil, fmt.Errorf("Overlay spec failed validation against IstioControlPlaneSpec: \n%v", overlayICPS)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

cmd/mesh/profile-dump.go
ostromart: Done.

cmd/mesh/profile-list.go
ostromart: This was a copy & past error - fixed.

cmd/mesh/profile.go
```diff
@@ -0,0 +1,46 @@
...
+
+	plc := profileListCmd(args)
+	pdc := profileDumpCmd(args, pdArgs)
+	pdfc := profileDiffCmd(args)
+
+	addFlags(plc, args)
+	addFlags(pdc, args)
+	addFlags(pdfc, args)
+
+	addProfileListFlags(plc, plArgs)
```
golangcibot: undeclared name: `addProfileListFlags` (from `typecheck`)

cmd/mesh/manifest.go
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

cmd/mesh/profile-dump.go
ayj: The analogous kubectl option is `jsonpath`. In lieu of adopting jsonpath syntax (has this been considered?) we could copy the name, e.g. `path`?

ostromart: I did consider path as a name but it seems too general. How about config-path?
The syntax here follows jsonpath to the extent that it makes sense - there are things that jsonpath allows that wouldn't be supported here so it would be misleading for users for us to say "use jsonpath" and not fully support that syntax. 
I think this is a reasonable compromise - close enough to jsonpath that it will be easy to understand but restrictive enough that all possible expressions are supported. 
One difference where we could be more jsonpath-like is list member selection (a.b.[key:value].c in this path syntax is a.b.[?(@.key==value)].c) but IMO the jsonpath syntax here is just horrible - I mean maybe it makes sense once you use jsonpath as a whole but not for this one case we have.

ayj: `config-path` is better. 

How often would users need to extra list members with `profile-dump` vs. a named subset (e.g. pilot config only). I suspect the latter would be the primary use case. Aligning with an existing scheme, even if its imperfect, does have advantages in that the user only has to learn one syntax.

ostromart: Probably never - in which case the syntax is the same as jsonpath, so there's nothing new for the user to learn. But the moment we actually call it jsonpath it creates an expectation from the user that they can do jsonpath type of things which we don't support. 
Anyway - I've changed to config-path. I'm open to revisiting the jsonpath question in the future since I do think there's a place for it in operator. Probably more work needed to figure out exactly how and where we could support it. Maybe that will be one of the feedbacks we'll get from users during the alpha.

ayj: sgtm

cmd/mesh/manifest.go
```diff
@@ -0,0 +1,130 @@
...
+	mc.AddCommand(mdc)
+	mc.AddCommand(mac)
+	mc.AddCommand(mmc)
+
+	return mc
+}
+
+//func genProfile(args *rootArgs, helmValues bool, inFilename, setOverlayYAML, configPath string) (string, error) {
+
+func genManifests(rootArgs *rootArgs, inFilename string, setOverlayYAML string) (name.ManifestMap, error) {
```
golangcibot: `genManifests` - `rootArgs` is unused (from `unparam`)

cmd/mesh/profile-dump.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	return finalYAML, err
+}
+
+func unmarshalAndValidateICP(crYAML string) (*v1alpha2.IstioControlPlaneSpec, string, error) {
+	// TODO: add GVK handling as appropriate.
+	icps, _, err := manifest.ParseK8SYAMLToIstioControlPlaneSpec(crYAML)
+	if err != nil {
+		return nil, "", fmt.Errorf("could not unmarshal the overlay file: %s\n\nOriginal YAML:\n%s\n", err, crYAML)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

cmd/mesh/profile-dump.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	if errs := validate.CheckIstioControlPlaneSpec(icps, false); len(errs) != 0 {
+		return nil, "", fmt.Errorf("input file failed validation with the following errors: %s\n\nOriginal YAML:\n%s\n", errs, crYAML)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

cmd/mesh/profile-dump.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return nil, "", fmt.Errorf("could not marshal: %s", err)
+	}
+	return icps, icpsYAML, nil
+}
+
+func unmarshalAndValidateICPS(icpsYAML string) (*v1alpha2.IstioControlPlaneSpec, error) {
+	icps := &v1alpha2.IstioControlPlaneSpec{}
+	if err := util.UnmarshalWithJSONPB(icpsYAML, icps); err != nil {
+		return nil, fmt.Errorf("could not unmarshal the merged YAML: %s\n\nYAML:\n%s\n", err, icpsYAML)
```
golangcibot: error strings should not be capitalized or end with punctuation or a newline (from `golint`)

cmd/mesh/profile-dump.go
```diff
@@ -0,0 +1,212 @@
...
+		return "", fmt.Errorf("could not read the profile values for %s: %s", fname, err)
+	}
+
+	_, baseYAML, err := unmarshalAndValidateICP(baseCRYAML)
+	if err != nil {
+		return "", err
+	}
+
+	// Merge base and overlay.
+	mergedYAML, err := helm.OverlayYAML(string(baseYAML), string(overlayYAML))
```
golangcibot: unnecessary conversion (from `unconvert`)

ostromart:
@morvencao I see a breakage in the tests here but I'm scratching my head how it could happen since I haven't modified any of the files in the following failed line:
helm template --namespace istio-system --name istio /usr/local/google/home/mostrowski/go/src/istio.io/operator/data/charts/crds -f /usr/local/google/home/mostrowski/go/src/istio.io/operator/data/charts/global.yaml -f /usr/local/google/home/mostrowski/go/src/istio.io/operator/tests/profiles/helm/values-istio-default.yaml
Error: render error in "istio-crds/templates/crds.yaml.orig": template: istio-crds/templates/crds.yaml.orig:72:14: executing "istio-crds/templates/crds.yaml.orig" at <.Values.certmanager....>: can't evaluate field enabled in type interface {}

cmd/mesh/root.go
```diff
@@ -60,26 +59,15 @@ func GetRootCmd(args []string) *cobra.Command {
...
 
 	rootArgs := &rootArgs{}
-	diffArgs := &manDiffArgs{}
-	dumpArgs := &dumpArgs{}
 
-	ic := installCmd(rootArgs)
 	mc := manifestCmd(rootArgs)
-	mdc := manifestDiffCmd(rootArgs, diffArgs)
-	dpc := dumpProfileDefaultsCmd(rootArgs, dumpArgs)
+	pc := profileCmd(rootArgs)
```
ayj: 1) `maniestCmd` and `profileCmd` need to be exported so we can integrate them into istioctl

2) Can you move `addFlags(mc, rootArgs)` into manifestCmd so that the caller (e.g. istioctl) doesn't need to deal with it? And a similar change for the profile flags.

ostromart: Done.

PR #106, opened 2019-07-21
fix port unmarshal issue and add ut.

morvencao:
Resolve: https://github.com/istio/istio/issues/15366

Also add back the removed UT for unmarshalling from YAML/JSON to go struct.

### DIFF
--- a/pkg/util/yaml_test.go
+++ b/pkg/util/yaml_test.go
@@ -15,14 +15,16 @@
 package util
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/gogo/protobuf/types"
 	"github.com/golang/protobuf/proto"
-	"k8s.io/api/autoscaling/v2beta1"
-	v1 "k8s.io/api/core/v1"
 
 	"istio.io/operator/pkg/apis/istio/v1alpha2"
+
+	v2beta1 "k8s.io/api/autoscaling/v2beta1"
+	v1 "k8s.io/api/core/v1"
 )
 
 var (
@@ -136,7 +138,6 @@ hub: docker.io/istio
 tag: 1.1.4
 defaultNamespacePrefix: istio-system
 profile: default
-
 trafficManagement:
   enabled: true
   components:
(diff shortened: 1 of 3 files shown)


pkg/util/yaml_test.go
```diff
@@ -233,3 +234,25 @@ func TestToYAMLWithJSONPB(t *testing.T) {
...
+		{"UnmarshalWithJSONPBToYAML", icpYaml, icp},
+	}
+
+	for _, tt := range unmarshalWithJSONPBTests {
+		t.Run(tt.desc, func(t *testing.T) {
+			got := &v1alpha2.IstioControlPlaneSpec{}
+			err := UnmarshalWithJSONPB(icpYaml, got)
+			if err != nil {
+				if !reflect.DeepEqual(got, tt.want) {
+					t.Errorf("TestUnmarshalWithJSONPB(%s): got:\n%v\n\nwant:\n%v", tt.desc, got, tt.want)
```
ayj: TestUnmarshalWithJSONPB should already be printed by the golang test runner and can be omitted.

pkg/util/yaml_test.go
```diff
@@ -233,3 +234,25 @@ func TestToYAMLWithJSONPB(t *testing.T) {
...
+		yaml string
+		want *v1alpha2.IstioControlPlaneSpec
+	}{
+		{"UnmarshalWithJSONPBToYAML", icpYaml, icp},
+	}
+
+	for _, tt := range unmarshalWithJSONPBTests {
+		t.Run(tt.desc, func(t *testing.T) {
+			got := &v1alpha2.IstioControlPlaneSpec{}
+			err := UnmarshalWithJSONPB(icpYaml, got)
```
richardwxn: should be tt.yaml instead of icpyaml? similar change need to be made for TestToYAMLWithJSONPB

pkg/util/yaml_test.go
```diff
@@ -233,3 +234,25 @@ func TestToYAMLWithJSONPB(t *testing.T) {
...
+			if err != nil {
```
richardwxn: should have err!=nil check before? then only compare when err is nil

morvencao: done.

pkg/util/yaml_test.go
morvencao: done.

pkg/util/yaml_test.go
morvencao: done.

pkg/util/yaml_test.go
ayj: s/t.Errorf/t.Fatalf unless you want the test case to continue after unexpected errors.

PR #107, opened 2019-07-22
Makefile: specify dependencies for vfsgen. Add iop target.

elfinhe:
Specify dependencies for vfsgen. Add iop target. Also update the Readme.

Having this PR out to get more ideas of how to improve the Makefile.

Resolve: https://github.com/istio/istio/issues/15237

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+export GO111MODULE=on
+
 gen_img := gcr.io/istio-testing/protoc:2019-03-29
 pwd := $(shell pwd)
 mount_dir := /src
@@ -42,6 +44,9 @@ gogofast_plugin := $(gogofast_plugin_prefix)$(gogo_mapping):$(out_path)
 api_path := pkg/apis/istio/v1alpha2
 api_protos := $(shell find $(api_path) -type f -name '*.proto' | sort)
 api_pb_gos := $(api_protos:.proto=.pb.go)
+
+.PHONY: default
+default: iop
 
 generate-api-go: $(api_pb_gos)
 	patch pkg/apis/istio/v1alpha2/istiocontrolplane_types.pb.go < pkg/apis/istio/v1alpha2/fixup_go_structs.patch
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -104,5 +109,9 @@ gen_patch_iscp:
 gen_patch_values:
 	diff -u pkg/apis/istio/v1alpha2/values/values_types.pb.go.orig pkg/apis/istio/v1alpha2/values/values_types.pb.go > pkg/apis/istio/v1alpha2/values/fix_values_structs.patch || true
 
-vfsgen:
+vfsgen: data/
+	go get github.com/shurcooL/vfsgen
```
ayj: This should eventually be baked into a docker image for reproducible builds. This pattern is used elsewhere in istio/istio to auto-generate code (e.g. see https://github.com/istio/istio/pull/7343). https://github.com/istio/istio/blob/master/docker/Dockerfile.go_generate_dependency exists and solves a similar problem but encoding files into golang code. However it uses github.com/jteeuwen/go-bindata/ which is not maintained. As we integrate parts of iop into istioctl it would be good to settle on a common package for encoding data files into istio binaries.

Makefile
```diff
@@ -104,5 +109,9 @@ gen_patch_iscp:
...
 	go generate ./cmd/iop.go
+
+iop: vfsgen
```
morvencao: Adding `vfsgen` as a dependency results in a dirty git status as `vfsgen` will always generate the go code with new timestamp. Should we remove this dependency? Unless dev want to make change to the baked in chart, in that case, we need to generate the code with `vfsgen`.

elfinhe: Yes, seems like we need to further improve the Makefile and make it smarter. But just like you mentioned, if I remove vfsgen dep, those who work on profiles and charts would say the build is broken. Ideally we should find a way to either `go generate` only when /data changes or exclude vfsgen_data.go from this git repo. Not quite sure, let me investigate more.

PR #109, opened 2019-07-23
Add a test target

sdake:
this is mostly to test if tide is doing its thing.


sdake:
WIll need to customize this page: https://prow.istio.io/tide if and when a transition is completed.

PR #110, opened 2019-07-23
support log level and scope.

morvencao:
Resolve: https://github.com/istio/istio/issues/15605

### DIFF
--- a/pkg/name/name.go
+++ b/pkg/name/name.go
@@ -26,7 +26,7 @@ import (
 )
 
 var (
-	DebugPackage = false
+	scope = log.RegisterScope("name", "name translator", 0)
 )
 
 // FeatureName is a feature name string, typed to constrain allowed values.
@@ -184,9 +184,9 @@ func GetFromStructPath(node interface{}, path string) (interface{}, bool, error)
 // getFromStructPath is the internal implementation of GetFromStructPath which recurses through a tree of Go structs
 // given a path. It terminates when the end of the path is reached or a path element does not exist.
 func getFromStructPath(node interface{}, path util.Path) (interface{}, bool, error) {
-	dbgPrint("getFromStructPath path=%s, node(%T)", path, node)
+	scope.Debugf("getFromStructPath path=%s, node(%T)", path, node)
 	if len(path) == 0 {
-		dbgPrint("getFromStructPath returning node(%T)%v", node, node)
+		scope.Debugf("getFromStructPath returning node(%T)%v", node, node)
 		return node, !util.IsValueNil(node), nil
 	}
 	kind := reflect.TypeOf(node).Kind()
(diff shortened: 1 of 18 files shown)


pkg/util/common.go
```diff
@@ -26,8 +28,7 @@ const (
 )
 
 var (
-	// DebugPackage controls verbose debugging in this package. Used for offline debugging.
-	DebugPackage = false
+	scope = log.RegisterScope("util", "common util", 0)
```
ostromart: s/common util/util/

pkg/patch/patch.go
```diff
@@ -100,8 +99,7 @@ import (
 )
 
 var (
-	// DebugPackage controls verbose debugging in this package. Used for offline debugging.
-	DebugPackage = false
+	scope = log.RegisterScope("patch", "yaml patch", 0)
```
ostromart: s/yaml patch/patch/

pkg/name/name.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 var (
-	DebugPackage = false
+	scope = log.RegisterScope("name", "name translator", 0)
```
ostromart: s/name translator/name/

pkg/validate/common.go
```diff
@@ -163,7 +162,7 @@ func validateIntRange(path util.Path, val interface{}, min, max int64) util.Erro
 
 // validateCIDR checks whether val is a string with a valid CIDR.
 func validateCIDR(path util.Path, val interface{}) util.Errors {
-	dbgPrintC("validateCIDR (%s): ", val)
+	scope.Debugf("validateCIDR (%s): ", val)
```
ostromart: This is going to mess up the output. dbgPrintC doesn't append a newline at the end. Here and wherever dbgPrintC is used.

pkg/util/common.go
morvencao: done

pkg/name/name.go
morvencao: done

pkg/patch/patch.go
morvencao: done

pkg/validate/common.go
morvencao: Then we need to concatenate the debug log and print before return.

ostromart: Yeah, this is exactly why I dragged my feet on moving this to scopes. One possibility is to port to scopes just the files that don't use dbgPrintC. From memory it's just this file that requires it, all others are fair game.

ayj: You could create a helper function that is called via `defer` which prints logs with optional error, e.g.

```golang
func logWithError(err *error, format string, args ...interface{}) {
	msg := fmt.Sprintf(format, args...)
	if err != nil {
		msg += fmt.Sprintf(": %v", err)
	}
    scope.Debug(msg)
}

func validateCIDR(path util.Path, val interface{}) util.Errors {
    var err error
    defer logWithError(&err, "validateCIRD (%v)", val)
}

```

morvencao: @ostromart @ayj thanks, it helps a lot. We may still need to optimize the logs format in `validateStringList` function, since it coheres with other validation functions.

PR #114, opened 2019-07-23
Add documentation to GalleyConfig.

elfinhe:
For https://github.com/istio/istio/issues/15696


pkg/apis/istio/v1alpha2/values/values_types.proto
```diff
@@ -136,18 +136,53 @@ message EnvoyMetricsConfig {
...
   bool enabled = 1;
 
+  // Image name used for Galley.
+  //
+  // This can be set either to image name if hub is also set, or can be set to the full hub:name string.
+  //
+  // Examples: custom-galley, docker.io/someuser:custom-galley
   string image = 2;
 
   map<string, string> mesh = 3;
```
ostromart: missing? maybe you can follow up with @ayj for docs on this.

rcaballeromx:
Added the do-not-merge label to give the docs team a chance to review.

pkg/apis/istio/v1alpha2/values/values_types.proto
```diff
@@ -136,18 +136,53 @@ message EnvoyMetricsConfig {
...
-// GalleyConfig is described in istio.io documentation.
+// GalleyConfig is a set of configuration parameters for Galley.
 message GalleyConfig {
+  // Controls whether Galley is enabled.
   bool enabled = 1;
 
+  // Image name used for Galley.
+  //
+  // This can be set either to image name if hub is also set, or can be set to the full hub:name string.
+  //
```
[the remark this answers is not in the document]
elfinhe: Sure, will update.

pkg/apis/istio/v1alpha2/values/values_types.proto
elfinhe: I cannot find any other code is using the GalleyConfig.Mesh in operator repo project. I can find something similar [in 'config.proto'](https://github.com/istio/api/blob/master/mesh/v1alpha1/config.proto#L30-L39). But looks like they are not the same thing.

@richardwxn, do you know more about the mesh field, or where it is coming?

elfinhe: Talked with @richardwxn offline. This field is coming from values_types.go.
Since it is not used yet. I will delete this field for now.

elfinhe: I asked around, but seems like no one knows this 'mesh' is used for.

elfinhe: Costin told me this is copied from Pilot. Because "one of the main changes between old installer and new installer is that each component is self-contained and independent of other components. We can't have Galley depend on a config map created by pilot."

So I suggest we'd better keep the 'mesh' here.

ozevren: That doesn't explain what this is meant for though. So why do we have this field?

elfinhe: maybe out of scope for this PR. I will add TODO for this mesh field just like we did [here](https://github.com/istio/operator/blob/8c190bfa9dc175a05544d75d95cd2d8ba93c5d3b/pkg/component/controlplane/testdata/charts/istio-control/istio-config/values.yaml#L14).

Created an issue to track: https://github.com/istio/istio/issues/15865

PR #115, opened 2019-07-24
Add documentation to SecurityConfig.

morvencao:
Resolve: https://github.com/istio/istio/issues/15692


morvencao:
@ostromart Looks like most of fields for `CertManager` and `NodeAgent` have not been migrated to `values_proto`, I can only find some basic fields, eg, image, tag...

richardwxn:
@morvencao I think most of the fields are already there, but I found out that some fields like NodeSelector is missing. The reason was because the initial version of values_types.proto was generated from values_types.go, but the tool could not handle type like map[string]interface{} properly so I commented out them before generator processed. However, if that commented field is the last field in the struct, the tool would ignore it instead of keeping that comment line.

morvencao:
@richardwxn Thanks.

rcaballeromx:
Added the do-not-merge label to give the docs team a chance to review.

PR #124, opened 2019-07-25
update function writeStringSafe so as to print diff string

waret:


pkg/object/objects.go
```diff
@@ -369,15 +369,15 @@ func ManifestDiff(a, b string) (string, error) {
...
+				writeStringSafe(&sb, "\n\nObject "+bk+" is missing:\n\n")
+				writeStringSafe(&sb, diff)
 			}
 		}
 	}
 	return sb.String(), err
 }
 
-func writeStringSafe(sb strings.Builder, s string) {
+func writeStringSafe(sb *strings.Builder, s string) {
```
golangcibot: `sb` can be `io.StringWriter` (from `interfacer`)

pkg/object/objects.go
```diff
@@ -301,7 +302,10 @@ func (os K8sObjects) Sort(score func(o *K8sObject) int) {
 func (os K8sObjects) ToMap() map[string]*K8sObject {
 	ret := make(map[string]*K8sObject)
 	for _, oo := range os {
-		ret[oo.Hash()] = oo
+		// ignore comments
+		if oo.Kind != "" || oo.Name != "" || oo.Namespace != "" {
```
ostromart: how about putting this into a helper function like (o *K8sObject) empty() bool{} ?
this will also make the code self-documenting.

pkg/object/objects.go
```diff
@@ -22,10 +22,11 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"io"
 	"sort"
 	"strings"
 
-	"github.com/ghodss/yaml"
+	ghodssyaml "github.com/ghodss/yaml"
```
ostromart: this is a good idea but we should stay consistent. i propose reverting this and starting a new PR to make this change throughout the project.

pkg/object/objects.go
```diff
@@ -310,11 +314,22 @@ func (os K8sObjects) ToMap() map[string]*K8sObject {
...
+			ret[oo.HashNameKind()] = oo
+		}
 	}
 	return ret
 }
 
+func (o *K8sObject) Valid() bool {
+	if o.Kind != "" && o.Name != "" {
+		return true
+	} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

pkg/object/objects.go
```diff
@@ -310,11 +314,21 @@ func (os K8sObjects) ToMap() map[string]*K8sObject {
...
-		ret[oo.HashNameKind()] = oo
+		// ignore invalid k8s object
+		if oo.Valid() {
+			ret[oo.HashNameKind()] = oo
+		}
 	}
 	return ret
 }
 
+func (o *K8sObject) Valid() bool {
```
ostromart: Add a comment since function is exported.

pkg/object/objects.go
```diff
@@ -301,7 +302,10 @@ func (os K8sObjects) Sort(score func(o *K8sObject) int) {
 func (os K8sObjects) ToMap() map[string]*K8sObject {
 	ret := make(map[string]*K8sObject)
 	for _, oo := range os {
-		ret[oo.Hash()] = oo
+		// ignore invalid k8s object
```
ostromart: I think these comments can be removed now since the code is pretty clear.

PR #126, opened 2019-07-26
Version plumbing for operator

ostromart:
This PR introduces a versions map and related tests. 
The operator version is in version/version.go.
The versions map is in the file version/versions.yaml.
The versions.yaml map uses go-versions SemVar constraints implementation to define for each version of operator the supported chart versions.


pkg/version/version.go
```diff
@@ -14,68 +14,176 @@
...
+import (
+	"fmt"
+	"strings"
 
+	goversion "github.com/hashicorp/go-version"
+)
+
+// IstioOperatorVersionCompatibility is a mapping from an Istio operator version and the corresponding recommended and
+// supported versions of Istio.
+type IstioOperatorVersionCompatibility struct {
```
ayj: VersionCompatibility or Compatibility to avoid stutter? IstioOperator seems redundant given the repository of this package :)

cmd/iop/install.go
```diff
@@ -50,7 +50,7 @@ func installManifests(args *rootArgs, logOpts *log.Options) {
 		logAndFatalf(args, "Could not generate manifest: %v", err)
 	}
 
-	out, err := manifest.ApplyAll(manifests, version.NewVersion("", 1, 2, 0, ""), args.dryRun, args.verbose)
+	out, err := manifest.ApplyAll(manifests, version.NewVersion(1, 2, 0, ""), args.dryRun, args.verbose)
```
ayj: Should this be 1.3.0 as well?

pkg/version/version.go
```diff
@@ -14,68 +14,176 @@
...
 }
 
+// String implements the Stringer interface.
 func (v MinorVersion) String() string {
 	return fmt.Sprintf("%s.%d", v.MajorVersion, v.Minor)
 }
 
+// String implements the Stringer interface.
 func (v PatchVersion) String() string {
 	return fmt.Sprintf("%s.%d", v.MinorVersion, v.Minor)
```
richardwxn: BTW the second argument should be v.Patch?

version/versions.yaml
```diff
@@ -0,0 +1,4 @@
+- operatorVersion: 1.3.0
+  supportedIstioVersions: 1.3.0
+  recommendedVersions: 1.3.0
```
morvencao: Should be `recommendedIstioVersions` instead of `recommendedVersions` according to yaml tag: https://github.com/istio/operator/blob/cbe894b2178fed887ff8982e2b02e4b0058210e9/pkg/version/version.go#L55

pkg/version/version.go
ostromart: Done.

cmd/iop/install.go
ostromart: Actually it should be using the top level version definition. I've removed this hardcoding, here and elsewhere.

pkg/version/version.go
ostromart: Good catch, thanks.

version/versions.yaml
ostromart: Done.

version/version.go
```diff
@@ -14,6 +14,30 @@
 
 package version
```
sdake: the root version dir seems like the wrong place for this file.  should it not be pkg/version/version.go?

ostromart: I struggled with this a bit but in practice it will probably be ok. The difference is that pkg/version is a general library for version handling, while version/version deals specifically with specifying the binary version of the operator and the supported version mappings. 
As such, the release team would be going into version, while developers would stick to pkg/version. 
We can discuss this and change it if you have some good idea here.

PR #128, opened 2019-07-27
Merge fixes

ostromart:
A bunch of problems fixed due to heavy merging activity:
 - update code with proto naming changes (e.g. defaultNamespacePrefix changed)
 - package source selection has changed (file:/// no longer required for local path)

The current code still has one breakage: it does not correctly set ingress / egress gateway enablement and namespaces. This requires some path selection work to be fixed.


pkg/helm/helm.go
```diff
@@ -118,6 +123,10 @@ func renderChart(namespace, baseValues, overlayValues string, chrt *chart.Chart)
 
 	var sb strings.Builder
 	for _, f := range files {
+		// add yaml separater if the rendered file doesn't have one at the end
```
richardwxn: nit: separator

richardwxn:
Actually it would be great if you can give details about what are the problems that this PR tries to fix because we might be missing contexts for some of them

cmd/mesh/manifest.go
```diff
@@ -28,7 +28,8 @@ import (
 	"istio.io/operator/pkg/translate"
 	"istio.io/operator/pkg/util"
 	"istio.io/operator/pkg/validate"
-	"istio.io/operator/pkg/version"
+
+	binversion "istio.io/operator/version"
```
ayj: Why is this alias needed? Also, there's an extra space between `validate` and `version` packages.

ostromart: I added this for consistency - we now have two version packages: a library under pkg/version and the compiled in version in version/version (which we could rename I suppose).

pkg/helm/helm.go
ostromart: Done.

ostromart:
@richardwxn I updated the PR description to give some more context.

PR #129, opened 2019-07-27
Add validation for IP ranges in operator's validate.go

elfinhe:
ValidateCIDR is implemented in common.go but not used by any non-test code.
Now use ValidateCIDR to validate  ip ranges in profiles.

Resolve: https://github.com/istio/istio/issues/15683

### DIFF
--- a/pkg/validate/common.go
+++ b/pkg/validate/common.go
@@ -88,7 +88,7 @@ var (
 // validateWithRegex checks whether the given value matches the regexp r.
 func validateWithRegex(path util.Path, val interface{}, r *regexp.Regexp) (errs util.Errors) {
 	switch {
-	case !isString(val):
+	case !util.IsString(val):
 		errs = util.AppendErr(errs, fmt.Errorf("path %s has bad type %T, want string", path, val))
 
 	case len(r.FindString(val.(string))) != len(val.(string)):
@@ -104,7 +104,7 @@ func validateWithRegex(path util.Path, val interface{}, r *regexp.Regexp) (errs 
 func validateStringList(vf ValidatorFunc) ValidatorFunc {
 	return func(path util.Path, val interface{}) util.Errors {
 		msg := fmt.Sprintf("validateStringList %v", val)
-		if reflect.TypeOf(val).Kind() != reflect.String {
+		if !util.IsString(val) {
 			err := fmt.Errorf("validateStringList %s got %T, want string", path, val)
 			printError(err)
 			return util.NewErrs(err)
@@ -123,7 +123,7 @@ func validateStringList(vf ValidatorFunc) ValidatorFunc {
 // validatePortNumberString checks if val is a string with a valid port number.
 func validatePortNumberString(path util.Path, val interface{}) util.Errors {
 	scope.Debugf("validatePortNumberString %v:", val)
-	if !isString(val) {
+	if !util.IsString(val) {
 		return util.NewErrs(fmt.Errorf("validatePortNumberString(%s) bad type %T, want string", path, val))
 	}
 	intV, err := strconv.ParseInt(val.(string), 10, 32)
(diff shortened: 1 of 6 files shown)


pkg/validate/validate.go
```diff
@@ -142,6 +144,28 @@ func validateDefaultNamespace(path util.Path, val interface{}) util.Errors {
 	return validateWithRegex(path, val, ObjectNameRegexp)
 }
 
+func validateIPRanges(path util.Path, val interface{}) util.Errors {
+	reflectVal := val
+	// Type switch.
```
richardwxn: this comment is not necessary

pkg/validate/validate.go
```diff
@@ -142,6 +144,28 @@ func validateDefaultNamespace(path util.Path, val interface{}) util.Errors {
...
+		// Handle extractable values with special methods
+		// since validate() does not handle them.
+		if f.IsValid() && f.CanInterface() {
+			reflectVal = f.Interface()
+		}
+	}
+
+	// Allow "*" as IP ranges
+	if reflect.TypeOf(reflectVal).Kind() == reflect.String {
+		valStr := reflectVal.(string)
```
richardwxn: not necessary to create new var

pkg/validate/validate.go
elfinhe: Removed.

pkg/validate/validate.go
elfinhe: Fixed.

pkg/validate/validate.go
```diff
@@ -31,6 +31,8 @@ var (
 		"BaseSpecPath":      validateInstallPackagePath,
 		"CustomPackagePath": validateInstallPackagePath,
 		"DefaultNamespace":  validateDefaultNamespace,
+		"TrafficManagement.Components.Proxy.Common.Values.includeIPRanges": validateIPRanges,
+		"TrafficManagement.Components.Proxy.Common.Values.excludeIpRanges": validateIPRanges,
```
ostromart: All values are validated in validate_values.go - please move this PR there.
These fields are already validate there but I think "*" is not checked.

pkg/validate/validate.go
```diff
@@ -142,6 +144,30 @@ func validateDefaultNamespace(path util.Path, val interface{}) util.Errors {
 	return validateWithRegex(path, val, ObjectNameRegexp)
 }
 
+func validateIPRanges(path util.Path, val interface{}) util.Errors {
```
ostromart: Need to update UT to check * case.

pkg/validate/validate.go
elfinhe: Thanks for clarification! Updated.

pkg/validate/validate.go
elfinhe: Test added.

pkg/validate/common.go
```diff
@@ -110,6 +110,9 @@ func validateStringList(vf ValidatorFunc) ValidatorFunc {
 			return util.NewErrs(err)
 		}
 		var errs util.Errors
+		if val.(string) == "*" {
```
ostromart: This logic doesn't seem quite right. This function is intended as a wrapper to call a validator function vf, which validates each field. There should not be any field specific validation logic here. 
Rather you can create a validator function which accepts either an IP address list or "*" - if it's "*" return ok, or otherwise call this function.

pkg/validate/validate_values_test.go
```diff
@@ -32,6 +32,15 @@ func TestValidateValues(t *testing.T) {
 		{
 			desc: "nil success",
 		},
+		{
+			desc: "AsteriskIPRange",
+			yamlStr: `
```
ostromart: Should also test "*, 1.2.3.4" as an error case.

elfinhe: Added below, thanks.

pkg/validate/common.go
elfinhe: Updated.

elfinhe:
pkg/component/controlplane/control_plane_test.go:47
TestRenderInstallationSuccessV13 failed on prow. But it shows success on my local box.

PR #131, opened 2019-07-28
switch to kube client to apply manifest and support kubeconfig context.

morvencao:
Resolve: https://github.com/istio/istio/issues/15844


cmd/mesh/manifest-apply.go
```diff
@@ -39,6 +45,8 @@ type manifestApplyArgs struct {
 func addManifestApplyFlags(cmd *cobra.Command, args *manifestApplyArgs) {
 	cmd.PersistentFlags().StringVarP(&args.inFilename, "filename", "f", "", filenameFlagHelpStr)
 	cmd.PersistentFlags().StringVarP(&args.kubeConfigPath, "kubeconfig", "c", "", "Path to kube config.")
+	cmd.PersistentFlags().StringVar(&args.context, "context", "", "The name of the kubeconfig context to use.")
+	cmd.PersistentFlags().StringVarP(&args.namespace, "namespace", "n", v1.NamespaceAll, "Namespace for the kubeConfig.")
```
ostromart: Is this really required? IstioControlPlane has defaultNamespace which defaults to istio-system in the profiles.

pkg/kubeclient/client.go
```diff
@@ -0,0 +1,162 @@
...
+				return err
+			}
+			return nil
+		}
+	}
+
+	labels := existingObj.GetLabels()
+	if containsSelector(labels, selector) {
+		if prune {
+			// if there already exists resource with the selector, then delete the existing resource before creating new one.
```
ostromart: --prune actually works a bit differently - it deletes anything that's not in the given manifest or file path. It's used to clean up resources from a previous version. Mostly because of this I recommend deferring this work until the kubectl library is publicly available - there are a lot of subtle behaviors to worry about that it will solve for us.
Also I think we'd need patch or some sort since I'm not sure that create will work for updating existing objects.

morvencao: Agreed, we should have more reconcile process here, eg. patch if the resource already exists, update the reversion...

cmd/mesh/manifest-apply.go
morvencao: Removed `namespace` flag.

ayj:
The `--context` portion of this PR has been addressed in https://github.com/istio/istio/pull/16392. 

Can we use the `kubectl` binary for the `apply` implementation until server-side apply is ready? This would simplify the implementation while retaining the existing apply semantics w.r.t patching and pruning. The downside is it would mean the `istioctl binary` would depend on the `kubectl` binary in the `PATH`. cc @esnible on whether he thinks this is a reasonable assumption for istioctl users deploying to Kubernetes.

esnible:
@ayj I don't understand the question ... this code already uses `kubectl`.  (This surprised me yesterday when I tested the new subcommands!)

ayj:
Is it reasonable for `istioctl manifest apply` to depend on the kubectl binary? The alternative is to implement the equivalent functionality with client-go.

esnible:
@ayj It is strange that _istioctl_ now depends on a _kubectl_ binary.  Even _helm_ doesn't need _kubectl_!  However, it seems too late to refactor for 1.3.  _istioctl_ now depends on _helm_ and it would be strange to allow that but complain about _kubectl_.

ostromart:
I've chatted about this with @morvencao previously and we agree to park this work for the time being and go with the kubectl approach for now. It's a lot more work to replicate the behavior of kubectl apply semantics, especially --prune, so we probably wait until server-side apply is available.

ayj:
Should we close this PR? It will get quiet out of date before server-apply is available by default.

ostromart:
yes, i think it's a good idea to close it, but i'll let morven decide what he wants to do with it.

morvencao:
I'm OK to close this.

PR #41, opened 2019-06-21
Initial implementation of translation of value.yaml

richardwxn:

PR #132, opened 2019-07-29
Add manifest migrate command

richardwxn:
Initial implementation, need to wait for #41 to merged


cmd/mesh/manifest-migrate.go
```diff
@@ -16,22 +16,78 @@ package mesh
...
+	if err := configLogs(rootArgs); err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "Could not configure logs: %s", err)
+		os.Exit(1)
+	}
+	f, err := ioutil.ReadFile(mmArgs.inFilename)
+	if err != nil {
+		logAndFatalf(rootArgs, "failed to open values.yaml file: %s", err.Error())
+	}
+
+	ts, err := translate.NewValueYAMLTranslator(version.NewMinorVersion(1, 3))
```
golangcibot: NewValueYAMLTranslator not declared by package translate (from `typecheck`)

cmd/mesh/manifest-migrate.go
```diff
@@ -16,22 +16,78 @@ package mesh
...
+	"istio.io/operator/pkg/translate"
+	"istio.io/operator/pkg/version"
 )
 
-func manifestMigrateCmd(rootArgs *rootArgs) *cobra.Command {
+type manifestMigrateArgs struct {
+	// inFilename is the path to the input values.yaml file.
+	inFilename string
+	// outFilename is the path to the translated output filename.
+	outFilename string
```
ostromart: I think it would be better to require a user to specify a file and output to stdout because this is the only mode for this command. Having these as flags implies that they are optional.

cmd/mesh/manifest-migrate.go
```diff
@@ -16,22 +16,78 @@ package mesh
...
-func manifestMigrate(_ *rootArgs) {
-	fmt.Println("Not yet implemented.")
+func manifestMigrate(rootArgs *rootArgs, mmArgs *manifestMigrateArgs) {
+	if err := configLogs(rootArgs); err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "Could not configure logs: %s", err)
+		os.Exit(1)
+	}
+	f, err := ioutil.ReadFile(mmArgs.inFilename)
+	if err != nil {
+		logAndFatalf(rootArgs, "failed to open values.yaml file: %s", err.Error())
```
ostromart: nit: the ioutil errors usually provide enough context that you don't need additional decoration, you can just print the error only.

cmd/mesh/manifest-migrate.go
richardwxn: Can you help clarify two points here, 1. we have a pending issue if we do not specify a input file then we need to read the specs from cluster directly, so input seems like an optional field? 2. output file is also an option if user wants to write it to new file

ostromart: 1. you could just as easily add the default to be cluster config if no file is specified. It does not need to be through -f flag. 
2. -o seems very limited value since it's just command > output file. Most linux commands follow this pattern of not having a -o flag if it can be supported with file redirect e.g. see helm template --help. manifest generate only has -o because it outputs to dirs.

richardwxn: yep that makes sense

cmd/mesh/manifest-migrate.go
richardwxn: done

cmd/mesh/manifest-diff.go
```diff
@@ -16,12 +16,11 @@ package mesh
 
 import (
 	"fmt"
+	"github.com/spf13/cobra"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

pkg/kubectlcmd/client.go
```diff
@@ -92,6 +92,37 @@ func (c *Client) Apply(dryRun, verbose bool, namespace string, manifest string,
 	return stdout.String(), stderr.String(), nil
 }
 
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

cmd/mesh/manifest-migrate.go
```diff
@@ -15,23 +15,131 @@
...
-func manifestMigrateCmd(rootArgs *rootArgs) *cobra.Command {
+type manifestMigrateArgs struct {
+	// migrateDir indicates whether to migrate all values.yaml files in directory
+	migrateDir bool
+	// namespace is the namespace to get the in cluster configMap
+	namespace string
+}
+
+func addManifestMigrateFlags(cmd *cobra.Command, args *manifestMigrateArgs) {
+	cmd.PersistentFlags().BoolVarP(&args.migrateDir, "migrateDir", "r", false, "translate directory")
```
ostromart: "translate directory" -> "Directory tree containing Helm values YAML files. Only files with .yaml suffix are read."
Note that if you elect to remove --migrateDir, you should still create a filename flag here to explain that it can be a dir or file.

cmd/mesh/manifest-migrate.go
```diff
@@ -15,23 +15,131 @@
...
+	cmd.PersistentFlags().StringVarP(&args.namespace, "namespace", "n", defaultNamespace, "specifies namespace")
```
ostromart: specifies namespace -> Default namespace for output IstioControlPlane CustomResource.

cmd/mesh/manifest-migrate.go
```diff
@@ -15,23 +15,131 @@
...
+)
+
+const (
+	defaultNamespace = "istio-system"
 )
 
-func manifestMigrateCmd(rootArgs *rootArgs) *cobra.Command {
+type manifestMigrateArgs struct {
+	// migrateDir indicates whether to migrate all values.yaml files in directory
+	migrateDir bool
```
ostromart: You could eliminate this flag and test if the given path is a dir.

cmd/mesh/manifest-migrate.go
```diff
@@ -15,23 +15,131 @@
...
 
+func valueFileFilter(path string) bool {
+	return filepath.Base(path) == "values.yaml"
 }
 
-func manifestMigrate(_ *rootArgs) {
-	fmt.Println("Not yet implemented.")
+// migrateFromFiles handles migration for local values.yaml files
+func migrateFromFiles(rootArgs *rootArgs, args []string, mmArgs *manifestMigrateArgs) {
+	if err := configLogs(rootArgs); err != nil {
```
ostromart: There's a checkLogsOrExit function now.

cmd/mesh/manifest-migrate.go
```diff
@@ -15,23 +15,131 @@
...
+	isCPSpec, err := ts.TranslateFromValueToSpec(&valueStruct)
+	if err != nil {
+		logAndFatalf(rootArgs, "error translating values.yaml: %s", err.Error())
+	}
+	isCPYaml, err := yaml.Marshal(isCPSpec)
+	if err != nil {
+		logAndFatalf(rootArgs, "error marshalling translated IstioControlPlaneSpec: %s", err.Error())
+	}
+
+	fmt.Println(string(isCPYaml))
```
ostromart: This should be a writer in case -o is set.

pkg/kubectlcmd/client.go
```diff
@@ -92,6 +92,37 @@ func (c *Client) Apply(dryRun, verbose bool, namespace string, manifest string,
...
+// Apply runs the kubectl apply with the provided manifest argument
```
ostromart: Update comment.

pkg/util/util.go
```diff
@@ -48,3 +55,32 @@ func PrettyJSON(b []byte) []byte {
 	}
 	return out.Bytes()
 }
+
+// ReadFromDir reads a directory recursively and gets the files based on filter.
```
ostromart: Add: It returns a concatenated output of all matching files' content.

cmd/mesh/manifest-migrate.go
richardwxn: the -o flag was actually removed based on last review

cmd/mesh/manifest-migrate.go
richardwxn: ok will change

pkg/kubectlcmd/client.go
richardwxn: ok will change

pkg/util/util.go
richardwxn: ok will update

cmd/mesh/manifest-migrate.go
richardwxn: done

cmd/mesh/manifest-migrate.go
richardwxn: done

cmd/mesh/manifest-migrate.go
richardwxn: actually I think specifying this flag makes it more clear to user that what they are expecting. But it depends on whether user use it more often with directory or single file. If they use it more often with directory then we should eliminate this flag.

cmd/mesh/manifest-migrate.go
```diff
@@ -15,23 +15,131 @@
...
+const (
+	defaultNamespace = "istio-system"
 )
 
-func manifestMigrateCmd(rootArgs *rootArgs) *cobra.Command {
+type manifestMigrateArgs struct {
+	// migrateDir indicates whether to migrate all values.yaml files in directory
+	migrateDir bool
+	// namespace is the namespace to get the in cluster configMap
+	namespace string
```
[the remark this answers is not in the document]
richardwxn: Can you clarify this point?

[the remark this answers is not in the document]
ostromart: namespace should be specified in the helm values you're reading from (istioNamespace). For in-cluster i thought it may be better to mention the map name too but probably this is unlikely to change so i think what you have is fine.

cmd/mesh/manifest-migrate.go
ostromart: consider a situation like the flag is set but the path is a file - you have to document it and do additional error handling. imo it's better to document the behavior to avoid such inconsistencies and keep the api such that it's not possible to create inconsistent parameter sets, but it's up to you.

cmd/mesh/manifest-migrate.go
ostromart: ah, you're right - sorry.

pkg/helm/http_renderer.go
```diff
@@ -0,0 +1,60 @@
...
+		helmChartDirPath: helmChartDirPath,
+		globalValues:     globalValues,
+	}
+}
+
+// Run implements the TemplateRenderer interface.
+func (h *HTTPTemplateRenderer) Run() error {
+	//
+
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)

cmd/mesh/manifest-migrate.go
```diff
@@ -15,23 +15,119 @@
...
+	value, err := util.ReadFiles(args[0], valueFileFilter)
+	if err != nil {
+		logAndFatalf(rootArgs, err.Error())
+	}
+	translateFunc(rootArgs, []byte(value))
+}
+
+// translateFunc translates the input values and output the result
+func translateFunc(rootArgs *rootArgs, values []byte) {
+	ts, err := translate.NewReverseTranslator(version.NewMinorVersion(1, 3))
```
golangcibot: NewReverseTranslator not declared by package translate (from `typecheck`)

cmd/mesh/manifest-migrate.go
```diff
@@ -15,23 +15,119 @@
...
+}
 
+func valueFileFilter(path string) bool {
+	return filepath.Base(path) == "values.yaml"
 }
 
-func manifestMigrate(_ *rootArgs) {
-	fmt.Println("Not yet implemented.")
+// migrateFromFiles handles migration for local values.yaml files
+func migrateFromFiles(rootArgs *rootArgs, args []string, mmArgs *manifestMigrateArgs) {
```
golangcibot: `migrateFromFiles` - `mmArgs` is unused (from `unparam`)

cmd/mesh/manifest-migrate.go
```diff
@@ -15,23 +15,119 @@
...
 		Use:   "migrate",
 		Short: "Migrates a file containing Helm values to IstioControlPlane format.",
 		Long:  "The migrate subcommand is used to migrate a configuration in Helm values format to IstioControlPlane format.",
-		Args:  cobra.ExactArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
-			manifestMigrate(rootArgs)
+			if len(args) == 0 {
+				migrateFromClusterConfig(rootArgs, mmArgs)
+			} else {
+				migrateFromFiles(rootArgs, args, mmArgs)
```
golangcibot: too many arguments (from `typecheck`)

PR #133, opened 2019-07-29
Remove repeated from gateway fields, related fixes

ostromart:
Gateways started off being a slice but many changes are required to make this work cleanly - therefore this is deferred for 1.4, with gateways being single instance, as they are for installer. 
This also fixes some related path issues discovered during testing.
This PR restores run_mandiff.sh back to working order.


richardwxn:
#41 would be rebased on this change

PR #135, opened 2019-07-30
Fix the invalid metadata.labels

clyang82:
Resolve: istio/istio#15920

Signed-off-by: Chun Lin Yang <clyang@cn.ibm.com>

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -182,6 +182,9 @@ func (v PatchVersion) String() string {
 
 // String implements the Stringer interface.
 func (v *Version) String() string {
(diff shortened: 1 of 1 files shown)


pkg/version/version.go
```diff
@@ -182,6 +182,9 @@ func (v PatchVersion) String() string {
 
 // String implements the Stringer interface.
 func (v *Version) String() string {
+	if v.Suffix == "" {
+		return fmt.Sprintf("%s", v.PatchVersion)
```
golangcibot: S1025: should use String() instead of fmt.Sprintf (from `gosimple`)

PR #138, opened 2019-07-31
Fixes to charts to make manifest apply work

ostromart:


pkg/object/objects.go
```diff
@@ -239,6 +239,7 @@ func ParseK8sObjectsFromYAMLManifest(manifest string) (K8sObjects, error) {
 	var objects K8sObjects
 
 	for _, yaml := range yamls {
+		yaml = removeCommentedLines(yaml)
```
richardwxn: the below blank check logic can be added to the new functionand also update the function name

pkg/object/objects.go
```diff
@@ -279,6 +291,32 @@ func (os K8sObjects) JSONManifest() (string, error) {
 	return b.String(), nil
 }
 
+// YAMLManifest returns a YAML representation of K8sObjects os.
+func (os K8sObjects) YAMLManifest() (string, error) {
+	var b bytes.Buffer
+
+	for i, item := range os {
+		if i != 0 {
+			if _, err := b.WriteString("\n\n"); err != nil {
```
richardwxn: why this is necessary since we already write YAMLSeparator at the end

pkg/vfsgen/vfsgen_data.go
```diff
@@ -1583,42 +1583,42 @@ var Assets = func() http.FileSystem {
 		},
 		"/profiles": &vfsgen۰DirInfo{
 			name:    "profiles",
-			modTime: time.Date(2019, 7, 29, 23, 17, 29, 916093878, time.UTC),
+			modTime: time.Date(2019, 7, 30, 22, 26, 48, 807533943, time.UTC),
```
richardwxn: Do we need to update the vfs gen data every time?

pkg/object/objects.go
ostromart: Done.

pkg/object/objects.go
ostromart: This is because the object YAMLs don't have separators. We only have those in the charts but they disappear after being parsed in to objects.

pkg/vfsgen/vfsgen_data.go
ostromart: No, only after charts change. This is just incidental but doesn't cause harm.

PR #140, opened 2019-07-31
Update hub and tag to master

ostromart:


morvencao:
Resolve:


@ostromart Please also update global.yaml: https://github.com/istio/operator/blob/master/data/charts/global.yaml#L36-L42
https://github.com/istio/operator/blob/master/pkg/component/controlplane/testdata/charts/global.yaml#L35-L41

ostromart:
Update global.yaml in the data directory only (it's fine to have a version number for the tests).
Also, keep in mind that we'll be deleting these charts soon and moving over to using istio/installer charts as source of truth.

PR #141, opened 2019-07-31
Release script

ostromart:
This script is intended to be called during the istio/istio release process to add the charts, profiles and version map for the operator. 
For either numbered or continuous releases from istio/istio, the process is the same: check out istio/operator at the desired version (master, branch, tag) and call the script with output path pointing to the root to the istio release tar. The release script internally checks out istio/installer at the same version and includes charts from there. 
The embedded version.yaml is the version of both the operator binary and the charts that are included in the release installation package. 
The embedded versions.yaml is a map between all supported operator versions and the corresponding supported installation package versions.


release/create_release_charts.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
```
utka: Where do you call this file in release process?

ostromart: From create_release_archives I think.

version/version.go
```diff
@@ -32,10 +36,21 @@ var (
 
 	// OperatorBinaryVersion is the Istio operator version.
 	OperatorBinaryVersion pkgversion.Version
+	// OperatorBinaryGoVersion is the Istio operator version in go-version format.
+	OperatorBinaryGoVersion *goversion.Version
 )
 
 func init() {
-	v, err := pkgversion.NewVersionFromString(OperatorBinaryVersionString)
+	b, err := ioutil.ReadFile("v")
```
morvencao: should be `b, err := ioutil.ReadFile(versionFileName)` ?

release/create_release_charts.sh
```diff
@@ -0,0 +1,98 @@
...
+    mkdir -p "${OUTPUT_CHARTS_DIR}"
+
+    for chart in "${INSTALLER_CHARTS[@]}"
+    do
+	    cp -R "${INSTALLER_REPO}/${chart}" "${OUTPUT_CHARTS_DIR}"
+    done
+}
+
+function copy_profiles() {
+    cp -R "${OPERATOR_BASE_DIR}/data/profiles" "${OUTPUT_DIR}"
```
morvencao: Should we update image hub and tag from daily build to released version in profiles here?

ostromart: I was thinking of doing that on the release branch. But we do need a script to update the hab and tag in all places rather than edit manually: https://github.com/istio/istio/issues/15997

sdake:
> /lgtm
> /hold

@utka this repo is on the new plan related to removing /foo commands.  As a result, /lgtm is non-functional (but I think /hold is operational).

Cheers
-steve

release/create_release_charts.sh
```diff
@@ -0,0 +1,98 @@
...
+OPERATOR_VERSION=$(get_version)
+TEMP_DIR=${TEMP_DIR:-"$(mktemp -d ${TEMP_DIR_DEFAULT}/istio.${OPERATOR_VERSION}.XXXX)"}
+
+INSTALLER_VERSION=${INSTALLER_VERSION:-"${OPERATOR_VERSION}"}
+OUTPUT_DIR="${OUTPUT_ROOT_DIR}/install/kubernetes/operator"
+
+mkdir -p "${OUTPUT_DIR}"
+
+function copy_installer_charts() {
+    cd "${TEMP_DIR}"
```
sdake: I know I'm late to the party...  I am not keen on bundling of the charts.  I thought the plan was to reference them directly from upstream?

release/create_release_charts.sh
```diff
@@ -0,0 +1,98 @@
...
+
+function copy_versions_files() {
+    cp "${OPERATOR_BASE_DIR}/version/version.yaml" "${OUTPUT_DIR}"
+    cp "${OPERATOR_BASE_DIR}/version/versions.yaml" "${OUTPUT_DIR}"
+}
+
+copy_installer_charts
+copy_profiles
+copy_versions_files
+
```
sdake: if you rev the PR, might as well remove these extra whitespaces.

release/create_release_charts.sh
```diff
@@ -0,0 +1,98 @@
...
+      echo `git describe --tags`
+      return 0
+    fi
+    echo "${ver}"
+}
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+OPERATOR_BASE_DIR="${SCRIPT_DIR}/.."
+OPERATOR_VERSION=$(get_version)
+TEMP_DIR=${TEMP_DIR:-"$(mktemp -d ${TEMP_DIR_DEFAULT}/istio.${OPERATOR_VERSION}.XXXX)"}
```
sdake: please use more XXXX to avoid collisions, and lowercase them (I am not sure XXXX works with mktemp

release/create_release_charts.sh
ostromart: Yes, that's exactly what this script is doing. We're cloning istio/installer and copying the charts from there.

release/create_release_charts.sh
ostromart: Done.

release/create_release_charts.sh
ostromart: Done. X's do work with mktemp and uppercase X seems to be the most commonly used based on  a quick search.

release/create_release_charts.sh
```diff
@@ -0,0 +1,95 @@
...
+    echo "${ver}"
+}
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+OPERATOR_BASE_DIR="${SCRIPT_DIR}/.."
+OPERATOR_VERSION=$(get_version)
+TEMP_DIR=${TEMP_DIR:-"$(mktemp -d ${TEMP_DIR_DEFAULT}/istio.${OPERATOR_VERSION}.XXXXXXXX)"}
+
+INSTALLER_VERSION=${INSTALLER_VERSION:-"${OPERATOR_VERSION}"}
+OUTPUT_DIR="${OUTPUT_ROOT_DIR}/install/kubernetes/operator"
```
richardwxn: looks like the path is align with what it is in istio/istio, but wondering would it be better to change to something like "${OUTPUT_ROOT_DIR}/install/istio/operator"

release/create_release_charts.sh
```diff
@@ -0,0 +1,95 @@
...
+################################################################################
+
+set -o errexit
+set -o pipefail
+
+OUTPUT_ROOT_DIR=""
+
+TEMP_DIR_DEFAULT="/tmp"
+# TODO: these should be under a subdir
+INSTALLER_CHARTS=(crds gateways istio-cni istiocoredns istio-telemetry istio-control istio-policy security)
```
richardwxn: do we need to have istiocoredns charts as well

release/create_release_charts.sh
ostromart: it should be sitting alongside helm because it's an alternative installation method for k8s.

release/create_release_charts.sh
richardwxn: ignore the above comment, didn't see it from the list

PR #144, opened 2019-08-02
Manifest versions command

ostromart:
This implements manifest version subcommand and adds version string to the top level mesh version command.


version/version.go
```diff
@@ -15,12 +15,16 @@
...
+	goversion "github.com/hashicorp/go-version"
+
 	pkgversion "istio.io/operator/pkg/version"
 )
 
 const (
-	// OperatorBinaryVersionString is the Istio operator version.
-	OperatorBinaryVersionString = "1.3.0"
+	// versionFileName is the name of the file containing the operator version.
+	versionFileName = "./version.yaml"
```
golangcibot: `versionFileName` is unused (from `deadcode`)

pkg/httpget/httpget.go
```diff
@@ -0,0 +1,36 @@
+package httpget
```
morvencao: Please add boilerplate license headers at the beginning.

pkg/httpget/httpget.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+// Get sends an HTTP GET request and returns the result.
+func Get(url string) ([]byte, error) {
```
richardwxn: Consider renaming the package to be something more general like httprequest? Also it is implemented similarly in urlfetcher, maybe we can consolidate them.

pkg/httpget/httpget.go
ostromart: Done.

pkg/httpget/httpget.go
ostromart: Done.

PR #147, opened 2019-08-02
add profile diff and list subcommand.

morvencao:
add profile diff and list subcommand.

### DIFF
--- a/cmd/mesh/profile-list.go
+++ b/cmd/mesh/profile-list.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
+
+	"istio.io/operator/pkg/helm"
 )
 
 func profileListCmd(rootArgs *rootArgs) *cobra.Command {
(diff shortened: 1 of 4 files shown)


cmd/mesh/profile-list.go
```diff
@@ -32,7 +34,16 @@ func profileListCmd(rootArgs *rootArgs) *cobra.Command {
 
 }
 
+// profileList list all the buildin profiles.
 func profileList(args *rootArgs) {
 	checkLogsOrExit(args)
-	fmt.Println("This command is not yet implemented.")
+	profiles := helm.ListBuildinProfiles()
+	if len(profiles) == 0 {
+		fmt.Println("No profile available.")
```
ostromart: s/profile/profiles/

pkg/helm/vfs_renderer.go
```diff
@@ -145,3 +145,14 @@ func stripPrefix(path, prefix string) string {
 	pv := strings.Split(path, "/")
 	return strings.Join(pv[pl:], "/")
 }
+
+// list all the buildin profiles.
+func ListBuildinProfiles() []string {
```
ostromart: ProfileNames is map[string]bool for fast existence check, so values will always be true, you don't have to check for it.
You could make this a more general function (maybe in util) called StringBoolMapToSlice to make the code more reusable.

pkg/helm/vfs_renderer.go
```diff
@@ -145,3 +145,14 @@ func stripPrefix(path, prefix string) string {
 	pv := strings.Split(path, "/")
 	return strings.Join(pv[pl:], "/")
 }
+
+// list all the buildin profiles.
```
ostromart: s/buildin/builtin/

cmd/mesh/profile-list.go
morvencao: done

pkg/helm/vfs_renderer.go
morvencao: done

pkg/helm/vfs_renderer.go
morvencao: done

PR #149, opened 2019-08-03
Fixes for issues discovered during demo

ostromart:
I found lots of breakage before the demo (as expected for demos:-))
This PR contains fixes for run_mandiff to get to a point of reasonable diff output where we can start fixing the diffs. It takes care of high levels problems like different namespaces which were making every resource different, so this gets down to a list of real diffs that need to be investigated and fixed. 
Running make mandiff produces two dirs under /tmp/istio-mandiff-out for the helm and mesh versions of the manifests and a bunch of mandiff-* files for each profile.


richardwxn:
btw I also discovered some issues here: [#16029](https://github.com/istio/istio/issues/16029)

Makefile
```diff
@@ -120,6 +122,16 @@ gen_patch_iscp:
 gen_patch_values:
 	diff -u pkg/apis/istio/v1alpha2/values/values_types.pb.go.orig pkg/apis/istio/v1alpha2/values/values_types.pb.go > pkg/apis/istio/v1alpha2/values/fix_values_structs.patch || true
 
+sync_charts:
```
morvencao: Do we also need to sync the `global.yaml`?

tests/profiles/helm/values-istio-default.yaml
```diff
@@ -1 +1,2 @@
 # default profile will use all the settings in global.yaml
+certmanager: {}
```
morvencao: I found the `certmanager` and `nodeAgent` are enabled for operator with default and dmeo profiles: https://github.com/istio/operator/blob/master/data/profiles/default.yaml#L185-L190
I'll log a separate issue to track that.

scripts/run_mandiff.sh
```diff
@@ -29,110 +61,58 @@ cd "${ROOT}"
...
 
-    local charts=${profile_charts_map[${profile}]}
+    local charts="${PROFILE_CHARTS_MAP[${profile}]}"
     for c in $(echo $charts | tr " " "\n")
     do
-        # create the parent directory if it doesn't exist.
-        mkdir -p $(dirname ${OUT}/helm-template/istio-${namespace}-${relase}-${profile}/${c}.yaml)
-        helm_render_template ${namespace} ${relase} ${chart}/${c} ${cfg} $* > ${OUT}/helm-template/istio-${namespace}-${relase}-${profile}/${c}.yaml
+       echo "Rendering ${c}"
+       mkdir -p "${out_dir}/${c}"
```
morvencao: This will create some empty directory. Can we replace with `mkdir -p $(dirname "${out_dir}/${c}.yaml")`?

Makefile
```diff
@@ -120,6 +122,16 @@ gen_patch_iscp:
...
+	cp -Rf $(charts_dir)/crds $(out_dir)
```
morvencao: need to add step to clone installer repo firstly?

scripts/run_mandiff.sh
ostromart: You mean leave out the subdirectory, since the filename contains the component name? We could do that but I was trying to be consistent with the layout generated with mesh manifest generate for easier diffing in meld.

Makefile
ostromart: Done, I also just moved it to a script since it should always be called with a branch name.

tests/profiles/helm/values-istio-default.yaml
ostromart: looks like you fixed this in https://github.com/istio/operator/pull/153, I will skip it here.

data/charts/crds/templates/crds.yaml
```diff
@@ -1,7 +1,9 @@
 {{ .Files.Get "files/crd-10.yaml" }}
 {{ .Files.Get "files/crd-11.yaml" }}
 {{ .Files.Get "files/crd-12.yaml" }}
+{{- if .Values.certmanager }}
```
sdake: should this change not go in upstream/istio-installer?  I am very concerned we are creating 3 different sets of gotpl instead of 2 looking at this pr.

data/charts/istiocoredns/values.yaml
```diff
@@ -0,0 +1,34 @@
+#
```
sdake: is all of this a copy of the existing charts?  I think it would be more appropriate to copy by reference rather than value :)

pkg/vfsgen/vfsgen_data.go
```diff
@@ -1650,6 +1729,7 @@ var Assets = func() http.FileSystem {
 	}
 	fs["/charts/crds/templates"].(*vfsgen۰DirInfo).entries = []os.FileInfo{
 		fs["/charts/crds/templates/crds.yaml"].(os.FileInfo),
+		fs["/charts/crds/templates/crds.yaml.orig"].(os.FileInfo),
```
sdake: why crds.yaml.orig?

scripts/sync_charts.sh
```diff
@@ -0,0 +1,46 @@
...
+# limitations under the License.
+
+# No unset vars, print commands as they're executed, and exit on any non-zero
+# return code
+set -u
+set -x
+set -e
+
+if [[ "$#" -ne 1 ]]; then
+    echo "Usage: sync_charts.sh <istio/installer branch-name>"
```
sdake: I don't mean to be super negative...  This appears to be a dev tool to sync the charts into the repo.  I am not sure what the advantage of baked in charts is.  I can think of numerous disadvantages.

Can you elaborate on the advantages?

tests/profiles/helm/values-istio-demo-auth.yaml
```diff
@@ -1,7 +1,7 @@
 # This is used to generate minimal demo mode. It is included from demo and demo-auth values.
```
sdake: should profiles not come from the installer repo?  If not, I think that is ok, but we should take an opinion on this in the environments WG meeting.

costinm: There is no such thing as 'profiles' in installer repo - it's supposed to be a-la-carte, each component has its own isolated config.

We do generate kustomize profiles - but as an output.

costinm:
Agree with @sdake, this PR should just remove all templates and instead use directly installer. 
We are not going to ship with the triplicated out-of-sync files.

data/charts/crds/templates/crds.yaml
ostromart: yes, but we're not ready to do that yet, because if something needs to be fixed in installer i don't want to block progress in operator on that. right now there are lots of such fixes.
for the time being i'm just syncing these manually but once installer e2e tests are running with the operator (this week i hope) the plan is to do a final sync and delete this copy of the charts.

data/charts/istiocoredns/values.yaml
ostromart: yes, see comment above.

pkg/vfsgen/vfsgen_data.go
ostromart: i think these ended up in the compiled-in charts by accident. anyway, the compiled in charts option is about to be removed.

scripts/sync_charts.sh
ostromart: regardless of whether they are baked in or not, they can still be sourced from istio/installer, and they will be by end of this week, per my other comment. 
the baked in stuff will be removed now that we have a release process for these charts defined. once that's implemented and working, the baked in stuff will be removed.

tests/profiles/helm/values-istio-demo-auth.yaml
ostromart: they are coming from istio/istio and stored here because they are manually translated.

ostromart:
The copies will be definitely removed by code freeze, we just need a bit more time to fix various problems and sync the two repos up. Having these copied locally allows us to iterate much faster and batch changes back to the installer repo.

costinm:
For demo or local development - you can use a branch or private repo. 

We don't copy files around in master for fast iteration - APIs go into istio/api, installer into istio/installer, etc. If by freeze time it is not ready - it'll go into the next release, making a mess to move faster is not right and code reviewers and maintainers are supposed to catch this.

scripts/run_mandiff.sh
```diff
@@ -29,110 +61,58 @@ cd "${ROOT}"
...
-profile_charts_map[${ISTIO_MINIMAL_PROFILE}]="crds istio-control/istio-discovery"
-profile_charts_map[${ISTIO_SDS_PROFILE}]="crds istio-control/istio-discovery istio-control/istio-config istio-control/istio-autoinject gateways/istio-ingress istio-telemetry/mixer-telemetry istio-policy security/citadel security/nodeagent"
-
-mesh_mandiff_with_profile ${ISTIO_DEFAULT_PROFILE}
-mesh_mandiff_with_profile ${ISTIO_DEMO_PROFILE}
-mesh_mandiff_with_profile ${ISTIO_DEMOAUTH_PROFILE}
-mesh_mandiff_with_profile ${ISTIO_MINIMAL_PROFILE}
-mesh_mandiff_with_profile ${ISTIO_SDS_PROFILE}
+mesh_mandiff_with_profile "${ISTIO_DEFAULT_PROFILE}" > "${OUT}/mandiff-default-profile.diff" || echo "default profile has diffs"
+mesh_mandiff_with_profile "${ISTIO_DEMO_PROFILE}" > "${OUT/}mandiff-demo-profile.diff" || echo "default profile has diffs"
```
richardwxn: the echo message is not accurate, change to corresponding profile names?

ostromart:
Here's the issue tracking this: https://github.com/istio/istio/issues/15493 so I think we are all in agreement. 
Right now (this week) it will be very disruptive to do this, so I hope you will not insist on dropping everything else to work on this at this very moment. Let's finish up the current batch of fixes, sync up the charts and target the merge for this end of this week.

scripts/run_mandiff.sh
morvencao: That makes sense.

morvencao:
We really need this to fix all the gaps found currently, agreed with @ostromart , we will remove the chart copy after all the gaps between `mesh manifest generate...` and `helm template...` get resolved this week.

sdake:
> We really need this to fix all the gaps found currently, agreed with @ostromart , we remove the chart copy after all the gaps between `mesh manifest generate...` and `helm template...` get resolved this week.

If this were to merge, I'd prefer any "gaps" reviewed by upstream istio/installer first.  If the gaps PRs are rejected, what is the process by which the operator repository rectifies that situation?

Cheers
-steve

morvencao:
I think most of the gap PRs should only contain changes in istio/operator part, we should try to keep compiled-in chart and upstream istio/installer unchanged except some critical issues in upstream.

sdake:
> I think most of the gap PRs should only contain changes in istio/operator part, we should try to keep compiled-in chart and upstream istio/installer unchanged except some critical issues in upstream.

I don't quite follow.  Are you proposing to have three different gotpl implementations of install? (1 in operator, 1 in istio/installer, and 1 in the hopefully-soon-to-be deprecated istio/install?)

morvencao:
@sdake Compiled chart in istio/operator is for dev iteration and it will be removed soon. The istio/installer repo should be the only source of truth for chart. 

> If this were to merge, I'd prefer any "gaps" reviewed by upstream istio/installer first. If the gaps PRs are rejected, what is the process by which the operator repository rectifies that situation?

To fix the gaps, we should try to just updating the operator code, instead of the compiled-chart which will be finally synced with upstream istio/installer.

sdake:
> @sdake no, compiled chart in istio/operator is for dev iteration and it will be removed soon. The istio/installer repo should be the only source of truth for chart.
> 
> > If this were to merge, I'd prefer any "gaps" reviewed by upstream istio/installer first. If the gaps PRs are rejected, what is the process by which the operator repository rectifies that situation?
> 
> To fix the gaps, we should try to just updating the operator code, instead of the compiled-chart which will be finally synced with upstream istio/installer.

Spot on...

Regarding your first point, I couldn't agree more.  I think the Env WG leads are in complete agreement that `istio/installer` should be the source of truth for the operator charts and API.

Here are the facts as I understand them:
* istio_controlplane.proto has undergone some API review, but was not approved.
* the operator repo has a variant of the istio-installer charts.
* The operator implementation uses translation to convert the `istio-installer` `values.yaml` file into an `istio_controlplane.proto`.
*  The operator repo has a variant of the `istio-installer` charts baked in to facilitate translation into `istiocontrolpalene.yaml`
* `istio/installer` is the source of truth for *ALL* work to be done within the operator repo.
* `istio/api` is the upstream for all API activity within the Istio project.
* `istio/installer` is the upstream for design activity within the [Istio](https://github.com/orgs/istio/people) project.

Here are the statements others have made about `istiocontrolplane.proto`:
* The `istiocontrolplane.proto` was explicitly rejected as an API as a result of too much complexity.
* The `istiocontrolplane.proto` was reviewed at least 6 times by the config WG and approved.
* Tthe `istiocontrolplane.proto` was reviewed at least 6 times by the config WG but not approved.

Which one is accurate?  I want recorded responses in github on this particular point, as it helps inform next steps with the API.

Here is what I think:
* We can all agree that the `istio/installer` repo should be the source of truth for `istio/operator`.
* I personally believe that deviation of either the charts or the API leads to a bad outcome.
* The API we produce should roughly match that of the `istio/installer` upstream chart API.
* The "next-gen" `values.yaml` API in `istio/installer` should be converted into a proto file with direct mapping with no translation.  That proto should be immediately merged into `istio/api` once produced, and further implementation work should be done there.

If **what I think is the plan we are on**, and this PR helps us get there, I am fine with some iterative work in this repository to get there.  Unlike @costinm, I believe an iterative approach directly in github permits developers to see and share other's work.  I further believe a fork is not a viable option, as it increases developer cognitive overhead especially around source of truth.  However, I additionally believe the project has very specific standards around API definition, and that work should happen within the `istio/api` repository.  To me, forklift of `istio_controlpalene.proto` is not the right answer.

Under the **what I think category** - @costinm if you could weigh in if that matches your expectations (minus the points where we may be in minor disagreement about iterative development of the code, but not the API), then at least the env WG leads will be in alignment.

@ostromart if you could weigh in on the **what I think category**, this will help @costinm and I understand how to guide the next steps and move things along more swiftly - with a **correct > speed** approach.

For other reviewers, if you could weigh in on this comment, your feedback would be greatly appreciated.

Cheers
-steve

ostromart:
@sdake, thanks for taking the time to summarize. Let me try to address all the point you raised as succinctly as possible:
>>> - istio_controlplane.proto has undergone some API review, but was not approved.
Correct. The review process was very long and in the end I think there was agreement between the people involved (there were many) but it never went for an official TOC vote. Had I realized that there was still this level of disagreement I'd have accelerated that but I do think we should put it up to a vote ASAP given this discussion.

>>> the operator repo has a variant of the istio-installer charts.
Yes, temporarily. Initially there was a lot of breakage in the operator and we had to disable whole swaths just to make progress. We worked systematically on converging with istio/installer. Now we are getting close to being 1:1 with istio/installer and our intention is to switch by the end of this week (if we can get unblocked). At that point, our copy will disappear. That has been the plan from the beginning, see https://github.com/istio/istio/issues/15493.

>>> Here are the facts as I understand them:
>>> The operator implementation uses translation to convert the istio-installer values.yaml file into an istio_controlplane.proto.
No, it's the reverse, IstioControlPlane is the API for operator and that converts into values.yaml (and other k8s settings/overlays)

>>> The operator repo has a variant of the istio-installer charts baked in to facilitate translation into istioc_ontrolpalene.proto
Yes, it has baked in charts, but we plan to remove this once the release process is producing artifacts was can use instead. Initially it was thought to be a good way to do a simple "one line" install but I think we can do the same without compiled in charts.
But those charts were also planned to eventually be from istio/installer so they have nothing to do with translation.

>>> istio/installer is the source of truth for ALL work to be done within the operator repo.
For the charts and upgrade patterns, yes. I don't know what else you mean by "all".

>>> istio/api is the upstream for all API activity within the Istio project.
That's where official APIs are published - I'm not sure if that's where we iterate on WIP APIs. 

istio/installer is the upstream for design activity within the Istio project.
>>> Going forward, istio/installer is the source of truth for Helm charts and install/upgrade shape of Istio. istio/istio Helm charts are going to be deprecated - that's the plan afaik at least.

ostromart:
> Here are the statements others have made about istiocontrolplane.proto:
> The istiocontrolplane.proto was explicitly rejected as an API as a result of too much complexity.

Well, @costin explicitly rejected it this Monday in a heated debate we had :-) Other than that I've not heard/been present in any discussion where that happened. 

> The istiocontrolplane.proto was reviewed at least 6 times by the config WG and approved.

It's been reviewed by the community as a whole between 2/27 and 5/8 and had two official API reviews: 3/22 in weekly API review after the TOC meeting and 4/4 (config group). 
There were significant changes after the first meeting, minor changes only after the second.

>> Tthe istiocontrolplane.proto was reviewed at least 6 times by the config WG but not approved.

See point above. Per my earlier comment, that's correct, it has not been formally approved.

ostromart:
> Here is what I think:
> We can all agree that the istio/installer repo should be the source of truth for istio/operator.
> I personally believe that deviation of either the charts or the API leads to a bad outcome.

We all agree and that has been the plan all along. I think the debate in this PR is whether changes we are making atm should be only upstream and I don't think that's feasible at this point. We can double commit the changes or just istio/operator and batch them when syncing back to istio/installer. The latter is what I've been doing and would be my preference. 

> The API we produce should roughly match that of the istio/installer upstream chart API.

Operator API passes through the exact Helm API to the charts (after validation), so it matches that part exactly. 
There are other parts of the API (k8s settings, namespaces, feature enablement) which are necessarily part of the operator API. 

> The "next-gen" values.yaml API in istio/installer should be converted into a proto file with direct mapping with no translation. That proto should be immediately merged into istio/api once produced, and further implementation work should be done there.

I'm not sure if there's going to be a next-gen Helm API, rather I believe it will disappear in favor of direct config of components and the operator API (for k8s settings).

sdake:
Given today's recommendation by @smawson to produce a mediation document for TOC to make a decision, I see no further reason to hold up this review.  I have aired my complaints in the PR thus far, and they serve as a record.  The current plan we are on is still in effect.  If the TOC wants us to shift the rudder, we will do so.

Please rebase the PR so it may be approved.

Cheers
-steve

PR #152, opened 2019-08-03
Fix unwanted change during merge

richardwxn:
There were couple changes in manifest-diff file reverted during rebase for commit #132. Bring those back.


richardwxn:
@sdake Can you help approve it?

PR #153, opened 2019-08-04
conditionize initing features and components.

morvencao:
Resolve: https://github.com/istio/istio/issues/16045


morvencao:
This will also fix https://github.com/istio/istio/issues/16041

pkg/component/feature/feature.go
```diff
@@ -57,8 +80,15 @@ func NewBaseFeature(opts *Options) *BaseFeature {
 	cff := &CommonFeatureFields{
 		Options: *opts,
 	}
-	cff.components = []component.IstioComponent{
-		component.NewCRDComponent(newComponentOptions(cff, name.IstioBaseFeatureName)),
+
+	if opts != nil && opts.Traslator != nil && opts.Traslator.FeatureMaps != nil {
```
ostromart: s/Traslator/Translator/

pkg/component/feature/feature.go
```diff
@@ -116,11 +153,17 @@ func NewSecurityFeature(opts *Options) *SecurityFeature {
 	cff := &CommonFeatureFields{
 		Options: *opts,
 	}
-	cff.components = []component.IstioComponent{
-		component.NewCitadelComponent(newComponentOptions(cff, name.SecurityFeatureName)),
-		component.NewCertManagerComponent(newComponentOptions(cff, name.SecurityFeatureName)),
-		component.NewNodeAgentComponent(newComponentOptions(cff, name.SecurityFeatureName)),
+
+	if opts != nil && opts.Traslator != nil && opts.Traslator.FeatureMaps != nil {
```
ostromart: This code looks to be repeated - could you put it in a function?

pkg/translate/translate.go
```diff
@@ -445,6 +481,15 @@ func (t *Translator) setEnablementAndNamespaces(root map[string]interface{}, icp
 	return nil
 }
 
+// IsFeatureEnabled reports whether the feature with name ft is enabled, according to the translations in t,
```
ostromart: I'm not sure if this is necessary - IsComponentEnabled takes feature enablement into account and we don't really check enablement at feature level on its own. Unless you have a specific use for this in mind I recommend leaving it out.

pkg/component/feature/feature.go
morvencao: Done.

pkg/component/feature/feature.go
morvencao: Good idea, done.

pkg/translate/translate.go
morvencao: I tried to figure out a solution to reuse the `IsComponentEnabled` function during initing the  `IstioControlPlane`, https://github.com/istio/operator/blob/9027646f6009e6b39135ca7505cf1d9f27ef8aea/pkg/component/controlplane/control_plane.go#L39-L45
Then I found that the most efficient method is to add `IsFeatureEnabled` function.

pkg/translate/translate.go
```diff
@@ -180,6 +190,43 @@ var (
...
+				},
+				name.TelemetryFeatureName: {
+					Components: []name.ComponentName{
+						name.TelemetryComponentName,
+						name.PrometheusComponentName,
+						name.PrometheusOperatorComponentName,
+						name.GrafanaComponentName,
+						name.KialiComponentName,
+						name.TracingComponentName,
+						},
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
					},
```

pkg/component/feature/feature.go
```diff
@@ -344,3 +313,20 @@ func renderComponents(cs []component.IstioComponent) (manifests name.ManifestMap
 	}
 	return
 }
+
+// buildCommonFeatureFields is an internal function to build the Common Feature Fields for specified feature.
+func buildCommonFeatureFields(opts *Options, ftname name.FeatureName) *CommonFeatureFields {
+	cff := &CommonFeatureFields{
+		Options: *opts,
+	}
+	if opts != nil && opts.Translator != nil && opts.Translator.FeatureMaps != nil {
```
ostromart: it's better to rewrite this condition as as the logical opposite and return if true - this reduces indentation.

morvencao: done.

telemetry-values.yaml
```diff
@@ -0,0 +1,284 @@
+certmanager:
```
ostromart: what is the purpose of this new file?

morvencao: sorry, it's my local test file, will remote that.

morvencao: removed this file.

PR #155, opened 2019-08-05
Get mac building working properly

sdake:
Use reallink to read the localtime path
Temporarily turn off timezone to determine if it is necessary
Output the binaries /work/out/bin

This last solution^H^H^^Hack is a little problematic.  We need
to determine a cononical place to put our binaries in the event
a toolchain is not present, or output them in $(GOPATH)/bin if
the toolchain is present or configured.

This last point needs some thought.

This PR should be merged as is to unblock Mac developers.

### DIFF
--- a/Makefile.common.mk
+++ b/Makefile.common.mk
@@ -22,19 +22,19 @@
 IMG = docker.io/sdake/build-tools:2019-08-03
 UID = $(shell id -u)
 PWD = $(shell pwd)
-GOBIN ?= $(GOPATH)/bin
+GOBIN_SOURCE ?= $(GOPATH)/bin
+GOBIN ?= /work/out/bin
 
 RUN = docker run -t --sig-proxy=true -u $(UID) --rm \
 	-v /etc/passwd:/etc/passwd:ro \
-	-v /etc/passwd:/etc/passwd:ro \
-	-v /etc/localtime:/etc/localtime:ro \
-	-v /etc/timezeone:/etc/timezeone:ro \
+	-v $(reallink /etc/localtime):/etc/localtime:ro \
 	--mount type=bind,source="$(PWD)",destination="/work" \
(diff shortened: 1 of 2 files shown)


Makefile.common.mk
```diff
@@ -22,19 +22,19 @@
 IMG = docker.io/sdake/build-tools:2019-08-03
 UID = $(shell id -u)
 PWD = $(shell pwd)
-GOBIN ?= $(GOPATH)/bin
+GOBIN_SOURCE ?= $(GOPATH)/bin
+GOBIN ?= /work/out/bin
```
ostromart: I think it's fine to get thing running again but could you leave a TODO or create an issue to fix this properly?

sdake: Sounds fair.  I created an issue instead of TODO so this can be  tracked properly.

https://github.com/istio/istio/issues/16056

Makefile.common.mk
```diff
@@ -22,19 +22,19 @@
...
 
 RUN = docker run -t --sig-proxy=true -u $(UID) --rm \
 	-v /etc/passwd:/etc/passwd:ro \
-	-v /etc/passwd:/etc/passwd:ro \
-	-v /etc/localtime:/etc/localtime:ro \
-	-v /etc/timezeone:/etc/timezeone:ro \
+	-v $(reallink /etc/localtime):/etc/localtime:ro \
```
ericvn: Should this have been readlink and not reallink?

sdake: `readlink` is not available on mac without brew.  `reallink` is available on both Mac and Linux.

Goal of the build system is make + docker and you are GTG.

Cheers
-steve

ericvn: Interesting as I don't have `reallink` on my Mac, but had installed `readlink` earlier.

PR #157, opened 2019-08-05
Imlementation of HTTP Template Renderer

richardwxn:


pkg/helm/helm.go
```diff
@@ -61,6 +65,9 @@ func NewHelmRenderer(chartsRootDir, helmBaseDir, profile, componentName, namespa
 	switch {
 	case chartsRootDir == "":
 		return NewVFSRenderer(helmBaseDir, globalValues, componentName, namespace), nil
+	case util.IsHTTPURL(dir):
+		return NewHTTPTemplateRenderer(dir, globalValues, componentName, namespace,
+			time.Duration(TemplatePollingIntervalInMinute*time.Minute)), nil
```
golangcibot: unnecessary conversion (from `unconvert`)

pkg/helm/urlwatcher.go
```diff
@@ -91,16 +95,17 @@ func NewPoller(dirURL string, destDir string, interval time.Duration) *Poller {
 
 //PollURL continuously polls the given url, which points to a directory containing an
 //installation package at the given interval and fetches a new copy if it is updated.
-func PollURL(dirURL string, interval time.Duration) error {
+func PollURL(dirURL string, interval time.Duration) (chan struct{}, error) {
```
morvencao: suggest to return read only channel: `<-chan struct{}`

pkg/helm/urlwatcher.go
```diff
@@ -49,33 +49,37 @@ const (
...
 		p.existingHash = newHash
-		return uf.fetchChart(shaF)
+		return true, uf.fetchChart(shaF)
 	}
-	return nil
+	return false, nil
 }
 
-func (p *Poller) poll() {
+func (p *Poller) poll(notify chan struct{}) {
```
morvencao: better to pass in a write-only channel: `notify chan<- struct{}`

ostromart: Also, I recommend creating a channel and returning it from this function.

pkg/helm/urlwatcher.go
```diff
@@ -49,33 +49,37 @@ const (
 
 // checkUpdate checks a SHA URL to determine if the installation package has been updated
 // and fetches the new version if necessary.
-func (p *Poller) checkUpdate() error {
+func (p *Poller) checkUpdate() (bool, error) {
```
ostromart: suggest renaming Poller to URLPoller to be more clear about its functionality

pkg/helm/http_renderer.go
```diff
@@ -0,0 +1,89 @@
...
+	helmChartDirPath string
+	chart            *chart.Chart
+	started          bool
+	globalValues     string
+	pollInterval     time.Duration
+}
+
+// NewHTTPTemplateRenderer creates a TemplateRenderer with the given parameters and returns a pointer to it.
+// helmChartDirPath must be an absolute HTTP URL hosting root of helm charts.
+func NewHTTPTemplateRenderer(helmChartDirPath, globalValues, componentName, namespace string, pollInterval time.Duration) *HTTPTemplateRenderer {
```
ostromart: we also need a non-polling mode. somewhere in the comments you should describe what the HTTPTemplateRenderer actually does.

pkg/helm/http_renderer.go
```diff
@@ -0,0 +1,89 @@
...
+	"k8s.io/helm/pkg/chartutil"
+	"k8s.io/helm/pkg/proto/hapi/chart"
+
+	"istio.io/pkg/log"
+)
+
+// HTTPTemplateRenderer is a helm template renderer for helm charts hosted on http url
+type HTTPTemplateRenderer struct {
+	namespace        string
+	componentName    string
```
ostromart: We need to discuss whether it makes sense to have a HTTPTemplateRenderer vs. doing the fetching externally and passing params to the file renderer. 
The reason is that we'll be releasing a build image with all the charts as a single tar and only one poller should be checking that, not every component.

richardwxn:
Closing this PR as HTTP template renderer is not needed, file template renderer should handle the installation when install package point to a http url

PR #161, opened 2019-08-06
Prom fix

ostromart:
This adds the remaining components (prom, grafana, kiali, CNI, tracing) which aren't part of the API but are in charts and therefore are required for passing through from Helm API.
Note: This is based off https://github.com/istio/operator/pull/149, those diffs will disappear when this is rebased.


pkg/helm/helm.go
```diff
@@ -50,19 +50,14 @@ type TemplateRenderer interface {
 // NewHelmRenderer creates a new helm renderer with the given parameters and returns an interface to it.
 // The format of helmBaseDir and profile strings determines the type of helm renderer returned (compiled-in, file,
 // HTTP etc.)
-func NewHelmRenderer(chartsRootDir, helmBaseDir, profile, componentName, namespace string) (TemplateRenderer, error) {
-	globalValues, err := ReadValuesYAML(profile)
```
morvencao: I think removing the `profile` here will result in plenty of empty fields, the chart renderer will fail unless we manually override all these fields, please check the prow test job: https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_operator/161/operator-test/88

pkg/component/feature/feature.go
```diff
@@ -277,6 +277,41 @@ func (f *GatewayFeature) RenderManifest() (name.ManifestMap, util.Errors) {
 	return renderComponents(f.components)
 }
 
+// ThirdPartyFeature is the traffic management feature.
```
ayj: "... is the traffic management feature" - typo?

ostromart:
Added DNM till https://github.com/istio/operator/pull/149 is merged.

PR #162, opened 2019-08-06
Fix mac building

sdake:
A couple notes:

Any binary tool that needs to be used should first be installed in
the istio/tools/docker/build-tools/Dockerfile.  Any go generate
should not compile, as this won't work cross platform.

### DIFF
--- a/Makefile.common.mk
+++ b/Makefile.common.mk
@@ -19,13 +19,38 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-IMG = gcr.io/istio-testing/build-tools:2019-08-05
+IMG = docker.io/sdake/build-tools:2019-08-06
 UID = $(shell id -u)
 PWD = $(shell pwd)
 GOBIN_SOURCE ?= $(GOPATH)/bin
 GOBIN ?= /work/out/bin
 
+LOCAL_ARCH := $(shell uname -m)
+ifeq ($(LOCAL_ARCH),x86_64)
+GOARCH_LOCAL := amd64
+else ifeq ($(shell echo $(LOCAL_ARCH) | head -c 5),armv8)
+GOARCH_LOCAL := arm64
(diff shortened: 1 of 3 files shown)


Makefile.common.mk
```diff
@@ -19,13 +19,38 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-IMG = gcr.io/istio-testing/build-tools:2019-08-05
+IMG = docker.io/sdake/build-tools:2019-08-06
```
ericvn: Do you really want to use the image from your repo? Just not changed after testing?

cmd/mesh.go
```diff
@@ -11,7 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-//go:generate go run ./vfsgen/vfsgen.go
+//go:generate
```
ericvn: Can the comment be removed since it's empty and not specifying anything? Or just a placeholder for something coming in a future PR?

Makefile.common.mk
sdake: @geeknoid indicated he would push the image this morning.  Unfortunately it wasn't pushed and I don't desire the repo to be blocked all day.

We can follow this up with a PR to change it to the official image once @geeknoid is back in the office.

cmd/mesh.go
sdake: will remove thanks for catching.

sdake: Actually, I am not sure if the comment can be removed.  I believe `// go generate` is needed to get vfsgen to operate on this file.

ostromart: Yes, this is a go generate directive and we have to keep it for the time being. Once the charts release process is working we will get rid of the built in charts.

PR #164, opened 2019-08-06
Update common files.

geeknoid:


ericvn:
@geeknoid I'll let @sdake decide if this makes sense to pull in since it'll revert some of his changes. I quickly worked on sort of a reverse idea and will create a PR in _common-files_ with some of those changes, which once merged can be pulled into _operator_.

PR #165, opened 2019-08-06
refactor commands for istioctl integration

ayj:
* Move usage of the unexported `rootArgs` struct into the profile and manifest command creation functions. 

* Move the `versions` subcommand under the manfiest command to match [CLI reference ](https://docs.google.com/document/d/16SWW14zrwFKc9OzH6biMQD1u9INllzMa56jde9vgQkM/edit#heading=h.7bqqr4ys5h17)

* Ran `go mod tidy`


cmd/mesh/manifest.go
```diff
@@ -71,6 +73,8 @@ func ManifestCmd(args *rootArgs) *cobra.Command {
 	mc.AddCommand(mmc)
 	mc.AddCommand(mvc)
 
+	mc.AddCommand(version.CobraCommand())
```
morvencao: The manifest versions subcommand should be already implemented by https://github.com/istio/operator/blob/c9c8bb2c59585f738b1c0563cf9f5fd31491fc81/cmd/mesh/manifest-versions.go
flags are added at: https://github.com/istio/operator/blob/c9c8bb2c59585f738b1c0563cf9f5fd31491fc81/cmd/mesh/manifest.go#L54

And this version is more like a general version for `mesh`.

ayj: Thanks - fixed.

PR #168, opened 2019-08-07
This is unneeded and probably incorrect.

sdake:
All common makefile work should go in Makefile, not in
Makefile.core.mk files.


sdake:
@martonsereg that is in alignment with the work that @ericvn is doing in https://github.com/istio/common-files/pull/24

PR #169, opened 2019-08-07
Update makefile for proto related rules

richardwxn:
1. Clean up the makefile for proto related rules, add README for patch instruction
2. Fetch the k8s protos dependencies to /tmp for it to work with container. Also update the proto path to include common image's shared protos
3. replace gofast plugin with gogofast plugin, which helps to fix the enum field unmarshalling issue


Makefile
```diff
@@ -36,6 +36,7 @@ GOBIN ?= /work/out/bin
 RUN = docker run -t --sig-proxy=true -u $(UID) --rm \
 	-v /etc/passwd:/etc/passwd:ro \
 	-v $(readlink /etc/localtime):/etc/localtime:ro \
+	-v $(GOPATH)/src/istio.io/operator:/go/src/istio.io/operator \
```
sdake: this will be overwritten when `make common` is run to update the makefiles, which may be often.

Makefile.core.mk
```diff
@@ -78,37 +78,29 @@ clean-proto:
...
 
-# get imported protos to $GOPATH
 get_dep_proto:
-	GO111MODULE=off go get k8s.io/api/core/v1 k8s.io/api/autoscaling/v2beta1 k8s.io/apimachinery/pkg/apis/meta/v1/ github.com/gogo/protobuf/...
+	GO111MODULE=off GOPATH=/tmp go get k8s.io/api/core/v1 k8s.io/api/autoscaling/v2beta1 k8s.io/apimachinery/pkg/apis/meta/v1/ github.com/gogo/protobuf/...
 
-proto_iscp:
-	protoc -I=${GOPATH}/src -I./pkg/apis/istio/v1alpha2/ --proto_path=pkg/apis/istio/v1alpha2/ --gofast_out=pkg/apis/istio/v1alpha2/ pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
+proto_iscp: get_dep_proto
+	protoc -I=/tmp/src -I=${GOPATH}/src/ -I=/usr/include/protobuf --gogofast_out=${GOPATH}/src ${GOPATH}/src/istio.io/operator/pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```
sdake: the first part looks good.  the ${GOPATH}/src - understand what your after, I 'll need to think a little bit about how we can generalize this.  I just woke up so I am not super sharp atm, but I think we need a system host bindmount on either `${GOPATH}` or `/`.`${GOPATH}`may work for our use case with operator repo, but may not work for every use case.  THe bindmount should be `rshared` - so will need to test if that will work as I expect.

I guess we can start with that, and expand.

Cheers
-steve

Makefile.core.mk
```diff
@@ -78,37 +78,29 @@ clean-proto:
...
 	sed -i -e 's|github.com/gogo/protobuf/protobuf/google/protobuf|github.com/gogo/protobuf/types|g' pkg/apis/istio/v1alpha2/istiocontrolplane_types.pb.go
 	patch pkg/apis/istio/v1alpha2/istiocontrolplane_types.pb.go < pkg/apis/istio/v1alpha2/fixup_go_structs.patch
 
-proto_iscp_orig:
-	protoc -I=${GOPATH}/src -I./pkg/apis/istio/v1alpha2/ --proto_path=pkg/apis/istio/v1alpha2/ --gofast_out=pkg/apis/istio/v1alpha2/ pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
+proto_iscp_orig: get_dep_proto
```
sdake: do we need this variant, or can we just delete it?

Makefile.core.mk
```diff
@@ -78,37 +78,29 @@ clean-proto:
 
 # is all of this needed or is this cruft?
 
-# get imported protos to $GOPATH
 get_dep_proto:
-	GO111MODULE=off go get k8s.io/api/core/v1 k8s.io/api/autoscaling/v2beta1 k8s.io/apimachinery/pkg/apis/meta/v1/ github.com/gogo/protobuf/...
+	GO111MODULE=off GOPATH=/tmp go get k8s.io/api/core/v1 k8s.io/api/autoscaling/v2beta1 k8s.io/apimachinery/pkg/apis/meta/v1/ github.com/gogo/protobuf/...
```
sdake: interesting solutoin, although these deps are pulled automatically by for example `make download`or are already likely in the image .  Protobuf itself is in the expected systemwide location.

Makefile.core.mk
ostromart: i think we should get rid of it. for cases where the patch file needs to be updated we can leave a README in the proto dir on how to update it. this just clutters up the makefile and is confusing to the average user.

Makefile.core.mk
```diff
@@ -78,37 +78,29 @@ clean-proto:
...
 
-proto_iscporig_setup: get_dep_proto proto_iscp_orig
-
-proto_iscp_setup: get_dep_proto proto_iscp
-
-proto_values_setup: get_dep_proto proto_values
-
-proto_valuesorig_setup: get_dep_proto proto_values_orig
 
 gen_patch_iscp:
```
ostromart: same here - could we get rid of the gen_patch commands and move the process to a README in the proto dir?

Makefile.core.mk
richardwxn: ok I will go remove it and add README

Makefile.core.mk
richardwxn: The k8s deps are not available in the common image now. Actually pulling in those deps to /tmp using go get does not require changes to existing proto import paths.

Makefile.core.mk
sdake: agreed any simplification of the `Makefile.core.mk` file is beneficial.  If things remain, they should be commented to indicate their purpose.

Makefile
sdake: I edited the above comment to indicate the correct make target that results in an overwrite of the Makefile.

Makefile.core.mk
sdake: The common image does include parts of `gogo/protobuf`: https://github.com/istio/tools/blob/master/docker/build-tools/Dockerfile#L46-L49

if parts of `gogo/protbuf` are not present, perhaps you could improve thecommon-tools image to include what is needed?  We really want to pin any dependencies needed using either go.mod or build-tools.

The dependencies are in `go.mod` referenced as `k8s.io/api` and `k8s.io/apimachinery`.  Do you mean these are not pulled into the go package cache when using the repository?  Or just that using `-I` is not desireable with `protoc`?

Makefile
richardwxn: ok I will move this logic to core instead

Makefile.core.mk
richardwxn: the protoc include path does not handle go package cache properly, that's why I have to pull those deps to temp directory

richardwxn:
@sdake Can you help please take another look?

Makefile
```diff
@@ -71,7 +71,7 @@ else
 export GOBIN ?= ./out/bin
 endif
 
-MAKE = $(RUN) make -e -f Makefile.core.mk
+MAKE = $(RUN) make --no-print-directory -e -f Makefile.core.mk
```
sdake: This looks a little redundant: https://github.com/istio/common-files/commit/4328839147bab7f10dd5790bd95c275482de83d2

pkg/apis/istio/v1alpha2/README.md
```diff
@@ -0,0 +1,18 @@
+### proto file patch instruction
+Currently there are two proto files defined temporarily in istio/operator: istiocontrolplane_types.proto and values.types.proto, which would be moved to istio/api after 1.3
+
+There are some type of fields in the original proto files, e.g. []map[string]interface{}, cannot be handled properly by protoc,
+so we comment out this fields and patch them back in the generated pb.go files
```
sdake: zug...

Makefile
richardwxn: removed

richardwxn:
> Thanks for the work!
> 
> I'm not sure this PR is quite right, but it removes a bunch of cruft, so approving - and we can iterate towards a better solution that works with CONTAINER_BUILD=1.
> 
> Cheers
> -steve

Thanks for checking. Now it works with both container and local.

PR #171, opened 2019-08-08
tune profiles to fix gaps.

morvencao:
- fix broken `manifest generate` with `demo`, `minimal` and `sds` profiles
- update buildin profiles to lower gaps with `hem template...`


data/profiles/default.yaml
```diff
@@ -502,53 +505,3 @@ spec:
       tolerations: []
       podAntiAffinityLabelSelector: []
       podAntiAffinityTermLabelSelector: []
-
-    grafana:
```
ostromart: I think this needs to remain doesn't it?

morvencao: I thought `grafana` is disabled in `default` profile, isn't it? https://github.com/istio/istio/blob/master/install/kubernetes/helm/istio/values.yaml#L65

PR #175, opened 2019-08-09
set the default target of make to mesh

waret:
export GO111MODULE=on, so we can run `make` command when this repo is in GOPATH.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+export GO111MODULE=on
 RUN =
 
 # Set the environment variable BUILD_WITH_CONTAINER to use a container
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -63,6 +63,8 @@ api_protos := $(shell find $(api_path) -type f -name '*.proto' | sort)
 api_pb_gos := $(api_protos:.proto=.pb.go)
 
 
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -19,6 +19,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+export GO111MODULE=on
```
sdake: Makefile should not be edited directly.  See top of this file for more details.

Makefile.core.mk
```diff
@@ -63,6 +63,8 @@ api_protos := $(shell find $(api_path) -type f -name '*.proto' | sort)
 api_pb_gos := $(api_protos:.proto=.pb.go)
 
 
+.PHONY: default
+.DEFAULT_GOAL :=
```
sdake: first time I've seen this in a Makefile.  The upstream documentation states: ```
.DEFAULT_GOAL
Sets the default goal to be used if no targets were specified on the command line (see Arguments to Specify the Goals). The .DEFAULT_GOAL variable allows you to discover the current default goal, restart the default goal selection algorithm by clearing its value, or to explicitly set the default goal. The following example illustrates these cases:
```

I don't understand why this is needed.  Could you expand?

waret: This is a bad submission. Sorry for the trouble, maybe I should close this pr.

PR #177, opened 2019-08-09
Update README

ostromart:
Fix some bad yamls and add links to samples.


samples/trafficmgmt-namespace.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: install.istio.io/v1alpha2
```
richardwxn: can we use traffiicmanagement instead of trafficmgmt

samples/policy-off.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: install.istio.io/v1alpha2
```
richardwxn: A general comment about the sample configuration files name, would it be more clear to rename them as {base-profile}-{customization part}, e.g: sds-policy-off

README.md
```diff
@@ -144,7 +164,7 @@ mesh profile dump --set profile=minimal
 
 #### Select a specific configuration profile
 
-The simplest customization is to select a profile different to `default` e.g. `sds`. Create the following config file:
+The simplest customization is to select a profile different to `default` e.g. `sds`. See [samples/sds.yaml](samples/sds.yaml):
 
 ```yaml
```
richardwxn: Do we still need to keep the below part or just pointing to samples is enough

ostromart: I think we should keep the text if users just want to read it.

samples/policy-off.yaml
ostromart: Done.

samples/trafficmgmt-namespace.yaml
ostromart: Done.

PR #178, opened 2019-08-09
Update the README file

richardwxn:
try to make the description clear between actual configuration profiles and provided base profiles

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ for deprecation.
 This repo currently provides pre-configured Helm values sets for different scenarios as configuration
 [profiles](https://istio.io/docs/setup/kubernetes/additional-setup/config-profiles/), which act as a starting point for
 an Istio install and can be customized by creating customization overlay files or passing parameters when
-calling Helm. Similarly, the operator API uses the same profiles (expressed internally through the new API, e.g. compiled in [base profiles files](https://github.com/istio/operator/tree/master/data/profiles)), which can be selected
+calling Helm. Similarly, the operator API uses the same profiles (expressed internally through the new API), which can be selected
 as a starting point for the installation. For comparison, the following example shows the command needed to install
 Istio using the SDS configuration profile using Helm:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -142,9 +142,9 @@ mesh profile dump --set profile=minimal
 ```
 
 
-#### Select a specific configuration profile
+#### Select a specific base profile
```
ostromart: Let's stick with istio.io terminology, which is configuration profile. I also thought about using "base" but I believe it may be confusing to users.

README.md
```diff
@@ -32,7 +32,7 @@ for deprecation.
 This repo currently provides pre-configured Helm values sets for different scenarios as configuration
 [profiles](https://istio.io/docs/setup/kubernetes/additional-setup/config-profiles/), which act as a starting point for
 an Istio install and can be customized by creating customization overlay files or passing parameters when
-calling Helm. Similarly, the operator API uses the same profiles (expressed internally through the new API), which can be selected
+calling Helm. Similarly, the operator API uses the same profiles (expressed internally through the new API, e.g. compiled in [base profiles files](https://github.com/istio/operator/tree/master/data/profiles)), which can be selected
```
ostromart: We probably won't be shipping with compiled in.

README.md
richardwxn: I think what we have right now is not clear enough though, actually the configuration profile should refer to the base+different layers of overlays
maybe change it to sth like base configuration profile?

README.md
richardwxn: ok I would remove this part. Actually I hope to add some pointers to a list of base profiles that we provides. Maybe that would be added to istio.io?

README.md
ostromart: "configuration profile" in istio.io terminology means the built-in profiles, without any customization. We've started using "base profile" as a term in operator but for Istio users in general they understand the concept of "configuration profile" but "base" will confuse them.

README.md
ostromart: It's already in istio.io: https://istio.io/docs/setup/kubernetes/additional-setup/config-profiles/
We just need to stay in sync with these. We also have istioctl profile list that dumps the profiles for any given version of the operator.

richardwxn: yep but some of the contents are not operator specific, for example the profile file name is referring to values.yaml file name

ostromart: ok, you convinced me :-)

ostromart: but please remove the compiled in bit since that's somewhat up in the air.

PR #180, opened 2019-08-10
Eliminate use of vfsgen

geeknoid:
- Remove use of vfsgen and replace in favor of go-bindata which we use in several
other places in Istio. This removes several lines of code, improves access performance,
and eliminates a snowflake case.


ostromart:
Removing cmd/vfsgen/vfsgen.go means that an external tool will have to be installed. Imo it's better to keep this file since it's essentially a self-installing vfsgen.

geeknoid:
@ostromart The source code for the tool has been moved to the tools repo and and slightly generalized (https://github.com/istio/tools/pull/255).

We want all build tools to be in the build-tools container and we expect everyone will be building using the container bits. As a general rule, if you don't build using the build-tools container, then you will be responsible for ensuring your environment has all the right tools present.

Building with the container, you environment will need git, make, and docker and that's it. If you don't use the container, then you need to install the right version of go, various tools from the tools repo, gogo plugins, golangci-lint, goimports, and more. vfsgen is not special in that way, it's just another tool.

To get vfsgen working outside the container, you simply need "go get istio.io/tools/cmd/vfsgen".

geeknoid:
I've updated this PR to completely eliminate the use of vfsgen. Elsewhere in Istio we use go-bindata to do the same kind of thing, so it makes sense to use that here too (especially since it ends up saving code).

PTAL

costinm:
In general I don't mind checkin in generated code - but in this case it may be a bit too big. 

AFAIK the long-term plan is to use istioctl, built as part of the release download binary, as well as a in-cluster operator. 

In both cases the actual templates will be included as files - either as part of a docker image or in the .tar.gz ( where they can also be used with helm template or kustomize ). 

What is the value of including them in the binary ( and in a separate PR to implement a downloader and package manager) ? Please confirm with the release team as well.

geeknoid:
@costinm This PR is not changing any behavior. It is simply changing how the content is being included in the binary. It was included in the binary before this PR and it still is.

geeknoid:
@ostromart You  set the do not merge label on this PR. Do you have more concerns to address?

ostromart:
@geeknoid I have to go through and try this out but am completely swamped due to the release. Just looking at the code, I don't see an Asset() in my generated code so I'd like to understand better what that is.
 Since this is not a critical fix, can it wait till next week to go in?

geeknoid:
There's no urgency, we're just trying to nail the container-based-build thing.

costinm:
Sure, but instead of doing more work and testing on a new generation method
- it is better to just stop using generation and simplify both the build
and the codebase.

ostromart:
This seems fine to me (I still need to fully try it out) but something to consider:
https://tech.townsourced.com/post/embedding-static-files-in-go/
This is why I originally went with vfsgen and we may want to consider it as a replacement. However if go-bindata is working for us I don't have any problem with continuing to use it.

geeknoid:
@ostromart Nice article. It does sound like for a web app scenario, vfsgen is a better choice. But for our use case, go-bindata serves the purpose and is simpler.

sdake:
related PR: https://github.com/istio/operator/pull/190

sdake:
@ostromart if 190 merges, doese that obliviate the need for this PR and vfsgen in genereal?

Cheers
-steve

sdake:
@costinm I'm good with a bit of an incremental approach to removing the code gen.  IMO it needs to go - it is hard to reason about.  I think @ostromart is on track for doing that work.

PR #181, opened 2019-08-11
add select and ignore flags for manifest diff command.

morvencao:
Resolve: https://github.com/istio/istio/issues/16178

### DIFF
--- a/cmd/mesh/manifest-diff.go
+++ b/cmd/mesh/manifest-diff.go
@@ -33,23 +33,27 @@ const YAMLSuffix = ".yaml"
 type manifestDiffArgs struct {
 	// compareDir indicates comparison between directory.
 	compareDir bool
-	// selection specifies the resources to compare, it's comma-separated of resource indicators
-	// in the form of "<kind>:<namespace>:<name>,<kind>:<namespace>:<name>".
-	selection string
-	// ignore specifies the resources to ignore, it's comma-separated of resource indicators
-	// in the form of "<kind>:<namespace>:<name>,<kind>:<namespace>:<name>".
-	ignore string
+	// selectResources constrains the list of resources to compare to only the ones in this list, ignoring all others.
+	// The format of each list item is :: and the items are comma separated. The * character represents wildcard selection.
+	// e.g.
+	// Deployment:istio-system:* - compare all deployments in istio-system namespace
+	// Service:*:istio-pilot - compare Services called "istio-pilot" in all namespaces.
+	selectResources string
+	// ignoreResources ignores all listed items during comparison. It uses the same list format as selectResources.
+	ignoreResources string
 }
 
 func addManifestDiffFlags(cmd *cobra.Command, diffArgs *manifestDiffArgs) {
 	cmd.PersistentFlags().BoolVarP(&diffArgs.compareDir, "directory", "r",
 		false, "compare directory")
-	cmd.PersistentFlags().StringVar(&diffArgs.selection, "select", "::",
-		"specifies the resources to compare, it's comma-separated of resource indicators"+
-			"in the form of \"<kind>:<namespace>:<name>,<kind>:<namespace>:<name>\".")
-	cmd.PersistentFlags().StringVar(&diffArgs.ignore, "ignore", "",
-		"specifies the resources to ignore, it's comma-separated of resource indicators"+
-			"in the form of \"<kind>:<namespace>:<name>,<kind>:<namespace>:<name>\".")
+	cmd.PersistentFlags().StringVar(&diffArgs.selectResources, "select", "::",
+		"selectResources constrains the list of resources to compare to only the ones in this list, ignoring all others.\n"+
+			"The format of each list item is \"::\" and the items are comma separated. The \"*\" character represents wildcard selection.\n"+
+			"e.g.\n"+
+			"    Deployment:istio-system:* - compare all deployments in istio-system namespace\n"+
+			"    Service:*:istio-pilot - compare Services called \"istio-pilot\" in all namespaces.")
+	cmd.PersistentFlags().StringVar(&diffArgs.ignoreResources, "ignore", "",
+		"ignoreResources ignores all listed items during comparison. It uses the same list format as selectResources.")
 }
 
 func manifestDiffCmd(rootArgs *rootArgs, diffArgs *manifestDiffArgs) *cobra.Command {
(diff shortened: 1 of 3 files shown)


cmd/mesh/manifest-diff.go
```diff
@@ -33,11 +33,23 @@ const YAMLSuffix = ".yaml"
 type manifestDiffArgs struct {
 	// compareDir indicates comparison between directory.
 	compareDir bool
+	// selection specifies the resources to compare, it's comma-separated of resource indicators
+	// in the form of "<kind>:<namespace>:<name>,<kind>:<namespace>:<name>".
```
ostromart: suggest changing names to selectResources and ignoreResources, as selection is a bit confusing.
also, i propose the comment to read "selectResources constrains the list of resources to compare to only the ones in this list, ignoring all others. The format of each list item is <kind>:<namespace>:<name> and the items are comma separated. The * character represents wildcard selection. e.g. 
Deployment:istio-system:* - compare all deployments in istio-system namespace
Service:*:istio-pilot - compare Services called "istio-pilot" in all namespaces.

cmd/mesh/manifest-diff.go
```diff
@@ -33,11 +33,23 @@ const YAMLSuffix = ".yaml"
...
+	selection string
+	// ignore specifies the resources to ignore, it's comma-separated of resource indicators
+	// in the form of "<kind>:<namespace>:<name>,<kind>:<namespace>:<name>".
```
ostromart: ignoreResources ignores all listed items during comparison. It uses the same list format as selectResources.

cmd/mesh/manifest-diff.go
```diff
@@ -33,11 +33,23 @@ const YAMLSuffix = ".yaml"
...
+	ignore string
 }
 
 func addManifestDiffFlags(cmd *cobra.Command, diffArgs *manifestDiffArgs) {
 	cmd.PersistentFlags().BoolVarP(&diffArgs.compareDir, "directory", "r",
 		false, "compare directory")
+	cmd.PersistentFlags().StringVar(&diffArgs.selection, "select", "::",
+		"specifies the resources to compare, it's comma-separated of resource indicators"+
+			"in the form of \"<kind>:<namespace>:<name>,<kind>:<namespace>:<name>\".")
```
ostromart: Use the text above, if it looks good to you.

cmd/mesh/manifest-diff.go
```diff
@@ -33,11 +33,23 @@ const YAMLSuffix = ".yaml"
...
+	cmd.PersistentFlags().StringVar(&diffArgs.ignore, "ignore", "",
+		"specifies the resources to ignore, it's comma-separated of resource indicators"+
+			"in the form of \"<kind>:<namespace>:<name>,<kind>:<namespace>:<name>\".")
```
ostromart: same here.

cmd/mesh/manifest-diff.go
```diff
@@ -47,17 +59,20 @@ func manifestDiffCmd(rootArgs *rootArgs, diffArgs *manifestDiffArgs) *cobra.Comm
 		Long:  "The diff-manifest subcommand is used to compare manifest from two files or directories.",
 		Args:  cobra.ExactArgs(2),
 		Run: func(cmd *cobra.Command, args []string) {
+			if diffArgs.selection != "::" && diffArgs.ignore != "" {
+				logAndFatalf(rootArgs, "Cannot specify both the selected and ignored resources.")
+			}
```
ostromart: I don't think these need to be mutually exclusive as long as the order they are applied is documented. I'd propose select before ignore. 
In that case you could do --selectResources Deployment:*:* --ignoreResources *:*:istio-pilot
That would compare all Deployments except the one called istio-pilot.

pkg/object/objects.go
```diff
@@ -410,6 +411,123 @@ func ManifestDiff(a, b string) (string, error) {
...
 
+// buildResourceRegexp translates the resource indicator to regexp.
+func buildResourceRegexp(s string) *regexp.Regexp {
+	hash := strings.Split(s, ":")
+	for i, v := range hash {
+		if v == "" {
+			hash[i] = ".*"
+		}
+	}
+	return regexp.MustCompile(strings.Join(hash, ":"))
```
ostromart: This will panic if it fails so we can't use it here.

pkg/object/objects.go
```diff
@@ -410,6 +411,123 @@ func ManifestDiff(a, b string) (string, error) {
...
+				}
+			}
+		}
+	}
+	return sb.String(), err
+}
+
+// ManifestDiffWithIgnore checks the manifest differences with ignored indicator.
+func ManifestDiffWithIgnore(a, b, ignore string) (string, error) {
+	ao, err := ParseK8sObjectsFromYAMLManifest(a)
```
ostromart: You could restructure the code as follows:
1. create a top level function called ManifestDiffWithSelectAndIgnore(...)
2. get objects slices for manifests a and b.
3. apply select and ignore filters to those lists
4. create the aom, bom maps and proceed as before.

cmd/mesh/manifest-diff.go
morvencao: done.

cmd/mesh/manifest-diff.go
morvencao: done.

cmd/mesh/manifest-diff.go
morvencao: done.

cmd/mesh/manifest-diff.go
morvencao: done.

cmd/mesh/manifest-diff.go
morvencao: done.

pkg/object/objects.go
morvencao: Yes, changed to `regexp.Compile(strings.Join(hash, ":"))`

pkg/object/objects.go
morvencao: done.

pkg/object/objects_test.go
```diff
@@ -732,3 +732,417 @@ func TestUnmarshalWithJSONPB(t *testing.T) {
...
+      - name: citadel
+        image: docker.io/istio/citadel:1.1.8
+        args:
+        - "--append-dns-names=true"
+        - "--grpc-port=8060"
+        - "--grpc-hostname=citadel"
+        - "--citadel-storage-namespace=istio-system"
+        - "--custom-dns-names=istio-pilot-service-account.istio-system:istio-pilot.istio-system"
+        - "--monitoring-port=15014"
+        - "--self-signed-ca=true"
```
ostromart: having this much stuff in the resource actually makes it harder to know what the test is doing and what is expected. i recommend defining very minimal resources so it's clear what the differences are.

pkg/object/objects_test.go
```diff
@@ -732,3 +732,417 @@ func TestUnmarshalWithJSONPB(t *testing.T) {
...
+      - name: citadel
+        image: docker.io/istio/citadel:1.2.2
+        args:
+        - "--append-dns-names=true"
+        - "--grpc-port=8060"
+        - "--grpc-hostname=citadel"
+        - "--citadel-storage-namespace=istio-system"
+        - "--custom-dns-names=istio-pilot-service-account.istio-system:istio-pilot.istio-system"
+        - "--monitoring-port=15014"
+        - "--self-signed-ca=true"
```
ostromart: same here, i recommend paring these resources back to be as small as possible to still be able to get a diff for the test.

morvencao:
@ostromart OK, I'll fix the UT in the followup PR.

PR #182, opened 2019-08-12
Split up stackdriver template

ostromart:
This is a workaround for https://github.com/istio/istio/issues/16207.
Unfortunately I could not find anything obvious as the root cause so that issue must be investigated more deeply.


ostromart:
Found root cause to issue, closing this.

PR #183, opened 2019-08-12
Fix for vfsgen reader

ostromart:
This also removes the config-stdio.yaml which was another workaround for the buffer limit problem.


ostromart:
Fixes https://github.com/istio/istio/issues/16207.

PR #185, opened 2019-08-12
Sync charts and values to installer

ostromart:
This merges latest changes from istio/installer.


costinm:
I thought the agreement was to stop duplicating the charts ( on the last PR ) ?

ostromart:
Yep, this is the last one. The only diff now is the outstanding PR in istio/installer - if we can get that merged I can switch these over tomorrow.

ostromart:
This is unneeded since https://github.com/istio/operator/pull/190 supersedes it. That PR cherry picks the profile changes only out of this PR.

PR #186, opened 2019-08-13
Add support for installation from url

richardwxn:


pkg/helm/urlfetcher.go
```diff
@@ -45,53 +54,50 @@ type URLFetcher struct {
...
+		}
+	}
 	uf := &URLFetcher{
-		url:         repoURL + "/" + chartName,
-		verifyURL:   repoURL + "/" + shaName,
-		versionsURL: repoURL + "/" + VersionsFileName,
-		verify:      true,
-		destDir:     destDir,
-		downloader:  NewFileDownloader(),
+		url:        filepath.Join(repoURL, cFileName),
```
golangcibot: File is not `goimports`-ed (from `goimports`)

cmd/mesh/manifest.go
```diff
@@ -89,6 +91,19 @@ func genManifests(inFilename string, setOverlayYAML string) (name.ManifestMap, e
...
+		// TODO: Make the filename logic more robust
+		uf, err := helm.NewURLFetcher(mergedICPS.InstallPackagePath, "", helm.InstallationChartsFileName, helm.InstallationShaFileName)
+		if err != nil {
+			return nil, err
+		}
+		errs := uf.FetchBundles()
+		if len(errs) != 0 {
+			return nil, errs.ToError()
+		}
+		mergedICPS.InstallPackagePath = filepath.Join(uf.GetDestDir(), "install", "kubernetes", "operator", "charts")
```
ostromart: you don't have to separate the path elements for filepath i.e. just "install/kubernetes/operator/charts" - but that path should come from a const, which you can put in here for now.

pkg/helm/urlfetcher.go
```diff
@@ -29,10 +28,20 @@ import (
...
+	"istio.io/operator/pkg/httprequest"
 	"istio.io/operator/pkg/util"
 	"istio.io/pkg/log"
 )
 
+const (
+	// InstallationChartsFileName is the name of the installation package to fetch.
+	InstallationChartsFileName = "istio-installer-1.3.0.tar.gz"
+	// InstallationShaFileName is Sha filename to verify
+	InstallationShaFileName = "istio-installer-1.3.0.tar.gz.sha256"
```
ostromart: the 1.3.0 part should come from the URL path/version.

pkg/helm/urlfetcher.go
```diff
@@ -124,79 +131,36 @@ func (f *URLFetcher) fetchChart(shaF string) error {
 			return fmt.Errorf("checksum of charts file located at: %s does not match expected SHA file: %s", saved, shaF)
 		}
 	}
-
-	return archiver.Unarchive(saved, f.destDir)
+	targz := archiver.TarGz{Tar: &archiver.Tar{OverwriteExisting: true}}
```
ostromart: somewhere, we should check if the tar is already there (using the sha perhaps). if so, we should not download it repeatedly.

pkg/util/common.go
```diff
@@ -42,3 +43,9 @@ func (t Tree) String() string {
 func IsFilePath(path string) bool {
 	return strings.Contains(path, "/") || strings.Contains(path, ".")
 }
+
+// IsHTTPURL checks whether the given URL is a HTTP URL, empty path or relative URLs would be rejected.
+func IsHTTPURL(path string) bool {
+	u, err := url.Parse(path)
```
ostromart: I'm not sure if this is restrictive enough. I think we'd also want to enforce a http:// or https:// prefix.

ostromart:
@costinm, the downloaded files are the istio release tar and that's where the signature comes from. The istio/installer charts and operator profiles are being bundled into the istio release tar. 
There's very little code required to implement this and it's a useful shortcut, according to UX feedback.

cmd/mesh/manifest.go
richardwxn: done

pkg/util/common.go
```diff
@@ -15,10 +15,11 @@
 package util
 
 import (
+	"net/url"
 	"strings"
 
 	"github.com/ghodss/yaml"
-
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

cmd/mesh/manifest.go
```diff
@@ -89,6 +93,27 @@ func genManifests(inFilename string, setOverlayYAML string) (name.ManifestMap, e
 	if err != nil {
 		return nil, err
 	}
+	// for installation path contains version number only
+	isp := mergedICPS.InstallPackagePath
+	if version.IsVersionString(isp) {
+		mergedICPS.InstallPackagePath = fmt.Sprintf(helm.InstallationPathTemplate, isp, isp)
```
ostromart: it would be more readable to write this as mergedICPS.InstallPackagePath = helm.InstallURLFromVersion(mergedICPS.version)
then, make InstallationPathTemplate private and InstallURLFromVersion should only need version string, not whole path.

cmd/mesh/manifest.go
```diff
@@ -89,6 +93,27 @@ func genManifests(inFilename string, setOverlayYAML string) (name.ManifestMap, e
 	if err != nil {
 		return nil, err
 	}
+	// for installation path contains version number only
+	isp := mergedICPS.InstallPackagePath
```
ostromart: should check that at most one of isp.InstallPackagePath or isp.Version are set

cmd/mesh/manifest.go
```diff
@@ -89,6 +93,27 @@ func genManifests(inFilename string, setOverlayYAML string) (name.ManifestMap, e
...
+	if version.IsVersionString(isp) {
```
ostromart: you can assume version string is ok, since we pass isp through validation. this should just check if it set to anything.

cmd/mesh/manifest.go
```diff
@@ -89,6 +93,27 @@ func genManifests(inFilename string, setOverlayYAML string) (name.ManifestMap, e
...
+		mergedICPS.InstallPackagePath = fmt.Sprintf(helm.InstallationPathTemplate, isp, isp)
+	}
+	if util.IsHTTPURL(mergedICPS.InstallPackagePath) {
+		uf, err := helm.NewURLFetcher(mergedICPS.InstallPackagePath, "")
```
ostromart: define a temp dir here since we want subsequent calls to this command to use the same dir.

cmd/mesh/manifest.go
```diff
@@ -89,6 +93,27 @@ func genManifests(inFilename string, setOverlayYAML string) (name.ManifestMap, e
...
+		if err != nil {
+			return nil, err
+		}
+		errs := uf.FetchBundles()
+		if len(errs) != 0 {
```
ostromart: if err := uf.FetchBundles().ToError(); err != nil {
...
}

cmd/mesh/manifest.go
```diff
@@ -89,6 +93,27 @@ func genManifests(inFilename string, setOverlayYAML string) (name.ManifestMap, e
 	if err != nil {
 		return nil, err
 	}
+	// for installation path contains version number only
+	isp := mergedICPS.InstallPackagePath
+	if version.IsVersionString(isp) {
+		mergedICPS.InstallPackagePath = fmt.Sprintf(helm.InstallationPathTemplate, isp, isp)
+	}
+	if util.IsHTTPURL(mergedICPS.InstallPackagePath) {
```
ostromart: i suggest moving this block to a helper function. it's too low level detail for the reader at this function level.

cmd/mesh/manifest.go
```diff
@@ -89,6 +93,27 @@ func genManifests(inFilename string, setOverlayYAML string) (name.ManifestMap, e
...
+		}
+		errs := uf.FetchBundles()
+		if len(errs) != 0 {
+			return nil, errs.ToError()
+		}
+		isp := path.Base(mergedICPS.InstallPackagePath)
+		// get rid of the suffix, installation package is untared to folder name istio-{version}, e.g. istio-1.3.0
+		idx := strings.LastIndex(isp, "-")
+
+		mergedICPS.InstallPackagePath = filepath.Join(uf.GetDestDir(), isp[:idx], helm.ChartsFilePath)
```
ostromart: we should not be mutating mergedICPS - i can see why you did it, but maybe leave a TODO to fix that.

pkg/helm/urlfetcher.go
```diff
@@ -15,83 +15,89 @@
...
 )
 
-// FileDownloader is wrapper of HTTP client to download files
-type FileDownloader struct {
-	// client is a HTTP/HTTPS client.
-	client *http.Client
-}
+const (
+	// InstallationPathTemplate is used to construct installation url based on version
+	InstallationPathTemplate = "https://github.com/istio/istio/releases/download/%s/istio-%s-linux.tar.gz"
```
ostromart: make this private and create a helper function per comment above

pkg/helm/urlfetcher.go
```diff
@@ -15,83 +15,89 @@
...
 	verify bool
 	// destDir is path of charts downloaded to, empty as default to temp dir
 	destDir string
-	// downloader downloads files from remote url
-	downloader *FileDownloader
 }
 
-// NewURLFetcher creates an URLFetcher pointing to urls and destination
-func NewURLFetcher(repoURL string, destDir string, chartName string, shaName string) *URLFetcher {
+// NewURLFetcher creates an URLFetcher pointing to installation package url and destination
```
ostromart: s/url/URL, add period.
... and returns a pointer to it.

pkg/helm/urlfetcher.go
```diff
@@ -15,83 +15,89 @@
...
+		verifyURL: insPackageURL + SHAFileSuffix,
+		verify:    true,
+		destDir:   destDir,
 	}
-	return uf
+	return uf, nil
+}
+
+// GetDestDir returns path of destination dir.
+func (f *URLFetcher) GetDestDir() string {
```
ostromart: usual Go idiom is to say DestDir, not GetDestDir if possible.

cmd/mesh/manifest.go
richardwxn: The goal here is not for validation. The problem is that there might be different kinds of install package path being set, e.g. local filepath, http url, version string only, here I want to construct the full url for version string only case.

cmd/mesh/manifest.go
richardwxn: done

cmd/mesh/manifest.go
richardwxn: done

cmd/mesh/manifest.go
richardwxn: discussed offline, update this part

cmd/mesh/manifest.go
richardwxn: done

cmd/mesh/manifest.go
richardwxn: done

pkg/helm/urlfetcher.go
richardwxn: done

pkg/helm/urlfetcher.go
richardwxn: done

pkg/helm/urlfetcher.go
richardwxn: done

cmd/mesh/manifest.go
richardwxn: shouldn't this be done at validation package?

cmd/mesh/manifest.go
richardwxn: it is actually using the same directory instead of creating new ones with random prefix. You can check the logic at NewURLFetcher. To avoid confusion, I will update the naming and comment.

costinm:
Any link to UX feedback ? What kind of testing do we expect/need to do ?

My concern is that instead of focusing on the core
'operations/upgrade/canary', a lot of the code and testing has to be
spent on non-critical features - longer term, after operator starts
supporting it's specific use cases it may be ok
to add more stuff, but right now downloading releases should be P4 at best.

And there are implications - it's not 'cost free' even if code to download
a file is easy.  For example kustomize and is
taking the approach of using git repos - why would we use .tar.gz and .sha
file ? I was not planning to have
.tar.gz and sha for the installer - but a rolling release expecting tools
to use git to merge.

Keep in mind most prod users are patching or adjusting our templates - a
.tar.gz doesn't work well with this
pattern. And for beginners we already have an install solution that is
super easy, with kubectl -k

cmd/mesh/manifest.go
ostromart: yes, that's a good idea. let's move it to validation later.

ostromart:
@costinm rather than holding up this PR, let's discuss it in UX group. I don't have an objection to a different approach to publishing. I do think that many users will benefit from having a single line install that points to a URL and given that the manifest commands are now in istioctl it makes a lot of sense to me to have the charts in there as well. But this is still alpha so we can get more feedback, evaluate and make a decision.

costinm:
I am not saying having a 'download' feature is bad.

I am holding releasing this feature as implemented because:
- downloading .tar.gz + .sha is ancient - kubectl is using git ( so is
golang ).
- the .sha has no purpose anyways - it provides no security. If you
download a file and the sha from the same place you don't verify any
provenance.
It was used for mirroring - but the sha would be verified against the main
site. But we don't plan to support mirroring the .tar.
- it has impact on release process - more complications to publish the
installer as .tar.gz, securing the distribution channel, etc
- what is the target audience for this feature ? Beginners who just want a
demo can use plain kubectl - which does download for you from git.
Advanced users will want to have the charts checked into their own CI/CD.
And downloading the istio release including both
istioctl + the charts + examples + other tools is not the hardest part.

We also publish a .deb file that includes istioctl and can also include the
templates - and a .rpm is supported ( not part of release, but it's
getting fixed). We can also publish a mac package if someone wants. And
that is the proper way to distribute binary packages with
extra data.

What next, turn istio into a package manager with "apt-get update" support
?

ostromart:
@costinm How is doing a download through a specified URL different from a user doing exactly the same thing, with an equally useless sha? Why prevent a useful shortcut from going in?
The format discussion is orthogonal. When we have a format for the charts that's more modern it will be easy and backwards compatible to add it as a target for installPackagePath.

pkg/helm/urlfetcher.go
```diff
@@ -15,83 +15,90 @@
...
 }
 
-// NewURLFetcher creates an URLFetcher pointing to urls and destination
-func NewURLFetcher(repoURL string, destDir string, chartName string, shaName string) *URLFetcher {
+// NewURLFetcher creates an URLFetcher pointing to installation package URL and destination,
+// and returns a pointer to it.
+func NewURLFetcher(insPackageURL string, destDir string) (*URLFetcher, error) {
 	if destDir == "" {
-		destDir = filepath.Join(os.TempDir(), ChartsTempFilePrefix)
+		destDir = filepath.Join(os.TempDir(), InstallationDirectory)
```
sdake: this looks insecure...  Please follow up with a secure mkdir implementation.

costinm:
Re SHA: The purpose of the SHA is to allow mirroring - in the old days
binary releases were mirrored by volunteers, over FTP or HTTP, and
the only safe way to download the binary was to check the SHA of the
download against the published SHA in the official web site.
You would go to apache.org, copy the SHA, select a mirror in your region
operated by a 3rd party, get the file - and verify that the
downloaded file was not altered.

We do not do this. Our release is available over https ( which has its own
security ), not from mirrors. It is absurd to check the SHA
of a file if it is downloaded from the same location with the binary - any
hacker can modify both the binary and sha at the same time.

Regarding 'useful shortcut going in' - it is called feature bloat and
complexity. All users have multiple tools to download - from browser
to curl/wget. Re-implementing curl is not providing value to users - quite
the opposite. It's another thing they need to learn and be
confused about. Not to mention concerns over security - few people want
their prod servers to download stuff from internet that runs
with cluster priviledges.

And besides: user already has to download a .tar.gz (or deb, rpm, mac pkg)
to get istioctl. What is the point to download a .tar.gz to
download another one - when the original packet includes the templates
already.

kubectl and go do download, as well as helm - but they do it because they
implements a dependency management system.
I certainly hope we are not going to implement a 'package management
system' and start downloading other operators - instead
of implementing the logic to upgrade and operate istio. And that is the
reason I'm concerned about this direction and focus.

PR #188, opened 2019-08-13
update version for release.

morvencao:


release/create_release_charts.sh
```diff
@@ -90,6 +91,18 @@ function copy_versions_files() {
     cp "${OPERATOR_BASE_DIR}/version/versions.yaml" "${OUTPUT_DIR}"
 }
 
+function update_version() {
+    local DOCKER_HUB="docker.io/istio"
+    local DOCKER_TAG=${OPERATOR_VERSION}
```
ostromart: I'm concerned putting this into the script will make it hard to understand what's going on when it's called. 
Also, I'm a bit confused about where docker.io and gcr.io should be used.

ostromart:
I've created https://github.com/istio/istio/issues/16482 - this PR could be a base, just need to find other instances of branch text and resolve the question of image hub.

morvencao:
I was thinking `create_release_charts.sh` script should just need to update the versions in the release package.
For all versions embedded in code/config when a branch is created, we should create another script for that, isn't it? Correct me if I'm not quite clear about the scenario.

ostromart:
Hi @morvencao, I'm trying to figure out what to do with this PR. I believe for creating the release charts we would not need any label fixups because we are using charts from istio/installer@release-x, so it would be their responsibility to ensure the charts are tagged appropriately, and have the correct defaults. We also should be auto-generating our default profile from their settings using manifest migrate. 
I believe that the main gap is in identifying all the places that need changes when a branch is created. If you are interested in doing that, I think it's tracked in https://github.com/istio/istio/issues/16482 so feel free to assign to yourself if you want to take it.

morvencao:
@ostromart I'm OK to close this as the work in this PR has been done by upstream.

PR #190, opened 2019-08-13
Remove charts

ostromart:
This completes the work to remove the local charts from operator. We are now using a installer.sha to specify the target in istio/installer. 
Right now we are pulling installer charts into data/charts for building. This is necessary because the embedded, self-compiling generator cannot reference external dirs easily. 
Once https://github.com/istio/operator/pull/180 goes in, we will be able to remove this limitation.
https://github.com/istio/istio/issues/16252 tracks this.


richardwxn:
Please update the README as well.

Can we default to update from the latest master if no sha provided

data/profiles/default.yaml
```diff
@@ -80,7 +79,7 @@ spec:
             resources:
               requests:
                 cpu: 100m
-                memory: 128Mi
+                memory: 1024Mi
```
morvencao: Is this a bug in upstream? Why requested memory size is bigger than limited memory size?
Should we set it as default resources settings for pilot https://github.com/istio/installer/blob/master/istio-control/istio-discovery/values.yaml#L17-L20

ostromart: This is one of several outstanding diffs that will still need to be addressed. I propose we merge this as is to progress, and start iterating on eliminating the remaining changes incrementally - wdyt?

ostromart:
@richardwxn what do you mean regarding the README?

data/profiles/default.yaml
morvencao: OK for me.

ostromart:
@richardwxn, you can already say "master" or branch name instead of sha - I updated the usage.

richardwxn:
> @richardwxn what do you mean regarding the README?

oh I mean there are some places in README still referencing the usage of compiled in chart. But we can fix that later

PR #193, opened 2019-08-14
fix mandiff gaps

morvencao:
fix mandiff gaps:
- update default resource settings for proxy and galley to align with upstream(istio/installer).
- move `env` settings for gateways from k8s overlay to legacy helm values, because k8s overlay will override other env, we need to append `env` here.
- update manifest namespace to `istio-system` in other profiles.
- add ignore list for manifest comparison.

Also there is related PR in upstream(istio/installer): https://github.com/istio/installer/pull/329


data/profiles/default.yaml
```diff
@@ -292,6 +285,11 @@ spec:
               suffix: global
             # Domain name for telemetry addons
             telemetry_domain_name: ""
+            env:
+              # A gateway with this mode ensures that pilot generates an additional
```
ostromart: Let's keep comments out of these configs - it makes for extremely huge files and we shouldn't be mixing schema and data anyway (as values.yaml does).

PR #198, opened 2019-08-14
Fix translate and profile issues

richardwxn:
replicaCount should be under k8s setting, and it is better to specify each component enablement explicitly.

Also fix translation of ControlPlaneMtls and DataPlaneMtlsStrict

### DIFF
--- a/data/profiles/default.yaml
+++ b/data/profiles/default.yaml
@@ -11,6 +11,7 @@ spec:
     components:
       pilot:
         common:
+          enabled: true
           k8s:
             env:
               - name: POD_NAME
@@ -104,6 +105,7 @@ spec:
     components:
       policy:
         common:
+          enabled: true
           k8s:
             replicaCount: 1
             hpaSpec:
@@ -135,6 +137,7 @@ spec:
     components:
       telemetry:
         common:
+          enabled: true
           k8s:
             env:
             - name: GODEBUG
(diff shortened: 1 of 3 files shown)


data/profiles/default.yaml
```diff
@@ -11,6 +11,7 @@ spec:
     components:
       pilot:
         common:
+          enabled: true
```
ostromart: all the component level enabled settings are redundant, since the feature is enabled. here and below.

pkg/translate/translate.go
```diff
@@ -119,9 +119,9 @@ var (
 
 				"Telemetry.Components.Namespace": {"global.telemetryNamespace", nil},
 
-				"Security.ControlPlaneMtls.Value":    {"global.controlPlaneSecurityEnabled", nil},
-				"Security.DataPlaneMtlsStrict.Value": {"global.mtls.enabled", nil},
-				"Security.Components.Namespace":      {"global.securityNamespace", nil},
+				"Security.ControlPlaneMtls":     {"global.controlPlaneSecurityEnabled", nil},
```
ostromart: This may work for reverse translation but I think it breaks the forward translation - since the unmarshaled value is a BoolValue rather than bool.

richardwxn: yep, looks like I have to add an override for reverse case

data/profiles/default.yaml
richardwxn: I think adding per component enablement value makes it more clear to understand. Also when doing reverse translation from values.yaml to iscp, each component enablement value is added, so it makes easier and less ambiguous during comparison

PR #203, opened 2019-08-15
Merge two install sections into one

sdake:


README.md
```diff
@@ -65,6 +65,15 @@ The quick start describes how to install and use the operator `mesh` CLI command
 
 ### Installation
```
geeknoid: Why is this called installation, if its about building the repo?

sdake: good point, addressed.

PR #205, opened 2019-08-16
Fix reverse translation issue

richardwxn:
1. Minor bug fix
2. Actually we don't need to unmarshall input values.yaml to values.types struct first and then marshall, which would lose some settings that don't have translation mappings. We can just take the input values.yaml value and do the reverse translation so that all untranslated path can be populated to common.values of different components.

### DIFF
--- a/cmd/mesh/manifest-migrate.go
+++ b/cmd/mesh/manifest-migrate.go
@@ -17,12 +17,10 @@ package mesh
 import (
 	"encoding/json"
 	"fmt"
-	"path/filepath"
-	"strings"
-
 	"github.com/ghodss/yaml"
 	"github.com/gogo/protobuf/jsonpb"
 	"github.com/spf13/cobra"
+	"path/filepath"
 
 	"istio.io/operator/pkg/kubectlcmd"
 	"istio.io/operator/pkg/translate"
@@ -59,18 +57,21 @@ func manifestMigrateCmd(rootArgs *rootArgs, mmArgs *manifestMigrateArgs) *cobra.
 }
 
 func valueFileFilter(path string) bool {
-	return filepath.Ext(path) == YAMLSuffix && strings.HasPrefix(filepath.Base(path), "values")
+	return filepath.Base(path) == "values.yaml"
 }
 
 // migrateFromFiles handles migration for local values.yaml files
 func migrateFromFiles(rootArgs *rootArgs, args []string) {
 	checkLogsOrExit(rootArgs)
-
-	logAndPrintf(rootArgs, "translating input values.yaml file at: %s to new API", args[0])
 	value, err := util.ReadFiles(args[0], valueFileFilter)
 	if err != nil {
 		logAndFatalf(rootArgs, err.Error())
 	}
+	if value == "" {
+		logAndPrintf(rootArgs, "no valid value.yaml file specified")
+		return
+	}
+	logAndPrintf(rootArgs, "translating input values.yaml file at: %s to new API", args[0])
 	translateFunc(rootArgs, []byte(value))
 }
 
--- a/cmd/mesh/manifest-migrate.go
+++ b/cmd/mesh/manifest-migrate.go
@@ -18,12 +18,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/ghodss/yaml"
 	"github.com/gogo/protobuf/jsonpb"
 	"github.com/spf13/cobra"
 
-	"istio.io/operator/pkg/apis/istio/v1alpha2"
 	"istio.io/operator/pkg/kubectlcmd"
 	"istio.io/operator/pkg/translate"
 	"istio.io/operator/pkg/util"
@@ -59,7 +59,7 @@ func manifestMigrateCmd(rootArgs *rootArgs, mmArgs *manifestMigrateArgs) *cobra.
 }
 
 func valueFileFilter(path string) bool {
-	return filepath.Base(path) == "values.yaml"
+	return filepath.Ext(path) == YAMLSuffix && strings.HasPrefix(filepath.Base(path), "values")
 }
 
 // migrateFromFiles handles migration for local values.yaml files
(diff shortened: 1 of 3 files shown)


cmd/mesh/manifest-migrate.go
```diff
@@ -59,7 +59,7 @@ func manifestMigrateCmd(rootArgs *rootArgs, mmArgs *manifestMigrateArgs) *cobra.
 }
 
 func valueFileFilter(path string) bool {
-	return filepath.Base(path) == "values.yaml"
+	return filepath.Ext(path) == YAMLSuffix && strings.HasPrefix(filepath.Base(path), "values")
```
ostromart: i'm concerned this is not going to be robust. we're relying on a filename being of the form values*.yaml - is there a better way to check for this, like file position wrt. charts root?

pkg/translate/translate_value.go
```diff
@@ -267,8 +263,11 @@ func translateHPASpec(inPath string, outPath string, value interface{}, valueTre
 	if !asEnabled {
 		return nil
 	}
-	newP := util.PathFromString(inPath)[0]
-	minPath := newP + ".autoscaleMin"
+
+	newP := util.PathFromString(inPath)
+	// last path element is k8s setting name
+	newPS := newP[:len(newP)-1].String()
+	minPath := newPS + ".autoscaleMin"
```
ostromart: the logic for autoscaleMin/Max should be the same, so i feel like some code below could be de-duped. that may be the case for other params too...

pkg/translate/translate_value.go
```diff
@@ -289,6 +288,46 @@ func translateHPASpec(inPath string, outPath string, value interface{}, valueTre
 	if _, err := tpath.DeleteFromTree(valueTree, util.ToYAMLPath(maxPath), util.ToYAMLPath(maxPath)); err != nil {
 		return err
 	}
+
+	cuPath := newPS + ".cpu.targetAverageUtilization"
+	cuVal, found, err := name.GetFromTreePath(valueTree, util.ToYAMLPath(cuPath))
```
ostromart: the operation here should probably be in a helper function - it's too detailed to follow at this level.

pkg/translate/translate_value.go
```diff
@@ -289,6 +288,46 @@ func translateHPASpec(inPath string, outPath string, value interface{}, valueTre
...
+apiVersion: apps/v1
+kind: Deployment
+name: istio-%s`
+
+	stString := fmt.Sprintf(stVal, newPS)
+	if err = yaml.Unmarshal([]byte(stString), &st); err != nil {
+		return err
+	}
+	log.Infof("path has value in helm Value.yaml tree, mapping to output path %s", outPath)
+	if err := tpath.WriteNode(cpSpecTree, util.ToYAMLPath(outPath+".scaleTargetRef"), st); err != nil {
```
ostromart: overall, i think the code above has repeated code that could be refactored at least to use helper function(s), if not to use a for loop and passing a list of mappings to it.

richardwxn: I would try to refactor and consolidate the repeating logic. But even that we can't really just do a simple mapping as there is some minor difference to handle different settings, only minReplica and maxReplicas are handled the exactly same way

pkg/translate/translate_value.go
richardwxn: I think the translateHPASpec function is already helper function to handle special translation, and since each hpa setting has slightly different handling so is not easy to extract general ways from it. So I try to refactor the logic inside this function.

cmd/mesh/manifest-migrate.go
richardwxn: I checked the helm package implementation, it just check whether the file name equals "values.yaml", like what was implemented here before. That would work for installer files, but I am thinking the requirement should not be too strict because we have seen "values*.yaml" naming.

I am not sure whether it is necessary to validate the file schema or content here, instead I think it might be more clear to just give user message what files are being translated or not.

pkg/translate/translate_value.go
```diff
@@ -267,26 +263,58 @@ func translateHPASpec(inPath string, outPath string, value interface{}, valueTre
...
 	}
-	newP := util.PathFromString(inPath)[0]
-	minPath := newP + ".autoscaleMin"
-	asMin, found, err := name.GetFromTreePath(valueTree, util.ToYAMLPath(minPath))
-	if found && err == nil {
-		if err := tpath.WriteNode(cpSpecTree, util.ToYAMLPath(outPath+".minReplicas"), asMin); err != nil {
+
+	newP := util.PathFromString(inPath)
+	// last path element is k8s setting name
+	newPS := newP[:len(newP)-1].String()
```
ostromart: could write above two lines as
newPS := util.PathFromString(inPath)[len(newP)-1]

pkg/translate/translate_value.go
```diff
@@ -267,26 +263,58 @@ func translateHPASpec(inPath string, outPath string, value interface{}, valueTre
...
+		".autoscaleMax":                 ".maxReplicas",
+		".cpu.targetAverageUtilization": ".metrics",
+	}
+
+	for key, newVal := range valMap {
+		valPath := newPS + key
+		asVal, found, err := name.GetFromTreePath(valueTree, util.ToYAMLPath(valPath))
+		if found && err == nil {
+			log.Infof("path has value in helm Value.yaml tree, mapping to output path %s", outPath)
+			if key == ".cpu.targetAverageUtilization" {
```
ostromart: rather than checking for a special key, it's better to move this outside the loop, then have both this and the loop code call the same helper function if possible.

pkg/translate/translate_value.go
```diff
@@ -416,8 +444,8 @@ func (t *ReverseTranslator) translateRemainingPaths(valueTree map[string]interfa
 				}
 				errs = util.AppendErr(errs, t.translateRemainingPaths(newMap, cpSpecTree, newPath))
 			}
-			if errs != nil {
-				return errs
+			if errs.ToError() != nil {
```
ostromart: if err := errs.ToError(); err != nil {
  return err
}

pkg/translate/translate_value.go
richardwxn: the newP is defined from util.PathFromString(inPath)
so for one line, we have to write util.PathFromString(inPath)[:len(util.PathFromString(inPath)-1].String(), which is too lengthy

pkg/translate/translate_value.go
richardwxn: removed the special key out of the loop

cmd/mesh/manifest-migrate.go
```diff
@@ -17,13 +17,11 @@ package mesh
 import (
 	"encoding/json"
 	"fmt"
-	"path/filepath"
-
 	"github.com/ghodss/yaml"
 	"github.com/gogo/protobuf/jsonpb"
 	"github.com/spf13/cobra"
+	"path/filepath"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

PR #206, opened 2019-08-16
support resource type for k8s mapping.

morvencao:
Resolve: https://github.com/istio/istio/issues/16290

### DIFF
--- a/pkg/translate/translate.go
+++ b/pkg/translate/translate.go
@@ -34,6 +34,10 @@ import (
 )
 
 const (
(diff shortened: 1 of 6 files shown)


pkg/translate/translate.go
```diff
@@ -234,60 +236,70 @@ var (
 					AlwaysEnabled:        true,
 				},
 				name.PilotComponentName: {
+					ResourceType:         "Deployment",
```
ostromart: change to string constant (would be better to reference k8s def but it seems like that's not available)

morvencao: done

PR #208, opened 2019-08-16
Fix flag not taking effect in manifest generate.

elfinhe:
Fix issue https://github.com/istio/istio/issues/16305:
--set option doesn't always override the values in the specified input: default.yaml is a copy of the built-in default profile spec.

$ ~go/bin/mesh manifest generate -f default.yaml -s security.components.citadel.common.values.selfSigned=false | grep self-signed-ca
- --self-signed-ca=true

Root cause:
1. TranslateHelmValues() is not translating citadel configs into root ["security"](https://istio.io/docs/reference/config/installation-options/#security-options) (incorrectly set to "citadel"), so it will not overwrite values from profile/set-value to the security chart.
2. The args passed from CLI is always typed to string rather than bool/int/float. When we marshal it into yaml or ICPS, the type is still mistakenly set to string. + The Helm Render engine will ignore it if the type does not match the corresponding path in chart.

### DIFF
--- a/cmd/mesh/manifest.go
+++ b/cmd/mesh/manifest.go
@@ -151,7 +151,7 @@ func makeTreeFromSetList(setOverlay []string) (string, error) {
 			return "", fmt.Errorf("bad argument %s: expect format key=value", kv)
 		}
 		k := kvv[0]
-		v := kvv[1]
+		v := util.ParseValue(kvv[1])
 		if err := tpath.WriteNode(tree, util.PathFromString(k), v); err != nil {
 			return "", err
 		}
--- a/pkg/translate/translate.go
+++ b/pkg/translate/translate.go
@@ -267,7 +267,7 @@ var (
 					ResourceName:         "istio-citadel",
 					ContainerName:        "citadel",
 					HelmSubdir:           "security/citadel",
-					ToHelmValuesTreeRoot: "citadel",
+					ToHelmValuesTreeRoot: "security",
 				},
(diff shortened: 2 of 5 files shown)


pkg/util/path_test.go
```diff
@@ -75,3 +75,79 @@ func stringSlicesEqual(a, b []string) bool {
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			if got, want := ParseValue(tt.in), tt.want; !(got == want) {
+				t.Errorf("%s: got:%v, want:%v", tt.desc, got, want)
+			}
+		})
+	}
+}
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
}
```

pkg/util/path.go
```diff
@@ -130,6 +131,21 @@ func RemoveBrackets(pe string) (string, bool) {
 	return pe[1 : len(pe)-1], true
 }
 
+// ParseValue parses string into a value
```
morvencao: nit: this function more suitable to be in https://github.com/istio/operator/blob/master/pkg/util/util.go

elfinhe: Sure, will move it.

elfinhe: Updated.

elfinhe:
> /cherrypick release-1.3

cherrypicked into release-1.3 already: https://github.com/istio/operator/pull/213

PR #209, opened 2019-08-17
Update README

richardwxn:
move the CLI part to an independent README.
Add description for manifest sub-commands.


ostromart:
I think it may be better to leave this as is, since it's intended as a walkthrough rather than a reference. Let's discuss on Monday.

richardwxn:
> I think it may be better to leave this as is, since it's intended as a walkthrough rather than a reference. Let's discuss on Monday.

Ok that makes sense, so where is the reference supposed to be sitting at, istio.io? I think we can at least add the remaining commands part to the quick tour if not creating a new independent section. Sure, let's discuss on Monday.

sdake:
@richardwxn the reference should be here:

https://preliminary.istio.io/docs/reference/commands/istioctl/

sync with @adammil2000 as I think he is working in this area.

Cheers
-steve

richardwxn:
talked offline with @ostromart, instead of creating an independent reference in the README, I made a change to the PR to only add missing commands for the quick tour part.

PR #210, opened 2019-08-17
Changes to API for mesh config integration

ostromart:
This reflects the first change to align better with mesh config:
 - removed all config items from API
 - removed redundant "common" path
 - removed per-component values overrides. All values overrides now go through single global path.
 - updated UT and profiles accordingly


README.md
```diff
@@ -292,19 +291,16 @@ spec:
...
 spec:
-  trafficManagement:
-    components:
-      pilot:
-        common:
-          values:
-            traceSampling: 0.1 # override from 1.0
+  values:
+    pilot:
+      traceSampling: 0.1 # override from 1.0
```
morvencao: Not sure if this example is suitable here, because i have tried this case, looks like it doesn't work:
```
# out/bin/mesh manifest generate -f samples/pilot-k8s.yaml | grep -i sampling
        - name: PILOT_TRACE_SAMPLING
    traceSampling: 1
```

As you can see `traceSampling` isn't overrided, because we have following k8s env overlay in `default` profile:
```
          k8s:
            env:
              - name: POD_NAME
                valueFrom:
                  fieldRef:
                    apiVersion: v1
                    fieldPath: metadata.name
              - name: POD_NAMESPACE
                valueFrom:
                  fieldRef:
                    apiVersion: v1
                    fieldPath: metadata.namespace
              - name: GODEBUG
                value: gctrace=1
              - name: PILOT_TRACE_SAMPLING
                value: "1"
              - name: CONFIG_NAMESPACE
                value: istio-config
```

The current workaround is override `traceSampling` is update it in k8s overlay:
```
kind: IstioControlPlane
spec:
  trafficManagement:
    components:
      pilot:
        common:
          k8s:
            env:
              - name: POD_NAME
                valueFrom:
                  fieldRef:
                    apiVersion: v1
                    fieldPath: metadata.name
              - name: POD_NAMESPACE
                valueFrom:
                  fieldRef:
                    apiVersion: v1
                    fieldPath: metadata.namespace
              - name: GODEBUG
                value: gctrace=1
              - name: PILOT_TRACE_SAMPLING
                value: "0.1"
              - name: CONFIG_NAMESPACE
                value: istio-config
```

data/profiles/default.yaml
```diff
@@ -10,458 +10,219 @@ spec:
...
+              cpu: 100m
+              memory: 128Mi
+            limits:
+              cpu: 2000m
+              memory: 1024Mi
+
       egressGateway:
-        common:
-          enabled: false
-          k8s:
```
morvencao: Do we want to remove the k8s overlay part for `egressGateway`? 
Although it is disabled in `default` profile, but other profiles(`demo` and `demo-auth`) and customized API will need this.

ostromart: Good point, done.

README.md
ostromart: Done.

pkg/translate/translate_value_test.go
```diff
@@ -27,6 +27,7 @@ import (
 )
 
 func TestValueToProto(t *testing.T) {
+	t.Skip("Disabled while components.common.values is removed.")
```
richardwxn: Add a TODO here?

cmd/mesh/manifest.go
```diff
@@ -111,13 +111,6 @@ func genManifests(inFilename string, setOverlayYAML string) (name.ManifestMap, e
 
 // fetchInstallPackageFromURL downloads installation packages from specified URL.
 func fetchInstallPackageFromURL(mergedICPS *v1alpha2.IstioControlPlaneSpec) error {
```
richardwxn: so we are removing this functionality for specifying version and install?

ostromart: for 1.3 only. we will discuss this again in 1.4 timeframe.

pkg/translate/translate_value_test.go
ostromart: I've created https://github.com/istio/istio/issues/16390.

PR #212, opened 2019-08-19
update doc.

morvencao:


ARCHITECTURE.md
```diff
@@ -60,16 +60,22 @@ Base | CRDs
 Traffic Management | Pilot
 Policy | Policy
 Telemetry | Telemetry
+Telemetry | Prometheus
+Telemetry | Prometheus Operator
+Telemetry | Grafana
+Telemetry | Kiali
+Telemetry | Tracing
```
ostromart: These need an asterisk as they are not core components (e.g. they are not in the proto). Maybe take them out of the table and include a note under it that mentions them?

ARCHITECTURE.md
```diff
@@ -60,16 +60,22 @@ Base | CRDs
...
 Security | Cert manager
 Configuration management | Galley
 Gateways | Ingress gateway
 Gateways | Egress gateway
 AutoInjection | Sidecar injector
+ThirdParty | CNI
 
 Features and components are defined in the
-[name](https://github.com/istio/operator/blob/905dd84e868a0b88c08d95b7ccf14d085d9a6f6b/pkg/name/name.go#L38) package.
+[name](https://github.com/istio/operator/blob/master/pkg/name/name.go#L246) package.
```
ostromart: Probably not a good idea to use master since this will be changing. You could update the sha to something more recent though.

ARCHITECTURE.md
morvencao: done.

ARCHITECTURE.md
morvencao: done.

PR #215, opened 2019-08-20
update reverse logic after api changes

richardwxn:


pkg/translate/translate_value.go
```diff
@@ -459,51 +459,19 @@ func (t *ReverseTranslator) translateRemainingPaths(valueTree map[string]interfa
...
 			}
-		// remaining leaf need to be put into common.values
+		// remaining leaf need to be put into root.values
 		default:
-			err := t.translateToCommonValues(cpSpecTree, newPath, val)
-			if err != nil {
+			if newPath[len(newPath)-1] == "enabled" {
+				continue
+			}
+			if err := tpath.WriteNode(cpSpecTree, util.ToYAMLPath("Values." + newPath.String()), val); err != nil {
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
			if err := tpath.WriteNode(cpSpecTree, util.ToYAMLPath("Values."+newPath.String()), val); err != nil {
```

PR #216, opened 2019-08-20
Use release-1.3 tag for release-1.3 branch

rlenglet:


cmd/mesh/manifest-versions.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	versionsMapURL = "https://raw.githubusercontent.com/istio/operator/master/version/versions.yaml"
+	versionsMapURL = "https://raw.githubusercontent.com/istio/operator/release-1.3/version/versions.yaml"
```
rlenglet: I'm not sure whether this change is required or correct.

At least, I'm sure updating the tags `master-latest-daily` -> `release-1.3-latest-daily` is required.

rlenglet:
Tests are passing. Please someone take a look.

cmd/mesh/manifest-versions.go
ostromart: Thanks for doing this. In this case it is not required because master always contains the version of the map we want users to have. That's because they may want to know about release versions more recent than their binary.

rlenglet: Ok, I'll revert that line.

rlenglet:
@ostromart Please review again.

rlenglet:
And I'll need an approval from @istio/release-managers-1-3. Thanks!

ostromart (approved):

data/profiles/default.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: install.istio.io/v1alpha2
 kind: IstioControlPlane
 spec:
   hub: gcr.io/istio-release
-  tag: master-latest-daily
+  tag: release-1.3-latest-daily
```
sdake: need to ensure this tag is available at some point or the operator will be doa in the firs trc.

sdake (approved):
assuming you grepped for master and replaced?

data/profiles/default.yaml
rlenglet: Good point.
@howardjohn do you know what needs to be done to get those daily builds created?

howardjohn:
it's already done we have had daily builds since the day we cut the release
branch

howardjohn:
https://gcsweb.istio.io/gcs/istio-prerelease/daily-build/release-1.3-latest-daily/

`docker run gcr.io/istio-release/pilot:release-1.3-latest-daily` works for
me

rlenglet:
Right. I was looking for an operator image...

howardjohn:
I don't think there is an operator image? It is just a part of istioctl

rlenglet:
Got it. I was confused. Thanks!

ostromart:
Right, there's no operator image atm, it's only in the CLI.

PR #218, opened 2019-08-20
Create e2e tests for manifest generate command

ostromart:
This PR cleans up the component/controlplane unit tests, which are better served through e2e tests. Included here:
- make changes to how logging/output are handled so output can be captured for tests
- use the istio/installer charts as inputs, to detect breakage when installer.sha is updated
- use data/profiles as inputs to detect breakage in the profiles
- use golden outputs, with select and ignore flags to constrain the amount of output to just the resources of interest


cmd/mesh/manifest-migrate.go
```diff
@@ -62,68 +63,68 @@ func valueFileFilter(path string) bool {
...
 	}
-	logAndPrintf(rootArgs, "translating input values.yaml file at: %s to new API", args[0])
-	translateFunc(rootArgs, []byte(value))
+	l.logAndPrint("translating input values.yaml file at: ", args[0], " to new API")
+	translateFunc(rootArgs, []byte(value), l)
 }
 
 // translateFunc translates the input values and output the result
-func translateFunc(rootArgs *rootArgs, values []byte) {
+func translateFunc(rootArgs *rootArgs, values []byte, l *logger) {
```
golangcibot: `translateFunc` - `rootArgs` is unused (from `unparam`)

cmd/mesh/shared.go
```diff
@@ -40,31 +58,42 @@ func getWriter(outFilename string) (*os.File, error) {
...
 
-func logAndFatalf(args *rootArgs, v ...interface{}) {
-	logAndPrintf(args, v...)
+func (l *logger) logAndFatal(v ...interface{}) {
+	l.logAndPrint(v...)
 	os.Exit(-1)
 }
+
+func (l *logger) print(s string) {
+	l.stdOut.Write([]byte(s))
```
golangcibot: Error return value of `l.stdOut.Write` is not checked (from `errcheck`)

tests/e2e/generate/generate_test.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func syncCharts() error {
+	cmd := exec.Command(filepath.Join(repoRootDir, "run_update_charts.sh"))
+	return cmd.Run()
+}
+
+func runManifestGenerate(path string) (string, error) {
+
+}
```
golangcibot: missing return (from `typecheck`)

cmd/mesh/shared.go
```diff
@@ -40,31 +58,42 @@ func getWriter(outFilename string) (*os.File, error) {
...
 }
 
-func logAndFatalf(args *rootArgs, v ...interface{}) {
-	logAndPrintf(args, v...)
+func (l *logger) logAndFatal(v ...interface{}) {
+	l.logAndPrint(v...)
 	os.Exit(-1)
 }
+
+func (l *logger) print(s string) {
```
golangcibot: U1000: func `(*logger).print` is unused (from `unused`)

cmd/mesh/shared.go
```diff
@@ -40,31 +58,42 @@ func getWriter(outFilename string) (*os.File, error) {
...
+func newLogger(logToStdErr bool, stdOut, stdErr io.Writer) *logger {
+	return &logger{
+		logToStdErr: logToStdErr,
+		stdOut:      stdOut,
+		stdErr:      stdErr,
 	}
-	return log.Configure(opt)
 }
 
 // TODO: this really doesn't belong here. Figure out if it's generally needed and possibly move to istio.io/pkg/log.
```
richardwxn: is the TODO part still a concern?

cmd/mesh/shared.go
```diff
@@ -18,15 +18,33 @@ package mesh
...
 	"istio.io/pkg/log"
 )
 
 const (
-	logFilePath = "./mesh-cli.log"
+	logFilePath = "./.mesh-cli.log"
 )
 
+func configLogsOrExit(args *rootArgs) {
+	_ = os.Remove(logFilePath)
```
richardwxn: would this file be created back somewhere?not sure whether log.Configure does the work

cmd/mesh/shared.go
ostromart: Yes :-/ this is one of the TODOs I want to address when we get the ball rolling on 1.4.
There are about 30 others!

cmd/mesh/shared.go
ostromart: I've left a comment with some explanation. Normally it would be weird to have "config log" function have a side effect of deleting something :-)
I think a more accurate name is initLogsOrExit and in this case it's probably reasonable to have the delete in there, since this function is the first call in just about all commands.

cmd/mesh/manifest-generate.go
```diff
@@ -49,36 +48,37 @@ func manifestGenerateCmd(rootArgs *rootArgs, mgArgs *manifestGenerateArgs) *cobr
...
 	manifests, err := genManifests(mgArgs.inFilename, overlayFromSet)
 	if err != nil {
-		logAndFatalf(args, err.Error())
+		l.logAndFatal(err.Error())
 	}
 
 	if mgArgs.outFilename == "" {
 		for _, m := range orderedManifests(manifests) {
-			fmt.Println(m)
+			l.print(m)
```
morvencao: Should we handle newline here?

ostromart: Done.

cmd/mesh/manifest-generate_test.go
```diff
@@ -0,0 +1,114 @@
+package mesh
```
morvencao: Please add the boilerplate license headers, because the linter is complaining.

PR #235, opened 2019-08-27
add an option to regenerate testdata for unit tests

ayj:
Run  `REFRESH_GOLDEN=true go test ./cmd/mesh/...` to regenerate the test data based on the current test output.

fixes https://github.com/istio/istio/issues/16493

### DIFF
--- a/cmd/mesh/manifest_shared_test.go
+++ b/cmd/mesh/manifest_shared_test.go
@@ -60,4 +60,3 @@ func readFile(path string) (string, error) {
 	b, err := ioutil.ReadFile(path)
 	return string(b), err
 }
-


cmd/mesh/manifest_shared_test.go
```diff
@@ -60,3 +60,10 @@ func readFile(path string) (string, error) {
 	b, err := ioutil.ReadFile(path)
 	return string(b), err
 }
+
+func refreshGoldenFiles() bool {
+	if os.Getenv("REFRESH_GOLDEN") == "true" {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

PR #242, opened 2019-08-30
Add basic e2e tests

richardwxn:
Run the e2e tests from istio/istio

### DIFF
--- a/prow/e2e_install.sh
+++ b/prow/e2e_install.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+# Copyright 2019 Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+set -x
+
+WD=$(dirname "$0")
+WD=$(cd "$WD"; pwd)
+ROOT=$(dirname "$WD")
+export GO111MODULE=on
+
+cd "${ROOT}"
+echo "installing istio with operator CLI"
+go run ./cmd/mesh.go manifest apply
+
+ISTIO_DIR=$(mktemp -d)/src/istio.io/istio
+ISTIO_CONTROL_NS=istio-system
+MODE=permissive
+ISTIOCTL_BIN=${ISTIOCTL_BIN:-/usr/local/bin/istioctl}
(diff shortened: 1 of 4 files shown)


prow/e2e_install.sh
```diff
@@ -0,0 +1,69 @@
...
+SIMPLE_AUTH=false
+E2E_ARGS="--skip_setup=true --use_local_cluster=true --istio_namespace=${ISTIO_CONTROL_NS}"
+TMPDIR=/tmp
+export IstioTop=${ISTIO_DIR}/../../..
+
+git clone https://github.com/istio/istio.git "${ISTIO_DIR}"
+
+function run-simple-base() {
+    kubectl delete ns ${NS} || true
+  	kubectl create ns ${NS} || true
```
ostromart: this ns should be created by the operator i believe. also, since this test will be run from prow i don't think we need to remove the ns, since anyway this will not fully clean the cluster.

howardjohn (commented):
what is setting up the cluster? You will need to use kind

richardwxn:
Added the kind setup

prow/e2e_install.sh
richardwxn: Removed the delete part, but for creation, this is the application ns not the istio ns.

ostromart (approved):

prow/e2e_install.sh
```diff
@@ -0,0 +1,69 @@
...
+export GO111MODULE=on
+
+cd "${ROOT}"
+echo "installing istio with operator CLI"
+go run ./cmd/mesh.go manifest apply
+
+ISTIO_DIR=$(mktemp -d)/src/istio.io/istio
+ISTIO_CONTROL_NS=istio-system
+MODE=permissive
+ISTIOCTL_BIN=${ISTIOCTL_BIN:-/usr/local/bin/istioctl}
```
howardjohn: istioctl is not in the prow docker image, you would need to build it

prow/e2e_install.sh
```diff
@@ -0,0 +1,69 @@
...
+WD=$(dirname "$0")
+WD=$(cd "$WD"; pwd)
+ROOT=$(dirname "$WD")
+export GO111MODULE=on
+
+cd "${ROOT}"
+echo "installing istio with operator CLI"
+go run ./cmd/mesh.go manifest apply
+
+ISTIO_DIR=$(mktemp -d)/src/istio.io/istio
```
howardjohn: I would recommend just looking in `$GOPATH/src/istio.io/istio`. Prow can check out additional repos for you so you don't need to `git clone` by passing `extra_refs`

prow/e2e_install.sh
howardjohn: nit: inconsistent spacing

howardjohn (commented):
I still don't see where kind cluster is created?

I would recommend testing locally with https://github.com/istio/istio/wiki/Working-with-Prow#testing-changes-locally

richardwxn:
oh I somehow pushed commit from my mac instead of my desktop which has latest commit for kind setup. I would create a new PR for that missing commit and also to address your comment

PR #244, opened 2019-09-01
Fix env value type

richardwxn:
ensure value type of env

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -70,7 +70,6 @@ require (
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/jonboulle/clockwork v0.1.0 // indirect
 	github.com/json-iterator/go v1.1.6 // indirect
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -70,6 +70,7 @@ require (
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/jonboulle/clockwork v0.1.0 // indirect
 	github.com/json-iterator/go v1.1.6 // indirect
+	github.com/jteeuwen/go-bindata v0.0.0-20180305030458-6025e8de665b // indirect
```
ostromart: why is this reappearing? i just removed it due to the release build crash. let's stick with vfsgen.

richardwxn: it was added when I built locally, anyway I removed it.

morvencao (approved):

PR #246, opened 2019-09-03
Add kind setup back

richardwxn:
put back the missing commit when pushing this PR: https://github.com/istio/operator/pull/242
also try to address newer comment from it.

### DIFF
--- a/prow/e2e_install.sh
+++ b/prow/e2e_install.sh
@@ -20,50 +20,60 @@ set -x
 WD=$(dirname "$0")
 WD=$(cd "$WD"; pwd)
 ROOT=$(dirname "$WD")
-export GO111MODULE=on
-
 cd "${ROOT}"
-echo "installing istio with operator CLI"
-go run ./cmd/mesh.go manifest apply
 
-ISTIO_DIR=$(mktemp -d)/src/istio.io/istio
-ISTIO_CONTROL_NS=istio-system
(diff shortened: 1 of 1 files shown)


prow/e2e_install.sh
```diff
@@ -20,50 +20,60 @@ set -x
...
-ISTIOCTL_BIN=${ISTIOCTL_BIN:-/usr/local/bin/istioctl}
 SIMPLE_AUTH=false
 E2E_ARGS="--skip_setup=true --use_local_cluster=true --istio_namespace=${ISTIO_CONTROL_NS}"
 TMPDIR=/tmp
+export GO111MODULE=on
 export IstioTop=${ISTIO_DIR}/../../..
 
-git clone https://github.com/istio/istio.git "${ISTIO_DIR}"
+#kind and istioctl setup
+KIND_IMAGE="kindest/node:v1.14.0"
```
howardjohn: I think the kind node needs to directly match with the kind version being used, so this may not work. I would just not pass anything unless you specifically need to test kubernetes 1.14 instead of 1.15. If you do, then it will need to be something else like 1.14.6 of something.. not sure. Look at the k8s-114 job from test-infra repo for reference

richardwxn: ok, removed the hardcoded node_image name so that if nothing pass in would just use what matches with kind version

ostromart (approved):

PR #247, opened 2019-09-03
strategic merge overlay from operator api to values.

### LINKED ISSUE #16551: Fields from values overrides are not merged with those from IstioControlPlane
Some values have paths in both IstioControlPlane and helm e.g. 
 spec.security.components.nodeAgent.k8s.env is the same as
 nodeagent.env

It should be possible to set the values in both places and merge the two trees, but currently IstioControlPlane settings override helm.

morvencao:
resolve: https://github.com/istio/istio/issues/16551


pkg/translate/translate.go
```diff
@@ -389,12 +392,48 @@ func (t *Translator) OverlayK8sSettings(yml string, icp *v1alpha2.IstioControlPl
 		if err != nil {
 			return "", fmt.Errorf("could not marshal to YAML in OverlayK8sSettings: %s", err)
 		}
-		mergedYAML, err := overlayK8s(baseYAML, overlayYAML, path[1:])
+
+		var dstNodeYaml, k8sSettingName string
+		if len(path) != 0 {
+			k8sSettingName = path[len(path)-1]
```
ostromart: You should be able to assume len >0 here because path is hard coded.

pkg/translate/translate.go
```diff
@@ -389,12 +392,48 @@ func (t *Translator) OverlayK8sSettings(yml string, icp *v1alpha2.IstioControlPl
...
+		}
+
+		switch k8sSettingName {
```
ostromart: Suggest moving all this code into overlayK8s, otherwise the function will be too long and complex to understand.

pkg/translate/translate.go
```diff
@@ -389,12 +392,48 @@ func (t *Translator) OverlayK8sSettings(yml string, icp *v1alpha2.IstioControlPl
...
+			if err := yaml.Unmarshal(baseYAML, &base); err != nil {
+				return "", fmt.Errorf("error: %s to unmarshal baseYAML:\n%s", err, baseYAML)
+			}
+
+			basePc, isBasePcFound, err := tpath.GetPathContext(base, path[1:])
+			if err != nil {
+				return "", err
+			}
+
+			if !isBasePcFound {
```
ostromart: I have a suggestion that may simplify the code and make it a bit more general. Instead of doing a special case for env, it should be possible to merge (rather than overwrite) each of the k8s destination subtrees inside overlayK8s.

pkg/translate/translate.go
```diff
@@ -406,6 +445,55 @@ func (t *Translator) OverlayK8sSettings(yml string, icp *v1alpha2.IstioControlPl
 	return objects.YAMLManifest()
 }
 
+// mergeEnv apply deep merge overlay env settings on the base env.
+func mergeEnv(baseEnv, overlayEnv string) (string, error) {
```
ostromart: This function should be unnecessary - see comments below.

pkg/translate/translate.go
```diff
@@ -406,6 +445,55 @@ func (t *Translator) OverlayK8sSettings(yml string, icp *v1alpha2.IstioControlPl
...
+		}
+	}
+
+	b, err := json.Marshal(dstEnvObj)
+	if err != nil {
+		return "", err
+	}
+	return string(b), nil
+}
+
```
ostromart: It will simplify the logic considerably if you think of the problem in the following way: the root issue here is that overlayK8s below is doing an overwrite using WriteNode, whereas we should be doing a merge for every node. You could either create a MergeNode wrapper similar to WriteNode that internally calls the traversal code and does a merge when the node is found, or you can read the node here, merge with the new value and WriteNode the result. 
In either case you can use overlayTrees or OverlayYAML to do the merge.

ostromart: btw. feel free to ping me tonight on slack is any of this doesn't make sense or you need some more pointers.

morvencao: @ostromart Have tried the overlayTrees and OverlayYAML, it can't meet our requirements. The reason is that we currently use the general json/yaml merge tool, eg. https://github.com/evanphx/json-patch
For `env`, we have original snippet:

```
env:
- name: n1
  value: v1
```

and override snippet:
```
env:
- name: n1
  value: v2
```

the general json/yaml merge will just reproduce:
```
env:
- name: n1
  value: v1
- name: n1
  value: v2
```

we need semantic merge, because what we really want is:
```
env:
- name: n1
  value: v2
```

morvencao: Creating a MergeNode wrapper similar to WriteNode make sense to me.

ostromart: That's a good point, we would need extra work to implement the correct merge semantics. However, I think if this code is added to a MergeNode function I think it would be a reasonable default behavior for that function and be generally useful. 
Thanks for tackling this tricky problem btw. :-)

morvencao: Actually I'm thinking to add a general method, eg, `K8sJsonMerge` that accepts base and overlay json, also somewhat about **the keys of array object**, but it's tricky, because I can't figure out how to pass the keys of array object.
Wdyt, does it make sense?

pkg/translate/translate.go
morvencao: done.

pkg/translate/translate.go
morvencao: done.

pkg/translate/translate.go
ostromart: Yes, it's tricky! I can suggest the following approach: recurse through the base tree A and for each node:
- if node doesn't exist in B, skip.
- if the node is a leaf-list, find the item that has the most keys in common and overlay B on that item. if no item matches on any key, append to the list.
- if the node is non-leaf list, find an item that matches B item exactly and continue to traverse.
Otherwise if that proves to be too much work, you could just do a JSON strategic merge on the trees. We probably don't have to handle such tricky cases as above, we can probably assume that lists are non-overlapping, otherwise it's a user error. For example, if the same env var has two different values in different places, we can treat this as a bad configuration. If it is duplicated, I think strategic merge will handle it.

morvencao: The JSON merge package https://github.com/evanphx/json-patch we're using now doesn't handle slice/array well, it just replace all items with new items, instead of doing a JSON strategic merge, check this example: https://play.golang.org/p/woFuyU1vWKU

morvencao: Switched to strategic merge.

morvencao:
@ostromart PTAL

pkg/name/name.go
```diff
@@ -390,3 +390,75 @@ func Set(val, out interface{}) error {
...
+			currentNode, ok := currentNode.(map[string]interface{})
+			if !ok {
+				return nil, fmt.Errorf("path %s has an unexpected non KV element %s", path, pe)
+			}
+			currentNode[pe] = node
+			break
+		}
+
+		// if current path element is kc element, then append new item to parent list.
+		if util.IsKVPathElement(pe) {
```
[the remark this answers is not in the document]
ostromart: Somehow I didn't see the part in the code that returns in PathKV returns error (which it will it the path is not found). I think I was just confused by the comment above about appending to the parent list. We expect to find the item already in the parent, if not, then fail. If you agree, could you pleasure update the comment? And the check you added is unnecessary. Sorry for the confusion!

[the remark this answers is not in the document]
morvencao: hmm, got your point. The list item with strategic merge key is actually inserted into parent here, and it should returns error if PV is not valid; Also the parent slice is created with empty content in the previous iteration, see: https://github.com/istio/operator/blob/94630121fb028683e46c29714447b4e8a9731516/pkg/name/name.go#L459-L460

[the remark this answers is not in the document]
morvencao: done, removed the confusing comment.

ostromart (approved):

PR #249, opened 2019-09-04
Stabilizing manifest generation to create text diffable output

### LINKED ISSUE #15273: Operator should produce manifests that are diffable - in a consistent way
At present a textual diff between `iop manifest` and `iop manifest -f custom.yaml`  is enormous. The reason is that the manifests are produced in a non-repeatable way. Most likely this is due 
* Raging over Maps inside the code without sorting
* merging slices in an inconsistent way.

You can use 
`istio.io/istio/bin/diff_yaml.py` to see what the actual differences.

stewartbutler:
This PR stabilizes the output order of the `mesh generate` family of commands, which will cause them to produce text diffable output (e.g. you can now run `mesh generate > test.yaml`, `mesh generate > test2.yaml`, and then `diff test.yaml test2.yaml` and get a zero diff).

This fix is to close https://github.com/istio/istio/issues/15273, which I believe was created before the operator was forked into its own repository.

### DIFF
--- a/cmd/mesh/manifest-generate_test.go
+++ b/cmd/mesh/manifest-generate_test.go
@@ -93,6 +93,26 @@ func TestManifestGenerateTelemetry(t *testing.T) {
 	})
 }
 
+func TestManifestGenerateOrdered(t *testing.T) {
+	// Since this is testing the special case of stable YAML output order, it
+	// does not use the established test group pattern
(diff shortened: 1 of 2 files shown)


stewartbutler:
@googlebot I signed it!

ostromart (approved):

PR #250, opened 2019-09-04
Usage messages, grammar, and manifest apply --yes opt

### LINKED ISSUE #16814: Operator CLI grammar and argument checking
Grammar
Better messages for arguments
Require an argument for `istioctl x manifest apply` so that it doesn't immediately start reworking your cluster.

esnible:
Resolves https://github.com/istio/istio/issues/16814 by addition argument checking, usage hints for positional messages.

Requires `--yes` in `mesh apply` if the file, kubeconfig, or context has not been specified.  (This is to keep the user from typing it by accident and applying a change to her cluster!)

### DIFF
--- a/cmd/mesh/manifest-apply.go
+++ b/cmd/mesh/manifest-apply.go
@@ -49,7 +49,7 @@ func addManifestApplyFlags(cmd *cobra.Command, args *manifestApplyArgs) {
 	cmd.PersistentFlags().StringVarP(&args.inFilename, "filename", "f", "", filenameFlagHelpStr)
 	cmd.PersistentFlags().StringVarP(&args.kubeConfigPath, "kubeconfig", "c", "", "Path to kube config")
 	cmd.PersistentFlags().StringVar(&args.context, "context", "", "The name of the kubeconfig context to use")
-	cmd.PersistentFlags().BoolVar(&args.yes, "yes", false, "Do not ask for confirmation")
+	cmd.PersistentFlags().BoolVarP(&args.yes, "yes", "y", false, "Do not ask for confirmation")
 	cmd.PersistentFlags().DurationVar(&args.readinessTimeout, "readiness-timeout", 300*time.Second, "Maximum seconds to wait for all Istio resources to be ready."+
 		" The --wait flag must be set for this flag to apply")
 	cmd.PersistentFlags().BoolVarP(&args.wait, "wait", "w", false, "Wait, if set will wait until all Pods, Services, and minimum number of Pods "+
--- a/cmd/mesh/manifest-apply.go
+++ b/cmd/mesh/manifest-apply.go
@@ -40,12 +40,15 @@ type manifestApplyArgs struct {
 	// set is a string with element format "path=value" where path is an IstioControlPlane path and the value is a
 	// value to set the node at that path to.
 	set []string
+	// yes means don't ask for confirmation (asking for confirmation not implemented)
+	yes bool
 }
 
 func addManifestApplyFlags(cmd *cobra.Command, args *manifestApplyArgs) {
 	cmd.PersistentFlags().StringVarP(&args.inFilename, "filename", "f", "", filenameFlagHelpStr)
 	cmd.PersistentFlags().StringVarP(&args.kubeConfigPath, "kubeconfig", "c", "", "Path to kube config.")
 	cmd.PersistentFlags().StringVar(&args.context, "context", "", "The name of the kubeconfig context to use")
+	cmd.PersistentFlags().BoolVar(&args.yes, "yes", false, "Do not ask for confirmation")
 	cmd.PersistentFlags().DurationVar(&args.readinessTimeout, "readiness-timeout", 300*time.Second, "Maximum seconds to wait for all Istio resources to be ready."+
 		" The --wait flag must be set for this flag to apply.")
 	cmd.PersistentFlags().BoolVarP(&args.wait, "wait", "w", false, "Wait, if set will wait until all Pods, Services, and minimum number of Pods "+
@@ -61,6 +64,10 @@ func manifestApplyCmd(rootArgs *rootArgs, maArgs *manifestApplyArgs) *cobra.Comm
 		Args:  cobra.ExactArgs(0),
 		Run: func(cmd *cobra.Command, args []string) {
 			l := newLogger(rootArgs.logToStdErr, cmd.OutOrStdout(), cmd.OutOrStderr())
+			if !maArgs.yes && maArgs.kubeConfigPath == "" && maArgs.context == "" {
+				cmd.Print("Supply --kubeconfig, --context, or --yes to apply manifest to cluster\n")
+				os.Exit(1)
+			}
 			manifestApply(rootArgs, maArgs, l)
 		}}
 }
--- a/cmd/mesh/manifest-apply.go
+++ b/cmd/mesh/manifest-apply.go
@@ -16,6 +16,7 @@ package mesh
 
 import (
 	"fmt"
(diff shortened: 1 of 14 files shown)


esnible:
Need help fixing the lint problem.

First issue: I can't run `make lint` locally:
```
make lint
xargs: illegal option -- r
usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements]] [-J replstr]
             [-L number] [-n number [-x]] [-P maxprocs] [-s size]
             [utility [argument ...]]
```

Second issue: `struct of size 96 bytes could be of size 88 bytes:` is a weird problem to have with a singleton struct, and without an interactive _lint_ it is a pain to fix.

cmd/mesh/manifest-apply.go
```diff
@@ -37,6 +37,8 @@ type manifestApplyArgs struct {
 	readinessTimeout time.Duration
 	// wait is flag that indicates whether to wait resources ready before exiting.
 	wait bool
+	// yes means don't ask for confirmation (asking for confirmation not implemented)
+	yes bool
```
richardwxn: the flag name is not clear enough about what it is doing. BTW there is existing dryrun flag we can probably use instead of creating the new one for confirmation?

cmd/mesh/manifest-diff.go
```diff
@@ -58,10 +58,15 @@ func addManifestDiffFlags(cmd *cobra.Command, diffArgs *manifestDiffArgs) {
 
 func manifestDiffCmd(rootArgs *rootArgs, diffArgs *manifestDiffArgs) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "diff",
+		Use:   "diff <file|dir> <file|dir>",
 		Short: "Compare manifests and generate diff.",
 		Long:  "The diff subcommand compares manifests from two files or directories.",
-		Args:  cobra.ExactArgs(2),
+		Args: func(cmd *cobra.Command, args []string) error {
```
richardwxn: shouldn't cobra.ExactArgs(2) does the number of arguments check already?

cmd/mesh/manifest-diff.go
```diff
@@ -78,11 +83,13 @@ func compareManifestsFromFiles(rootArgs *rootArgs, args []string, selectResource
 
 	a, err := ioutil.ReadFile(args[0])
 	if err != nil {
+		fmt.Fprintf(os.Stderr, "Could not read %q: %v\n", args[0], err.Error())
```
richardwxn: there is logandprint helper function you may use here

cmd/mesh/manifest-apply.go
esnible: @richardwxn Good point.  I was thinking of the `--yes` in `apt-get install`.  I considered `--force`.  I considered implementing the "Are you sure" but we don't do that anywhere else in Istio and I didn't want to get this wrong so close to 1.3 release.

`<cmd> mesh apply` has a `--dry-run` flag, but the default is false.  I want some kind of feature that keeps me from changing my cluster if I type this command with no arguments.  Do you have a suggestion?

cmd/mesh/manifest-diff.go
esnible: @richardwxn There is a general request from @sdake to explain the problem.  See https://github.com/istio/istio/issues/16575

ExactArgs(2) outputs `Error: requires at least 1 arg(s), only received 0`.  The new code outputs `Error: diff requires two files or directories` then a usage message that includes `diff <file|dir> <file|dir>`.

cmd/mesh/manifest-apply.go
sdake: I'd suggest removing dry-run entirely from the commands as it defaults to false.  I think @esnible 's thinking on --yes makes sense given this is how most other distro tools operate.

sdake (commented):
looks like some good improvements.  need to come to some conclusion about changing the dry-run flag default, or removing it entirely and replacing with a `--yes` option.

I think your approach of asking via user input is good, although that should probably be done throughout istioctl if needed, and more consideration should be given to that design as you implied.

Cheers
-steve

sdake:
> Need help fixing the lint problem.
> 
> First issue: I can't run `make lint` locally:
> 
> ```
> make lint
> xargs: illegal option -- r
> usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements]] [-J replstr]
>              [-L number] [-n number [-x]] [-P maxprocs] [-s size]
>              [utility [argument ...]]
> ```
> 
> Second issue: `struct of size 96 bytes could be of size 88 bytes:` is a weird problem to have with a singleton struct, and without an interactive _lint_ it is a pain to fix.

your lint problem occured because you are running on a mac with BUILD_WITH_CONTAINER=0.  While BUILD_WITH_CONTAINER=0 should work well on a mac, it doesn't yet today.  There is a very strong push to move all repos to BUILD_WITH_CONTAINER=1 first and re-implementing the broken mac parts second.  I think some community members will not want BUILD_WTIH_CONTAINER=1, so we must be able to handle that on Macs.  Can you file an issue about the lint argument error so it can be tracked for 1.4 cleanup work?

Cheers
-steve

cmd/mesh/manifest-apply.go
esnible: `--dry-run` is nice and familiar.  I propose we keep it.  We still need to agree on a mechanism to make the no-arg default not do any unexpected work.

esnible:
> BUILD_WITH_CONTAINER

If I unset `BUILD_WITH_CONTAINER` then `make lint` reports

```
Missing license: ./data/charts/istio-telemetry/grafana/fix_datasources.sh
Missing copyright: ./data/charts/istio-telemetry/grafana/fix_datasources.sh
common/Makefile.common.mk:38: recipe for target 'lint-copyright-banner' failed
```

Is there a work item for operator developer instructions?  I don't want to have this conversation here?

cmd/mesh/manifest-apply.go
ostromart: There were previously complaints that dry-run by default is not intuitive and I agree with this. I also think we need to keep dry run as an option. 
My understanding is the concern is only about the no-arg version, as kubectl also lacks confirmation when a file is passed to apply. In that case we could have a CLI prompt for that version of the command only.

morvencao: +1 to keep `--dry-run` as it aims to print result in console only before user decides to apply it to cluster or write to file.

esnible: @ostromart added "Are you sure?" which only appears if the user supplies no targetting argument.

ostromart:
@esnible your lint problem is due to installer charts which are copied in during build. delete data/charts and the problem will go away.

cmd/mesh/manifest-apply.go
ostromart: sg. i'll let others approve the PR but overall lgtm.

esnible:
@sdake @morvencao morvencao @richardwxn @ostromart Although this is `size/L` it is just text nits and a scanf() question preventing automatic `manifest apply`.  I propose someone review this today, let it merge, and let me cherry pick it to 1.3.  Everything except the scanf() is trivial and the scanf() adds safety.

ostromart:
lgtm, one minor suggestion would be to include -y as a shortcut for --yes.
Otherwise, I'm happy if @sdake is happy, I'll let him approve. Please ping me if you can't get hold of him since this is time sensitive and in that case I can approve.

esnible:
@sdake I have add the `-y` as shorthand for `--yes` as suggested.  Please review.

ostromart (approved):

PR #251, opened 2019-09-04
Fix ordering of manifest diff output

### LINKED ISSUE #16834: manifest diff ordering should be stable
When multiple resources are different, the ordering changes non-deterministically. 
Ideally, the ordering should be grouped by kind, then namespace and name - this can be easily achieved by sorting the resource keys alphabetically.

ostromart:
Fixes https://github.com/istio/istio/issues/16834.


pkg/object/objects.go
```diff
@@ -465,17 +466,27 @@ func manifestDiff(aom, bom map[string]*K8sObject, im map[string]string) (string,
...
-			writeStringSafe(&sb, "\n\nObject "+ak+" has diffs:\n\n")
-			writeStringSafe(&sb, diff)
+			out[ak] = fmt.Sprint("\n\nObject "+ak+" has diffs:\n\n", diff)
 		}
 	}
 	for bk := range bom {
 		ao := aom[bk]
 		if ao == nil {
-			writeStringSafe(&sb, "\n\nObject "+bk+" is missing in A:\n\n")
+			out[bk] = fmt.Sprint("\n\nObject " + bk + " is missing in B:\n\n")
```
morvencao: `B` -> `A` ?

ostromart: Done.

richardwxn (approved):

elfinhe (approved):
LGTM

pkg/object/objects.go
```diff
@@ -448,14 +448,15 @@ func buildResourceRegexp(s string) (*regexp.Regexp, error) {
...
+	out := make(map[string]string)
 	for ak, av := range aom {
 		ay, err := av.YAML()
 		if err != nil {
 			return "", err
 		}
 		bo := bom[ak]
 		if bo == nil {
-			writeStringSafe(&sb, "\n\nObject "+ak+" is missing in B:\n\n")
+			out[ak] = fmt.Sprint("\n\nObject " + ak + " is missing in B:\n\n")
```
ayj: `fmt.Sprint` or string concatenation but not both at the same time, e.g.

```golang
out[ak] = `\n\nObject ` + ak + ` is missing in B:\n\n`
// or
out[ak] = fmt.Sprintf("`\n\nObject %v is missing in B:\n\n`, ak)
```

ostromart: Done, thanks.

ayj (approved):

PR #252, opened 2019-09-05
[WIP][TEST]add method for merge k8s json.

morvencao:
/hold
/do-not-merge
add experimental method for merge k8s json.

### DIFF
--- a/pkg/util/json.go
+++ b/pkg/util/json.go
@@ -1,0 +1,118 @@
+// Copyright 2019 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


pkg/util/json.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	case []interface{}:
+		overlay, ok := overlay.([]interface{})
+		if !ok {
+			return nil, fmt.Errorf("invalid type for current node: %s, expect slice type", currentNode)
+		}
+		dst := base[:0]
+		dst = append(dst, base...)
+		for _, v2 := range overlay {
+			// non-leaf list, expect to match item by key.
```
ostromart: This comment is a bit confusing... for non-leaf list I think we should do a full match of all keys. For leaf list, we would select best match (the list item with most keys matching) and if nothing matches, append.

ostromart:
Overall this is definitely on the right track, thanks for persisting with this challenging task. I'm not sure if the leaf/non-leaf key matching is quite right - see comment. Some UT's would be super useful to validate the behavior is as expected.

morvencao:
Closing this in favor of https://github.com/istio/operator/pull/247

PR #256, opened 2019-09-05
manifest diff: add text diff when --verbose flag is set.

### LINKED ISSUE #16827: mesh manifest diff should have text diff option
Sometimes, the text diff format (used in mesh profile diff) is easier to read than the compact diff style used in mesh manifest diff. 
The request is to use the --verbose flag to cause mesh manifest diff to output a text based diff instead.

elfinhe:
Resolve: https://github.com/istio/istio/issues/16827

manifest diff: add text diff when --verbose flag is set. 
Sometimes, the text diff format (used in mesh profile diff) is easier to read than the compact diff style used in mesh manifest diff. Reuse the existing tests to also test verbose output.

### DIFF
--- a/pkg/object/objects_test.go
+++ b/pkg/object/objects_test.go
@@ -630,13 +630,10 @@ spec:
 
 	for _, tt := range manifestDiffTests {
 		t.Run(tt.desc, func(t *testing.T) {
-			got, err := ManifestDiff(tt.yamlStringA, tt.yamlStringB)
-			if err != nil {
-				t.Fatalf("unexpected error: %v", err)
-			}
-			if !strings.Contains(got, tt.want) {
-				t.Errorf("%s:\ngot:\n%v\ndoes't contains\nwant:\n%v", tt.desc, got, tt.want)
-			}
+			got, err := ManifestDiff(tt.yamlStringA, tt.yamlStringB, false)
+			reportDiffErr(err, t, tt.desc, got, tt.want)
+			gotVerbose, err := ManifestDiff(tt.yamlStringA, tt.yamlStringB, true)
+			reportDiffErr(err, t, tt.desc, gotVerbose, tt.want)
 		})
 	}
 }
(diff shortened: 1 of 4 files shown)


pkg/object/objects_test.go
```diff
@@ -816,13 +813,21 @@ spec:
...
 	for _, tt := range manifestDiffWithSelectAndIgnoreTests {
 		t.Run(tt.desc, func(t *testing.T) {
-			got, err := ManifestDiffWithSelectAndIgnore(tt.yamlStringA, tt.yamlStringB, tt.selectResources, tt.ignoreResources)
-			if err != nil {
-				t.Fatalf("unexpected error: %v", err)
-			}
-			if !strings.Contains(got, tt.want) {
-				t.Errorf("%s:\ngot:\n%v\ndoes't contains\nwant:\n%v", tt.desc, got, tt.want)
-			}
+			got, err := ManifestDiffWithSelectAndIgnore(tt.yamlStringA, tt.yamlStringB,
```
ostromart: It's generally not good Go practice to pass testing into helper functions. You could restore the original check and save some duplication by adding an additional loop for v := range []bool{true, false}

pkg/object/objects_test.go
```diff
@@ -816,13 +813,21 @@ spec:
...
+				tt.selectResources, tt.ignoreResources, false)
+			reportDiffErr(err, t, tt.desc, got, tt.want)
+			gotVerbose, err := ManifestDiffWithSelectAndIgnore(tt.yamlStringA, tt.yamlStringB,
+				tt.selectResources, tt.ignoreResources, true)
+			reportDiffErr(err, t, tt.desc, gotVerbose, tt.want)
 		})
 	}
 }
+
+func reportDiffErr(err error, t *testing.T, desc, got, want string) {
```
ostromart: See above. It's probably better to leave that in the original location and remove duplication as suggested.

pkg/object/objects_test.go
elfinhe: sure, will do.

ostromart (approved):

PR #259, opened 2019-09-06
Fix proto patching hack

ostromart:
This cleanup removes the troublesome go structs patch step and replaces it with synthetic types. 
This is required to correctly auto-generate the docs from the proto.


pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```diff
@@ -552,6 +556,15 @@ message ObjectMeta {
...
 }
 
+// GOTYPE: map[string]interface{}
+message TypeMapStringInterface {}
+
+// GOTYPE: interface{}
+message TypeInterface {}
+
+// GOTYPE: *IntOrStringForPB
+message IntOrStringForPB {}
```
ayj: `TypeIntOrString` to match the other synthetic types?

ostromart: Done.

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```diff
@@ -544,6 +546,8 @@ message DeploymentStrategy {
 
 // Mirrors k8s.io.api.apps.v1.RollingUpdateDeployment for unmarshaling.
 message RollingUpdateDeployment {
+    TypeIntOrStringForPB maxUnavailable = 1;
```
ayj: ni: extra space here

ostromart: Done.

ayj (approved):

PR #261, opened 2019-09-06
Clean up IstioControlPlane proto for docs

ostromart:
This adds the required header to the proto so that istio.io make update_ref_docs will recognize it.


pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```diff
@@ -14,8 +14,16 @@
 
 syntax = "proto3";
 
+// $title: Operator Installation
+// $description: Configuration for Istio control plane installation through the Operator.
+// $location: https://istio.io/docs/reference/config/istio.operator.v1alpha12.pb.html
+
 package v1alpha2;
 
+// TODO: for some reason the below fails proto compilation in this repo.
```
ayj: Can you link to an issue for this TODO unless you're planning to fix this in a follow-up?

ostromart: https://github.com/istio/istio/issues/16899

ayj (approved):

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
ayj: Thanks!

PR #266, opened 2019-09-06
WIP: Add force flag to bypass validation

ostromart:
In case users have custom charts, or there's a problem with our schema, we need a bypass for the values validation. This flag will cause the tool to still report validation errors but proceed in spite of them.


richardwxn (commented):
It is good to go after rebase and addressing test failure

ostromart:
Moved to https://github.com/istio/operator/pull/353 due to some bad merges being pulled in.

PR #268, opened 2019-09-07
Fix doc generation.

geeknoid:
- Fix doc build on macOS.

- Reorg a bit of doc content so it gets recognized by the doc extractor

- Improve some of the doc content for better site formatting.


ostromart (approved):

ostromart:
Thanks for fixing this. You can run make update-goldens to refresh the test outputs, this is because https://github.com/istio/operator/pull/258 went in without the outputs being updated and somehow the tests passed. I'm looking into that, but running update-goldens should insert that extra parameter in your PR.

PR #272, opened 2019-09-09
Update values schema

richardwxn:


pkg/validate/validate_values.go
```diff
@@ -37,8 +37,8 @@ func CheckValues(root map[string]interface{}) util.Errors {
 	if err != nil {
 		return util.Errors{err}
 	}
-	val := &v1alpha2.Values{}
-	if err := yaml.Unmarshal(vs, val); err != nil {
+	val := &v1alpha22.Values{}
```
ostromart: I assume this is temporary - ultimately we'll want to move this into the same v1alpha2 package.

richardwxn: not sure whether we want to do so since there might be some conflict between each other if we put them into same package. Also essentially they belong to two groups of APIs, the values one is for validation purpose only. So I think it might better to keep it in different package.

cmd/mesh/testdata/profile-dump/output/all_off.yaml
```diff
@@ -641,7 +613,7 @@ values:
       hub: docker.io/omnition
       resources:
         limits:
-          cpu: 1
+          cpu: 1Gi
```
ostromart: Is this right? Seems like a lot of CPU :-)

pkg/apis/istio/v1alpha1/common.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2019 Istio Authors
```
ostromart: Suggest moving these out to pkg/apis/istio since they are used on all proto packages.

pkg/apis/istio/v1alpha1/fixup_structs/main.go
```diff
@@ -111,11 +111,34 @@ func main() {
...
+	if strings.Contains(filePath, "values_types") {
+		out = patchValues(out)
+	}
 	if err := ioutil.WriteFile(filePath, []byte(strings.Join(out, "\n")), 0644); err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
+}
 
+// patchValues is helper function to patch specific fields of the generated values_types.pb.go
```
ostromart: could you expand on why this is needed? it will be useful to maintainers trying to understand why this was created.

pkg/values/testdata/values/discovery.yaml
```diff
@@ -27,10 +27,10 @@ pilot:
     topologyKey: "kubernetes.io/hostname"
   podAntiAffinityTermLabelSelector: []
 
-  keepaliveMaxServerConnectionAge: 30m
+  keepaliveMaxServerConnectionAge: 1800s
 
-  deploymentLabels:
-    foo: bar
+#  Fix map[string]interface{}
```
ostromart: Is this a todo?

pkg/values/testdata/values/global.yaml
```diff
@@ -38,19 +36,15 @@ global:
...
     autoInject: enabled
     envoyStatsd:
       enabled: false
     envoyMetricsService:
       enabled: false
       host: metrics-service.istio-system
       port: 15000
-
-    tracer: "zipkin"
+#    tracer: "zipkin"
```
ostromart: TODO?

pkg/values/testdata/values/global.yaml
```diff
@@ -126,10 +114,11 @@ global:
         port: 443
 
   localityLbSetting:
-    distribute:
-    - from: "us-central1/*"
-      to:
-        "us-central1/*": 80
-        "us-central2/*": 20
+    enabled: false
+#  - distribute:
```
ostromart: Same here - if there are still some gaps maybe collect these into an issue?

pkg/apis/istio/v1alpha1/fixup_structs/main.go
```diff
@@ -111,11 +111,34 @@ func main() {
...
+func patchValues(lines []string) (output []string) {
+	for _, line := range lines {
+		// patching naming issues
+		if strings.Contains(line, "istioEgressgateway") {
```
morvencao: I think the if check is unnecessary.

pkg/apis/istio/v1alpha1/fixup_structs/main.go
```diff
@@ -111,11 +111,34 @@ func main() {
...
+			line = strings.ReplaceAll(line, "istioEgressgateway", "istio-egressgateway")
+		}
+		if strings.Contains(line, "istioIngressgateway") {
+			line = strings.ReplaceAll(line, "istioIngressgateway", "istio-ingressgateway")
+		}
+		if strings.Contains(line, "name=proxy_init") {
```
morvencao: `proxy_init` should be `proxyInit`?

pkg/apis/istio/v1alpha1/fixup_structs/main.go
```diff
@@ -111,11 +111,34 @@ func main() {
...
+			line = strings.ReplaceAll(line, "proxyInit", "proxy_init")
+		}
+		if strings.Contains(line, "name=istio_cni") {
```
morvencao: `istio_cni` to `istioCni`?

cmd/mesh/testdata/profile-dump/output/all_off.yaml
richardwxn: changed back to 1

pkg/values/testdata/values/discovery.yaml
richardwxn: as discussed offline, the values test would be removed

pkg/apis/istio/v1alpha1/fixup_structs/main.go
richardwxn: updated the logic

pkg/apis/istio/v1alpha1/fixup_structs/main.go
richardwxn: done

pkg/apis/istio/v1alpha1/fixup_structs/main.go
richardwxn: done

pkg/apis/istio/v1alpha1/fixup_structs/main.go
richardwxn: added more comments

pkg/apis/istio/v1alpha1/common.go
richardwxn: as discussed offline, keep them in both packages, otherwise it would mess up with the import path

Makefile.core.mk
```diff
@@ -87,7 +87,7 @@ $(types_v1alpha2_pb_gos) $(types_v1alpha2_pb_docs) $(types_v1alpha2_pb_pythons):
 	@$(protoc) $(go_plugin) $(protoc_gen_docs_plugin)$(types_v1alpha2_path) $(protoc_gen_python_plugin) $^
 	@cp -r ${TMPDIR}/pkg/* pkg/
 	@sed -i -e 's|github.com/gogo/protobuf/protobuf/google/protobuf|github.com/gogo/protobuf/types|g' $(types_v1alpha2_path)/istiocontrolplane_types.pb.go
-	@go run $(values_v1alpha2_path)/fixup_structs/main.go -f $(types_v1alpha2_path)/istiocontrolplane_types.pb.go
+	@go run $(values_v1alpha1_path)/fixup_structs/main.go -f $(types_v1alpha2_path)/istiocontrolplane_types.pb.go
```
ostromart: Sorry, I should have been more clear - I think fixup_structs still needs to be in a dir common to alpha1 and 2 - perhaps just move it up a level?

pkg/apis/istio/v1alpha1/fixup_structs/main.go
```diff
@@ -47,6 +47,15 @@ const (
 	goTypeToken = "// GOTYPE: "
 )
 
+var (
+	nameMapping = map[string]string{
```
ostromart: Could you put a comment with the reason for this mapping?

pkg/apis/istio/v1alpha1/fixup_structs/main.go
```diff
@@ -120,21 +128,14 @@ func main() {
 	}
 }
 
-// patchValues is helper function to patch specific fields of the generated values_types.pb.go
+// patchValues is helper function to patch specific naming of fields of the generated values_types.pb.go
+// some fields namings are inconsistent between proto file and values.yaml file, eg. istio-ingressgateway
+// is not a valid field name in proto but used in values.yaml.
```
ostromart: You can move some of this text as the explanation for the map where I left the other comment.

Makefile.core.mk
richardwxn: done

pkg/apis/istio/v1alpha1/fixup_structs/main.go
richardwxn: added

pkg/apis/istio/v1alpha1/fixup_structs/main.go
richardwxn: done

ostromart (approved):

PR #277, opened 2019-09-09
Improve error messages

richardwxn:
try to address some of the concern raised here: https://github.com/istio/istio/issues/16858

### DIFF
--- a/pkg/kubectlcmd/client.go
+++ b/pkg/kubectlcmd/client.go
@@ -22,6 +22,7 @@ import (
 	"os/exec"
 	"strings"
 
+	"istio.io/operator/pkg/util"
 	"istio.io/pkg/log"
 )
 
@@ -89,7 +90,7 @@ func (c *Client) Apply(dryRun, verbose bool, kubeconfig, context, namespace stri
 	log.Infof("applying to namespace %s:\n%s\n", namespace, cmdStr)
 
 	err := c.cmdSite.Run(cmd)
-	csError := consolidateError(stderr.String())
+	csError := util.ConsolidateLog(stderr.String())
 
 	if err != nil {
 		logAndPrint("error running kubectl apply: %s", err)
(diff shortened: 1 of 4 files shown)


pkg/kubectlcmd/client.go
```diff
@@ -89,14 +89,38 @@ func (c *Client) Apply(dryRun, verbose bool, kubeconfig, context, namespace stri
...
+		return stdout.String(), csError, fmt.Errorf("error running kubectl apply: %s", err)
 	}
 
 	logAndPrint("kubectl apply success")
 
-	return stdout.String(), stderr.String(), nil
+	return stdout.String(), csError, nil
+}
+
+// consolidateError is a helper function to dedup the error messages
```
ostromart: This doesn't have to be error specific, actually it's a useful general function for deduping log lines. I suggest changing the name and vars and putting it in util.

pkg/kubectlcmd/client.go
```diff
@@ -89,14 +89,38 @@ func (c *Client) Apply(dryRun, verbose bool, kubeconfig, context, namespace stri
...
+		} else {
+			errorMap[item] = 1
+		}
+	}
+	var sb strings.Builder
+	for msg, count := range errorMap {
+		if msg == "" {
+			continue
+		}
+		sb.WriteString(fmt.Sprintf("%s (repeated %v times)\n", msg, count))
```
ostromart: I think this will reorder the resulting text block. It's better to preserve the original ordering.

richardwxn: Assume that one line of error corresponds to one k8s object. If we want to keep the original ordering in this case, then we need to only suppress consecutive errors, right? However, even though we suppress consecutive errors only, the original ordering becomes less meaningful, because it is hard to match the consolidated from of errors with original k8s manifests

ostromart: Determinism is important in such cases because people expect it. It's confusing to users to run a command twice and see two different outputs for the same input.

pkg/kubectlcmd/client.go
richardwxn: ok moved it to util and make it general

pkg/kubectlcmd/client.go
richardwxn: ok I updated the logic to make the output consistent ordered.

ostromart (approved):

PR #283, opened 2019-09-16
add cni feature

### LINKED ISSUE #16955: CNI should be a part of core feature, rather than third party
Operator feature/component groupings currently put CNI into ThirdParty feature which is not accurate. It should be in its own feature group, or grouped with another Istio component if appropriate.

morvencao:
resolve: https://github.com/istio/istio/issues/16955


pkg/name/name.go
```diff
@@ -76,7 +76,7 @@ const (
 	PrometheusOperatorComponentName ComponentName = "PrometheusOperator"
 	GrafanaComponentName            ComponentName = "Grafana"
 	KialiComponentName              ComponentName = "Kiali"
-	CNIComponentName                ComponentName = "CNI"
+	CNIComponentName                ComponentName = "Cni"
```
ayj: Why not all caps? CNI is an initialism.

morvencao: We translate first character to lower case for `ComponentName` when translating `values` to `IstioControlPlaneSpec`, if we update `CNIComponentName` to all caps, then we need to treat it as an exception. see: https://github.com/istio/operator/blob/master/pkg/util/path.go#L60-L67

ayj: If feature and component name is always camel-cased to match value, would it make more sense to also define the constants here to match?

morvencao: It makes sense, but we need to refactor code about translation from values to IstioControlPlaneSpec.

ayj: Doesn't that mean just removing `firstCharToLowerCase` from `ToYAMLPath`? All of the should be using the constants defined here.

morvencao: I also thought it was, it's not that easy, because `ToYAMLPath` is general function invoked at couples of places, the path contains not only component name and feature name, also we should not keep operator relevant stuff in `util` package.

sdake: @ayj thiis PR LGTM.  I do not think we should block on this concern.  @morvencao can you file an issue tracker so we don't lose track of this particular issue and can treat it separately from the rest of the PR?

CHeers
-steve

morvencao: Filed an issue to track this: https://github.com/istio/istio/issues/17472

sdake (approved):

PR #286, opened 2019-09-18
support telemetry addons

irisdingbj:
https://github.com/istio/istio/issues/16970
https://github.com/istio/istio/issues/17068


irisdingbj:
@ostromart @richardwxn @morvencao Please take a look if this a correct direction for fixing this.  Thanks.  The test failures are most because the profile change.

data/profiles/default.yaml
```diff
@@ -131,6 +131,16 @@ spec:
             rollingUpdate:
               maxSurge: "100%"
               maxUnavailable: "25%"
+      tracing:
+        enabled: false
+      kiali:
+        enabled: false
+      grafana:
+        enabled: false
+      prometheus:
```
morvencao: do we really want to disable prometheus for default profile?

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```diff
@@ -256,6 +256,16 @@ message TelemetryFeatureSpec {
...
 
         TelemetryComponentSpec telemetry = 20;
+        // Optional
+        KialiComponentSpec kiali = 21;
+        // Optional
+        TracingComponentSpec tracing = 22;
+        // Optional
+        PrometheusComponentSpec prometheus = 23;
+        // Optional
+        PrometheusOperatorComponentSpec prometheusOperator = 24;
```
morvencao: prometheus operator will not work here if you see it from upstream: https://github.com/istio/installer/tree/master/istio-telemetry/prometheus-operator/templates
Since it just create prometheus operator api, we may need figure out how to integrate istio operator with these third party operators(prometheus, kiali, jaeger...)

irisdingbj: In cmd/mesh/manifest-generate_test.go:TestManifestGenerateFlags-all_on we expects the prometheus operator resource there.

data/profiles/default.yaml
irisdingbj: according to doc, we turned prometheus on for default profile. Will change this.

cmd/mesh/testdata/profile-dump/output/all_off.yaml
```diff
@@ -32,10 +32,10 @@ gateways:
         hpaSpec:
           maxReplicas: 5
           metrics:
-          - resource:
-              name: cpu
-              targetAverageUtilization: 80
-            type: Resource
+            - resource:
```
ostromart: how did these come to be indented? i just want to make sure it's not a manual change that would be reverted next time we regenerate.

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
ostromart: We already have prometheus and grafana under a third-party feature: 
https://github.com/istio/operator/blob/3a5dede6e691832e0be49e4080b62c4dba7ca9fe/pkg/component/feature/feature.go#L261
The way we support these is just in the code but not through this proto, so fields are just passed through without defining a schema of any kind. For now I suggest following the same approach, we can consider adding some kind of extensions framework in a future release.

irisdingbj: Currently,  ICP.RenderManifest() will loop all features in the ICP and generate manifest accordingly. Since the ThirdPartyFeature is not in ICP proto, so it will be ignored. This is also the reason why all add-on can not be installed even if you specify them in `Values` part. 
Do you suggest to put them into ThirdPartyFeature and add en entry into the ICP proto?

ostromart: There is a check in the code for ThirdParty feature so it does not need to be in the proto. I'm suggesting adding them as components in the code which will map to the correct chart directories. Then values can be passed through to the charts directly.

irisdingbj: added logic to check thirdparty feature from `Values` part instead of modifing the proto.

ostromart: I don't see any changes, did you push?

ostromart (commented):
Hi, thanks for the contribution!

Could you please capitalize the commit message per git conventions: https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53
Cheers,
Martin

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
irisdingbj: No, this is not the case. with current code base, you can not have prometheus installed with default or demo profile. as described in  istio/istio#16970. Even if we have `prometheus enabled :true` in the default.yaml.

So I will use this PR to submit the fix.

irisdingbj:
@ostromart @morvencao PTAL

ostromart:
@irisdingbj could you please confirm the indentation is not due to manual changes (like your editor)? Otherwise lgtm.

irisdingbj:
> @irisdingbj could you please confirm the indentation is not due to manual changes (like your editor)? Otherwise lgtm.

The indentation is not due to manual changes. They mainly caused by prometheus will be installed for default  profile.

ostromart (approved):

PR #290, opened 2019-09-20
Add obj diff inside the yaml text in ConfigMap data

### LINKED ISSUE #16828: Ability to diff ConfigMaps in istio manifest
ConfigMaps are typically text blocks of YAML and are unmarshaled as a single string, making YAML diffing impossible. Try making a small change in ConfigMap:istio-system:istio data.mesh and diffing the result with manifest diff to see the problem.
Some possible solutions:
1. Automatically try to unmarshal ConfigMap data string into YAML and if that succeeds, do a YAML diff of the result.
2. Add a flag like --is-yaml which is a list of paths that are known to be YAML strings

elfinhe:
Resolve: https://github.com/istio/istio/issues/16828

ConfigMaps are typically text blocks of YAML and are unmarshaled as a single string, making YAML diffing impossible. 

Solution: automatically try to unmarshal ConfigMap data string into YAML and if that succeeds, diff based on the unmarshaled yaml tree.

### DIFF
--- a/pkg/compare/compare.go
+++ b/pkg/compare/compare.go
@@ -92,12 +92,12 @@ func YAMLCmpWithIgnore(a, b string, ignorePaths []string) string {
 		return err.Error()
 	}
 
-	if kind, ok := ao["kind"]; ok && kind == "ConfigMap" {
+	if kind := ao["kind"]; kind == "ConfigMap" {
 		if err := UnmarshalInlineYaml(ao, "data"); err != nil {
 			log.Warnf("Unable to unmarshal ConfigMap Data, error: %v", err)
 		}
 	}
-	if kind, ok := bo["kind"]; ok && kind == "ConfigMap" {
+	if kind := bo["kind"]; kind == "ConfigMap" {
 		if err := UnmarshalInlineYaml(bo, "data"); err != nil {
 			log.Warnf("Unable to unmarshal ConfigMap Data, error: %v", err)
 		}
--- a/pkg/compare/compare.go
+++ b/pkg/compare/compare.go
@@ -23,6 +23,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	"istio.io/operator/pkg/tpath"
+	"istio.io/pkg/log"
 )
 
 // YAMLCmpReporter is a custom reporter to generate tree based diff for YAMLs, used by cmp.Equal().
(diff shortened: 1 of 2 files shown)


pkg/compare/compare.go
```diff
@@ -90,11 +90,56 @@ func YAMLCmpWithIgnore(a, b string, ignorePaths []string) string {
...
 	var r YAMLCmpReporter
 	cmp.Equal(ao, bo, cmp.Reporter(&r), genPathIgnoreOpt(ignorePaths))
 	return r.String()
 }
 
+func unmarshalConfigMapData(obj map[string]interface{}) {
+	InlineYamlData(obj, "ConfigMap", "data")
+}
+
+func InlineYamlData(obj map[string]interface{}, targetKind, targetPath string) {
```
richardwxn: maybe better to name it as sth like UnmarshalInlineYaml?

richardwxn: comment missing

pkg/compare/compare.go
```diff
@@ -90,11 +90,56 @@ func YAMLCmpWithIgnore(a, b string, ignorePaths []string) string {
...
+	nodeList := strings.Split(targetPath, ".")
+	if len(nodeList) == 0 {
+		return
+	}
+
+	cur := obj
+	for _, nname := range nodeList {
+		ndata, ok := cur[nname]
+		if !ok || ndata == nil {
+			return // target path does not exist
```
richardwxn: maybe helpful to add logging messages for some return cases?

pkg/compare/compare.go
```diff
@@ -90,11 +90,56 @@ func YAMLCmpWithIgnore(a, b string, ignorePaths []string) string {
...
+
+	unmarshalConfigMapData(ao)
+	unmarshalConfigMapData(bo)
+
 	var r YAMLCmpReporter
 	cmp.Equal(ao, bo, cmp.Reporter(&r), genPathIgnoreOpt(ignorePaths))
 	return r.String()
 }
 
+func unmarshalConfigMapData(obj map[string]interface{}) {
```
ostromart: Include a comment what this function does, esp. since it mutates the passed-in map.

pkg/compare/compare_test.go
```diff
@@ -662,3 +662,230 @@ metadata:
...
+
+func TestYAMLCmpWithYamlInline(t *testing.T) {
+	tests := []struct {
+		desc string
+		a    string
+		b    string
+		want interface{}
+	}{
+		{
+			desc: "ConfigMap data identical",
```
ostromart: you could probably omit this test, since the reordered test below is a superset.

pkg/compare/compare_test.go
```diff
@@ -662,3 +662,230 @@ metadata:
...
+  validatingwebhookconfiguration.yaml: |-
+    apiVersion: admissionregistration.k8s.io/v1beta1
+    kind: ValidatingWebhookConfiguration
+    metadata:
+      name: istio-galley`,
+			b: `kind: ConfigMap
+data:
+  validatingwebhookconfiguration.yaml: |-
+    apiVersion: admissionregistration.k8s.io/v1beta2
+    kind: ValidatingWebhookConfiguration
```
ostromart: suggest also reordering here to ensure it doesn't leak into output.

pkg/compare/compare_test.go
```diff
@@ -662,3 +662,230 @@ metadata:
...
+    metadata:
+      name: istio-galley`,
+			want: `data:
+  validatingwebhookconfiguration.yaml:
+    apiVersion: admissionregistration.k8s.io/v1beta1 -> admissionregistration.k8s.io/v1beta2
+`,
+		},
+		{
+			desc: "ConfigMap data string change",
```
ostromart: suggest omitting this test since the one below is a superset. this will reduce the amount of tests to update if anything changes.

ostromart (commented):
Some minor comments, otherwise LGTM. Thanks Tao!

pkg/compare/compare.go
```diff
@@ -90,11 +90,56 @@ func YAMLCmpWithIgnore(a, b string, ignorePaths []string) string {
...
+	InlineYamlData(obj, "ConfigMap", "data")
+}
+
+func InlineYamlData(obj map[string]interface{}, targetKind, targetPath string) {
+	if kind, ok := obj["kind"]; !ok || kind != targetKind {
+		return
```
morvencao: suggest to add some error message for unexpected return.

pkg/compare/compare.go
elfinhe: Updated, thanks!

elfinhe: Added, thanks so much.

pkg/compare/compare.go
elfinhe: will return an error and add logs.

pkg/compare/compare.go
elfinhe: Sure, comments added. Many thanks!

pkg/compare/compare.go
elfinhe: Move this "kind check" out of this func to decouple the responsibilities.

Also added errors and logs for other unexpected scenarios.

pkg/compare/compare_test.go
elfinhe: Sure, reordered.

pkg/compare/compare_test.go
elfinhe: Changed test description. This is intended as a deep nested change test.

elfinhe: Also simplified and improved the test input.

pkg/compare/compare_test.go
elfinhe: Sure, removed.

pkg/compare/compare.go
```diff
@@ -91,39 +92,43 @@ func YAMLCmpWithIgnore(a, b string, ignorePaths []string) string {
 		return err.Error()
 	}
 
-	unmarshalConfigMapData(ao)
-	unmarshalConfigMapData(bo)
+	if kind, ok := ao["kind"]; ok && kind == "ConfigMap" {
```
ostromart: ok is redundant here and below. ao["kind"] will return "" if no present.

elfinhe: Fixed, thanks!

ostromart (approved):

PR #292, opened 2019-09-20
Fix e2e test

howardjohn:


howardjohn:
Test works now but fails with a real error https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_operator/292/manifest-apply_operator/21

howardjohn:
Passing now! https://prow.istio.io/view/gcs/istio-prow/pr-logs/pull/istio_operator/292/manifest-apply_operator/24 it is hidden though, once this merges I'll make it not hidden

ayj (approved):

PR #295, opened 2019-09-23
Add test for setting output directory for manifest generate

### LINKED ISSUE #17230: Add output flag unit test
manifest-generate_test.go TestManifestGenerateFlags currently missing a test for output flag.

dgn:
Fixes https://github.com/istio/istio/issues/17230

Note that this also
- removes `cmd/mesh/testdata/manifest-generate/input/flag_set_values.yaml` (it was never read anyway)
- makes tests that do not read from an input YAML explicit using `noInput: true`
- adds two tests that combine `-s` and `-o` flags, one with input YAML, one without


cmd/mesh/manifest-generate_test.go
```diff
@@ -25,11 +27,17 @@ import (
...
 	desc       string
 	flags      string
+	noInput    bool
+	outputDir  string
 	diffSelect string
 	diffIgnore string
 }
 
 func TestManifestGenerateFlags(t *testing.T) {
+	outDirectory, err := ioutil.TempDir("", "output-testing")
```
ostromart: clean up the temp dir after test?

cmd/mesh/manifest-generate_test.go
```diff
@@ -120,11 +144,37 @@ func runTestGroup(t *testing.T, tests testGroup) {
...
+			if tt.noInput {
+				inPath = ""
+			}
+
 			got, err := runManifestGenerate(inPath, tt.flags)
 			if err != nil {
 				t.Fatal(err)
 			}
 
+			// if outputDir is set, walk the directory and concatenate all YAMLs
```
ostromart: util.ReadFilesWithFilter can do this work for you.

ostromart (commented):
Thanks for the contribution Daniel!

cmd/mesh/manifest-generate_test.go
dgn: I am cleaning up, but in `runTestGroup()` - which is not the right place. Will put it in here

cmd/mesh/manifest-generate_test.go
```diff
@@ -156,12 +194,30 @@ func runTestGroup(t *testing.T, tests testGroup) {
...
+	if path != "" {
+		args += " -f " + path
+	}
+	if flags != "" {
+		args += " " + flags
 	}
 	return runCommand(args)
 }
+
+func createTempDirOrFail(t *testing.T, prefix string) string {
```
ostromart: It's ok as is but in general it's best to avoid passing testing.T into helper functions, because it makes the execution of the code harder to reason about. It's best to stick with the pattern or returning error - even though I get that here the intention was to avoid cluttering the main function.

ostromart (approved):

PR #300, opened 2019-09-24
Move translations info table into translateConfig-1.3.yaml

irisdingbj:
https://github.com/istio/istio/issues/17238

### DIFF
--- a/data/translateConfig/translateConfig-1.3.yaml
+++ b/data/translateConfig/translateConfig-1.3.yaml
@@ -1,0 +1,203 @@
+apiMapping:
+  Hub:
+    outPath: "global.hub"
+  Tag:
+    outPath: "global.tag"
+  K8SDefaults:
+    outPath: "global.resources"
+  DefaultNamespace:
+    outPath: "global.istioNamespace"
+  Values.Proxy:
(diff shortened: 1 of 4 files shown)


ostromart (approved):
Great work Iris, thank you!

ostromart:
test_operator is failing because prometheus: enabled is flipping between true and false for some reason - I haven't yet investigated further.

irisdingbj:
> test_operator is failing because prometheus: enabled is flipping between true and false for some reason - I haven't yet investigated further.

@ostromart  PrometheusOperator and Prometheus have the same path `prometheus.enabled` defined in their Values.yaml: https://github.com/istio/installer/blob/master/istio-telemetry/prometheus-operator/values.yaml  this is reason why it is flipping.  Sorting the components for now to bypass the issue. Wondering do we need to fix this in installer's repo to make the path in `Values.yaml` for them are different?

ostromart:
The problem is more with how we are using the charts - installer does not define prometheus.enabled in global values. I think we should probably remove the per-chart values.yamls for the compiled-in charts to prevent this sort of hidden value problem in the future.

PR #301, opened 2019-09-25
Integrate v1alpha2 helm renderer with controller framework

richardwxn:
Initial commit to implement the helmreconciler logic.

Co-Authored-By: iamwen@google.com
Co-Authored-By: ostromart@gmail.com


pkg/controller/istiocontrolplane/inputsv2.go
```diff
@@ -0,0 +1,43 @@
...
+	return &IstioRenderingInputV2{instance: instance}
+}
+
+// GetCRPath returns the path of IstioControlPlane CR.
+func (i *IstioRenderingInputV2) GetCRPath() string {
+	return i.crPath
+}
+
+// GetProcessingOrder returns the order in which the rendered charts should be processed.
+func (i *IstioRenderingInputV2) GetProcessingOrder(manifests helmreconciler.ChartManifestsMap) ([]string, error) {
```
ostromart: I'm not sure if this is sufficient to express dependency relationships. Right now we have crds, which have to complete, then everything else, which can be done in parallel. 
I think what you have here implies that we serialize everything, which will make the apply run slower.

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
```diff
@@ -44,13 +45,17 @@ const finalizer = "istio-operator"
 // Add creates a new IstioControlPlane Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
-	return add(mgr, newReconciler(mgr))
+	return add(mgr, newReconciler(mgr, v1))
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+func newReconciler(mgr manager.Manager, apiVersion string) reconcile.Reconciler {
```
ostromart: Why not just have newReconcilerV1 and newReconcilerV2?

pkg/controller/istiocontrolplane/istiocontrolplane_controllerv2.go
```diff
@@ -0,0 +1,97 @@
...
+			reqLogger.Error(finalizerError, "error removing finalizer")
+		}
+		return reconcile.Result{}, err
+	} else if finalizerIndex < 0 {
+		reqLogger.V(1).Info("Adding finalizer", "finalizer", finalizer)
+		finalizers = append(finalizers, finalizer)
+		instance.SetFinalizers(finalizers)
+		err = r.client.Update(context.TODO(), instance)
+		return reconcile.Result{}, err
+	}
```
ostromart: The code above this line appears to be duplicated for v1 and v2. Maybe try to figure out a way to restructure to avoid the duplication?

pkg/controller/istiocontrolplane/listenersv2.go
```diff
@@ -0,0 +1,63 @@
...
+		instance:                 instance,
+	}
+}
+
+// EndReconcile updates the status field on the IstioControlPlane instance based on the resulting err parameter.
+//TODO
+func (u *IstioStatusUpdaterV2) EndReconcile(_ runtime.Object, err error) error {
+	status := u.instance.Status
+	vstatus := &v1alpha2.InstallStatus_VersionStatus{}
+	if err == nil {
```
ostromart: more idiomatic to write this as:
vstatus := &v1alpha2.InstallStatus_VersionStatus{
  Status: v1alpha2.InstallStatus_HEALTHY
}
if err != nil {
etc.

pkg/controller/istiocontrolplane/listenersv2.go
```diff
@@ -0,0 +1,63 @@
...
+// NewIstioStatusUpdater returns a new IstioStatusUpdater instance for the specified IstioControlPlane
+func NewIstioStatusUpdaterV2(instance *v1alpha2.IstioControlPlane) helmreconciler.RenderingListener {
+	return &IstioStatusUpdaterV2{
+		DefaultRenderingListener: &helmreconciler.DefaultRenderingListener{},
+		instance:                 instance,
+	}
+}
+
+// EndReconcile updates the status field on the IstioControlPlane instance based on the resulting err parameter.
+//TODO
```
ostromart: Could you state here what is still todo?

pkg/helmreconciler/interfaces.go
```diff
@@ -84,6 +84,18 @@ type RenderingInput interface {
 	GetProcessingOrder(manifests ChartManifestsMap) ([]string, error)
 }
 
+// RenderingInputV2 specifies the details used for rendering charts for v1alpha2 IstioControlPlane.
+type RenderingInputV2 interface {
+	// GetCRtPath returns the path locating the IstioControlPlane CR.
```
ostromart: typo

pkg/helmreconciler/interfaces.go
```diff
@@ -178,6 +190,13 @@ type ReconcilerListener interface {
 	RegisterReconciler(reconciler *HelmReconciler)
 }
 
+// ReconcilerListener is an interface that may be implemented by objects which require access to the HelmReconciler.
```
ostromart: This description is quite confusing. Could you take another stab at explaining what it's really for? You can start by looking at the methods in the interface as a starting point.

pkg/helmreconciler/manifest_reconciler.go
```diff
@@ -0,0 +1,114 @@
...
+		if err != nil {
+			allErrors = append(allErrors, err)
+		}
+		err = h.GetCustomizer().Listener().EndChart(chartName)
+		if err != nil {
+			allErrors = append(allErrors, err)
+		}
+	}
+
+	// delete any obsolete resources
```
ostromart: I think the code below is common - would it be possible to refactor to avoid duplication?

pkg/helmreconciler/manifest_rendering.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"istio.io/operator/pkg/component/controlplane"
+	operatormanifest "istio.io/operator/pkg/manifest"
+	"istio.io/operator/pkg/name"
+	"istio.io/operator/pkg/translate"
+)
+
+var (
+	kindRegex = regexp.MustCompile("kind:(.*)\n")
+	componentToChartName = map[name.ComponentName]string{
```
ostromart: Seems like this is already defined somewhere. It should also be versioned.

pkg/helmreconciler/manifest_rendering.go
```diff
@@ -0,0 +1,119 @@
...
+		if chartName == "" {
+			return nil, fmt.Errorf("no matching chart name for component name: %s", cn)
+		}
+		m := manifest.Manifest{Name: chartName, Content: v, Head: &releaseutil.SimpleHead{Kind: h}}
+		listManifests = append(listManifests, m)
+	}
+	return CollectManifestsByChart(listManifests), nil
+}
+
+func (rc *ISCPReconciler) ProcessManifests(manifests []manifest.Manifest) error {
```
ostromart: Is this any different from the other reconciler? Seems like duplication.

rcernich:
Is the goal behind this to simply swap the initial helm renderer with @ostromart's?

ostromart:
@rcernich, it's to add the new renderer without removing the old one. I'm not sure what the plan is for the older one so we're preserving it as is - should we remove it?

rcernich:
I assume we're going to be using the same input that goes through the cli, so I think we should just plugin `genProfile()` (or components within at least).  That will take care of the rendering (I think), which would only leave sorting and componentization, if necessary.  If the ordering of the resources in the returned `string` is adequate, then it just needs to be split into individual resource manifests.  FWIW, I'm in the middle of doing this now.

Validation should probably be done using a webhook, which is easy enough to add with the controller framework.  This also allows for semantic, in addition to syntactic, checking.

Regarding ordering, it would be helpful if CRD installation was handled separately, so it could be managed externally, either through operator installation (OLM metadata includes CRDs managed by the operator, which theoretically extends to the Istio components themselves) or through a bootstrapping.  This is how all cluster scoped resources should be handled.  (JFYI, Red Hat's Istio distribution supports multitenant installations of control planes (i.e. multiple, segregated control planes within the same k8s cluster, which of course could be linked using a typical multi-mesh configuration), so it helps if global resources are managed separately from control plane resources.

ostromart:
Yes, I think it's a good idea to put the validation in the webhook. 
I think the CRD installation is something we want to retain as part of the controller since we want a one step install solution, but I'm open to making that component optional if your use case requires that.

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
richardwxn: updated to have v1 and v2

pkg/apis/istio/v1alpha2/deepcopy.go
```diff
@@ -0,0 +1,43 @@
...
+package v1alpha2
+
+import (
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	"reflect"
+)
+
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
+func (in *IstioControlPlane) DeepCopyInto(out *IstioControlPlane) {
+	reflect.ValueOf(out).Elem().Set(reflect.ValueOf(in).Elem())
```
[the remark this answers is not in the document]
richardwxn: done

pkg/controller/istiocontrolplane/listenersv2.go
richardwxn: done

pkg/controller/istiocontrolplane/listenersv2.go
richardwxn: ok that's something for my personal reminder to check this part of logic. would remove at the end

pkg/helmreconciler/manifest_reconciler.go
richardwxn: done

pkg/controller/istiocontrolplane/inputsv2.go
sdake: fwiw, I don't think it is a grand idea to serialize.  I also don't think its a grand idea to run all of the chart installs in parallel.  While the iniital apply may be faster, what we are after is a faster overall startup experience.  The ordering should be something like
1. crds
1. galley
1. citadel (galley and citadel may need to be swapped)
1. everything else can likely be started in parallel after the above ordering is met.

Not sure where CNI fits in the ordering.  where is defaultProcessingOrder deifned?

I'd recommend a list of lists for this purpose.

pkg/controller/istiocontrolplane/istiocontrolplane_controllerv2.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+		reqLogger.Info("Deleting IstioControlPlane")
+
+		reconciler, err := r.factory.NewISCPReconciler(instance, r.client, reqLogger)
+		if err == nil {
+			err = reconciler.Delete()
+		} else {
+			reqLogger.Error(err, "failed to create reconciler")
+		}
+		// XXX: for now, nuke the resources, regardless of errors
```
sdake: what does XXX represent?  A TODO?  If so, please change it as such.

pkg/controller/istiocontrolplane/istiocontrolplane_controllerv2.go
```diff
@@ -0,0 +1,97 @@
...
+		finalizers = append(finalizers[:finalizerIndex], finalizers[finalizerIndex+1:]...)
+		instance.SetFinalizers(finalizers)
+		finalizerError := r.client.Update(context.TODO(), instance)
+		for retryCount := 0; errors.IsConflict(finalizerError) && retryCount < 5; retryCount++ {
```
sdake: retryCount of 5 seems low.  Should be in a const.

sdake (commented):
@richardwxn can you push your updated PR when possible?  We have a very tight schedule.  I'd prefer to iterate on the controller rather than have a perfect implementation going into the code base.  Please file issues for outstanding work items so we can distribute the work after submitting your updated working implementation.

Cheers
-steve

pkg/controller/istiocontrolplane/inputsv2.go
rcernich: CRDs should probably be managed as part of the operator install.  When using Operator Lifecycle Manager (OLM), OLM manages installation of all CRDs associated with an operator.  This prevents multiple operators that manage the same resource types from being installed at the same time, while also allowing for multiple instances of the operator (depending on which namespaces the operator is configured to manage).

FWIW, in our testing on Maistra, full install takes around two minutes, which includes waiting for deployments of each component to become available before proceeding to subsequent components.  Parallel install is much faster, but there is also a settling period.

rcernich: That said, it's unclear whether or not the operator should claim the resources used by the Istio components themselves (we've been debating that ourselves).

pkg/controller/istiocontrolplane/customizer.go
```diff
@@ -29,13 +30,20 @@ var _ helmreconciler.RenderingCustomizerFactory
...
+			InputValue:          NewIstioRenderingInput(v),
+			PruningDetailsValue: NewIstioPruningDetails(v),
+			ListenerValue:       NewIstioRenderingListener(v),
+		}, nil
+	case *v1alpha2.IstioControlPlane:
+		return &helmreconciler.SimpleRenderingCustomizer{
+			InputValue:          NewIstioRenderingInputV2(v),
+			PruningDetailsValue: NewIstioPruningDetailsV2(v),
+			ListenerValue:       NewIstioRenderingListenerV2(v),
+		}, nil
```
linsun: @ostromart @richardwxn , are we supporting both v1alph1 and v1alph2 at the same time?

pkg/controller/istiocontrolplane/inputsv2.go
```diff
@@ -0,0 +1,43 @@
...
+	"istio.io/operator/pkg/helmreconciler"
+)
+
+// IstioRenderingInput is a RenderingInput specific to an v1alpha2 IstioControlPlane instance.
+type IstioRenderingInputV2 struct {
+	instance *v1alpha2.IstioControlPlane
+	crPath   string
+}
+
+// NewIstioRenderingInput creates a new IstioRenderiongInput for the specified instance.
```
linsun: typo IstioRenderiongInput

pkg/controller/istiocontrolplane/inputsv2.go
linsun: I think I agree with CRDs should be installed as part of operator install.  This is pretty much what the operator framework recommends in their sample operator.  @richardwxn I assume you are not installing crds per this code because crds have its own helm chart anyway.

linsun: @sdake that is defined in inputs.go, currently set as ```
// defaultProcessingOrder for the rendered charts
var defaultProcessingOrder = []string{
	"istio",
	"istio/charts/security",
	"istio/charts/galley",
	"istio/charts/prometheus",
	"istio/charts/mixer",
	"istio/charts/pilot",
	"istio/charts/gateways",
	"istio/charts/sidecarInjectorWebhook",
	"istio/charts/grafana",
	"istio/charts/tracing",
	"istio/charts/kiali",
}
```

File: pkg/controller/istiocontrolplane/customizer.go (hunk 14 above)
Comment (ostromart) [reply pkg/controller/istiocontrolplane/customizer.go]: After discussing with @rcernich we are removing v1alpha1 as no-one will be using this going forward.

File: pkg/controller/istiocontrolplane/inputsv2.go (hunk 1 above)
Comment (ostromart) [reply pkg/controller/istiocontrolplane/inputsv2.go]: @linsun @rcernich we don't want to make OLM mandatory and what we have today handles  completely arbitrary dependency relationships (e.g. all wait for CRDs, mixer and policy wait for galley): https://github.com/istio/operator/blob/master/pkg/manifest/installer.go#L103
I don't see any reason to complicate matters by treating CRDs as anything other that just another component.

File: pkg/controller/istiocontrolplane/customizer.go (hunk 14 above)
Comment (linsun) [reply pkg/controller/istiocontrolplane/customizer.go]: ok i think that is reasonable. I'm quite confused when I saw v1alpha1 which is pretty much based on helm values if i understand it correctly.

File: pkg/controller/istiocontrolplane/inputsv2.go (hunk 1 above)
Comment (linsun) [reply pkg/controller/istiocontrolplane/inputsv2.go]: CRDs are global and you only apply once per cluster... and nothing specific to namespace.  Basically you'd run through that "applyall" code with every istio control plane resource change including CRDs.

File: pkg/controller/istiocontrolplane/istiocontrolplane_controller.go [hunk 16]
```diff
@@ -44,13 +45,17 @@ const finalizer = "istio-operator"
...
 	factory := &helmreconciler.Factory{CustomizerFactory: &IstioRenderingCustomizerFactory{}}
-	return &ReconcileIstioControlPlane{client: mgr.GetClient(), scheme: mgr.GetScheme(), factory: factory}
+	if apiVersion == v1 {
```
linsun: I assume no v1 is needed now?

pkg/controller/istiocontrolplane/istiocontrolplane_controllerv2.go
```diff
@@ -0,0 +1,97 @@
...
+			// workaround for https://github.com/kubernetes/kubernetes/issues/73098 for k8s < 1.14
+			reqLogger.Info("confilict during finalizer removal, retrying")
```
linsun: typo confilict

pkg/controller/istiocontrolplane/istiocontrolplane_controllerv2.go
```diff
@@ -0,0 +1,97 @@
...
+			finalizerIndex = indexOf(finalizers, finalizer)
+			finalizers = append(finalizers[:finalizerIndex], finalizers[finalizerIndex+1:]...)
+			instance.SetFinalizers(finalizers)
+			finalizerError = r.client.Update(context.TODO(), instance)
+		}
+		if finalizerError != nil {
+			reqLogger.Error(finalizerError, "error removing finalizer")
+		}
+		return reconcile.Result{}, err
+	} else if finalizerIndex < 0 {
```
linsun: does this indicate it is an add istio control plane resource?  Can we make info msg more intuitive to users?  I don't think finalizer will mean much to users.

pkg/controller/istiocontrolplane/listenersv2.go
```diff
@@ -0,0 +1,63 @@
...
+		vstatus.Status = v1alpha2.InstallStatus_HEALTHY
+	} else {
+		vstatus.Status = v1alpha2.InstallStatus_ERROR
+	}
+	status.TrafficManagement = vstatus
+	status.ConfigManagement = vstatus
+	status.PolicyTelemetry = vstatus
+	status.Security = vstatus
+	status.IngressGateway = vstatus
+	status.EgressGateway = vstatus
```
linsun: all these share the same status?

pkg/helmreconciler/interfaces.go
```diff
@@ -84,6 +84,18 @@ type RenderingInput interface {
...
 }
 
+// RenderingInputV2 specifies the details used for rendering charts for v1alpha2 IstioControlPlane.
+type RenderingInputV2 interface {
+	// GetCRtPath returns the path locating the IstioControlPlane CR.
+	GetCRPath() string
+	// GetProcessingOrder is a hook which allows a user to specify the order in which the generated charts
+	// should be applied.  manifests maps chart name to a list of manifests.  Examples of chart names:
+	// istio, istio/charts/security, istio/charts/galley, etc.  Subcharts will have the form:
+	// <main-chart-name>/charts/<subchart-name>
```
linsun: Does this include CRDS?

pkg/helmreconciler/manifest_rendering.go
```diff
@@ -0,0 +1,119 @@
...
+		name.IstioBaseComponentName:       "istio",
+		name.PilotComponentName:           "istio/charts/pilot",
+		name.GalleyComponentName:          "istio/charts/galley",
+		name.SidecarInjectorComponentName: "istio/charts/sidecarInjectorWebhook",
+		name.PolicyComponentName:          "istio/charts/mixer-policy",
+		name.TelemetryComponentName:       "istio/charts/mixer-telemetry",
+		name.CitadelComponentName:         "istio/charts/security-citadel",
+		name.CertManagerComponentName:     "istio/charts/security-certmanager",
+		name.NodeAgentComponentName:       "istio/charts/security-nodeagent",
+		name.IngressComponentName:         "istio/charts/gateways-istio-engress",
```
linsun: typo engress

pkg/helmreconciler/manifest_rendering.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+func (rc *ISCPReconciler) renderManifests(input RenderingInputV2) (ChartManifestsMap, error) {
+	cr, err := ioutil.ReadFile(input.GetCRPath())
+	if err != nil {
+		return nil, fmt.Errorf("could not read cr file %s: %s", input.GetCRPath(), err)
+	}
+	icps, _, err := operatormanifest.ParseK8SYAMLToIstioControlPlaneSpec(string(cr))
+	if err != nil {
+		return nil, fmt.Errorf("could not IstioControlPlane CR YAML:\n%s", cr)
```
linsun: Could not parse

pkg/helmreconciler/manifest_rendering.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	t, err := translate.NewTranslator(binversion.OperatorBinaryVersion.MinorVersion)
+	if err != nil {
+		return nil, err
```
linsun: can we give an error msg here too?

pkg/helmreconciler/pruning.go
```diff
@@ -24,22 +24,22 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func (h *HelmReconciler) prune(all bool) error {
+func (h *HelmReconciler) Prune(all bool) error {
```
linsun: Can we add a comment for Prune?  Is it to remove dups?

pkg/controller/istiocontrolplane/inputsv2.go
sdake: > @linsun @rcernich we don't want to make OLM mandatory and what we have today handles completely arbitrary dependency relationships (e.g. all wait for CRDs, mixer and policy wait for galley): https://github.com/istio/operator/blob/master/pkg/manifest/installer.go#L103
> I don't see any reason to complicate matters by treating CRDs as anything other that just another component.

@ostromart I am not clear how OLM was conflated with ordering.  I'd like to clear a few things up so we can avoid confusion in the future.

1. OLM is a framework for managing the operator lifecycle.  It simply requires a controller that operates based upon changes to a CRD (IstioControlPlane).  Nothing more.
1. The discussion about operator-sdk is not relevant here.  As far as I can tell, there is no strong motivation for Istio to take a hard dependency on operator-sdk.
1. OLM and operator-sdk are two distinct technologies.
1. OLM consumes an operator, therefore, Istio's operator would never take a hard dependency on OLM.  Instead, OLM would take a (soft) dependency on the Istio operator.  
1. All OLM requires is 1) a bunch of metadata 2) an operator that functions properly.
1. I recommend a focus on _merging_ a controller that functions properly and worry about OLM after we have a functional controller in the tree.

As for ordering - this is critical to get right, and I'm not clear this ordering approach is correct.  We can iterate independently on improvements once this PR merges rather than getting jammed up in a debate about how the implementation should be.

sdake: > @sdake that is defined in inputs.go, currently set as
> 
> ```
> // defaultProcessingOrder for the rendered charts
> var defaultProcessingOrder = []string{
> 	"istio",
> 	"istio/charts/security",
> 	"istio/charts/galley",
> 	"istio/charts/prometheus",
> 	"istio/charts/mixer",
> 	"istio/charts/pilot",
> 	"istio/charts/gateways",
> 	"istio/charts/sidecarInjectorWebhook",
> 	"istio/charts/grafana",
> 	"istio/charts/tracing",
> 	"istio/charts/kiali",
> }
> ```
Thanks @linsun. I'll wait for this PR to merge, then take on sorting out a proper start order and synchronization.  The ordering should be (off top of my head):

1. CRDs
1. security
1. galley
1. sidecar-injector
1. pilot, mixer, gateways, prometheus
1. grafana, tracing, kiali

In the meantime, @richardwxn can you get moving on this PR?  At this pace, we will have to pull the operator from our 1.4 roadmap...

pkg/helmreconciler/manifest_rendering.go
sdake: errors should be returned in the stack rather than logged per function.  I think the assumption here is that the result of translate.NewTranlater is properly formatted.

pkg/controller/istiocontrolplane/inputsv2.go
linsun: Also @richardwxn @ostromart @sdake I think there are 2 types of CRDs and let's not confuse with them.  1) CRD is the istio control plane CRD (do we have other CRDs??) used by the operator.  2) CRDs used by istio control plane, we currently have ~20 CRDs or so.  For 2, I agree, it should be handled as the rest of Istio components.  For 1, I think the istio control plane CRD should be installed whenever the istio operator is installed.

richardwxn:
@ostromart @sdake @linsun @rcernich, thanks for the comments. I would publish a newer revision soon when I am back to office next week.

pkg/helmreconciler/pruning.go
ostromart: Done.

pkg/controller/istiocontrolplane/istiocontrolplane_controllerv2.go
ostromart: Done.

pkg/controller/istiocontrolplane/istiocontrolplane_controllerv2.go
ostromart: Done.

pkg/controller/istiocontrolplane/inputsv2.go
ostromart: Done.

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
ostromart: Removed.

pkg/controller/istiocontrolplane/istiocontrolplane_controllerv2.go
ostromart: Done.

pkg/controller/istiocontrolplane/istiocontrolplane_controllerv2.go
ostromart: I think this is intended to be for debugging - have changed the level to 2.

pkg/controller/istiocontrolplane/listenersv2.go
ostromart: Created a TODO.

pkg/apis/istio/v1alpha2/deepcopy.go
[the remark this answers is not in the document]
ostromart: Fixed using marshal/unmarshal.

pkg/controller/istiocontrolplane/inputsv2.go
ostromart: @linsun, the latest code does exactly as you say. The Istio CRDs are treated as a base component by the operator and all other components have a dependency on it. Beyond that, I have all other components as siblings - the relationship is defined in inputs.go componentDependencies and applied in reconciler.go processRecursive(). But that can be changed if we desire a more complex dependency tree (e.g. wait for citadel etc.).
The operator itself is installed using a bootstrap manifest (see ./build and ./deploy directories) and uses the MeshConfig CR, which is installed alongside it.

pkg/helmreconciler/interfaces.go
ostromart: Yes, that is one of the components. See inputs.go componentDependencies for the dependency tree currently defined - the CRDs are the root of the tree.

pkg/helmreconciler/manifest_rendering.go
ostromart: Removed.

pkg/helmreconciler/manifest_rendering.go
ostromart: Done.

ostromart:
@richardwxn - sorry for taking over your PR but there's some urgency to get this in. I worked on top of your original commit which is still included with your authorship. 
@rcernich could you please take a look? Per our chat at the Monday meeting, I abandoned preserving the original v1alpha1 renderer and changed the code to incorporate the new renderer. 
I had to make some small changes to the APIs - dependencies are now expressed as a tree rather than slice and the rendering inputs are an interface{} type rather than values map to allow different CRD structs to be used. I wonder if it would be possible to take a pass at simplifying the framework somewhat - I found it quite hard to navigate all the abstractions and I imagine future authors will have similar difficulty. If it's not possible to simplify, it would be very useful to have an architecture doc to explain the design for those ramping up. But ideally, it would be great to prune back some of the abstractions at least for the time being to make it easier for folks to contribute - we can always bring them back if we find ourselves managing multiple implementations of reconcilers.
@linsun and @sdake thanks for the reviews - I think I addressed your comments but please take another pass if you have time.

pkg/helmreconciler/pruning.go
```diff
@@ -24,6 +24,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// Prune removes any resources not specified in manifests generated by h. If all is set to true, it prunes all
```
sdake: `by h` -> `by the HelmReconciler`.  `it prunes` -> `this function prunes`

pkg/helmreconciler/pruning.go
```diff
@@ -39,6 +41,8 @@ func (h *HelmReconciler) Prune(all bool) error {
 	return utilerrors.NewAggregate(allErrors)
 }
 
+// Prune removes any resources not specified gvks. If all is set to true, it prunes all
```
sdake: gvks->groupVersionKindSpec - I'd change the function prototype as well.

pkg/controller/istiocontrolplane/inputs.go
```diff
@@ -31,12 +62,11 @@ func (i *IstioRenderingInput) GetTargetNamespace() string {
...
+}
 
-	// everything else to the end
-	for chart := range manifests {
-		ordering = append(ordering, chart)
-	}
-	return ordering, nil
+func buildInstallTree() {
+	// Starting with root, recursively insert each first level child into each node.
+	helmreconciler.InsertChildrenRecursive(name.IstioBaseComponentName, installTree, componentDependencies)
```
sdake: recursion is taught by universities to make programming harder and weed out students...  It would be better if this could be unrolled ;)

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
```diff
@@ -140,9 +144,9 @@ func (r *ReconcileIstioControlPlane) Reconcile(request reconcile.Request) (recon
 		finalizers = append(finalizers[:finalizerIndex], finalizers[finalizerIndex+1:]...)
 		instance.SetFinalizers(finalizers)
 		finalizerError := r.client.Update(context.TODO(), instance)
-		for retryCount := 0; errors.IsConflict(finalizerError) && retryCount < 5; retryCount++ {
+		for retryCount := 0; errors.IsConflict(finalizerError) && retryCount < finalizerMaxRetries; retryCount++ {
 			// workaround for https://github.com/kubernetes/kubernetes/issues/73098 for k8s < 1.14
-			reqLogger.Info("confilict during finalizer removal, retrying")
+			reqLogger.Info("conflict during finalizer removal, retrying")
```
sdake: a human operator will not have the faintest clue what this info means.  I dont htink a log here is really necessary, unless this is just for debug.  Might be better to lower the level of this error if the string is kept as is.

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
```diff
@@ -155,7 +159,7 @@ func (r *ReconcileIstioControlPlane) Reconcile(request reconcile.Request) (recon
 		}
 		return reconcile.Result{}, err
 	} else if finalizerIndex < 0 {
-		reqLogger.V(1).Info("Adding finalizer", "finalizer", finalizer)
+		reqLogger.V(2).Info("Adding finalizer", "finalizer", finalizer)
```
sdake: same comment as above.

pkg/helmreconciler/dependencies.go
```diff
@@ -0,0 +1,31 @@
...
+	"io"
+	"strings"
+
+	"istio.io/operator/pkg/name"
+)
+
+func InsertChildrenRecursive(componentName name.ComponentName, tree ComponentTree, children ComponentNameToListMap) {
+	tree[componentName] = make(ComponentTree)
+	for _, child := range children[componentName] {
+		InsertChildrenRecursive(child, tree[componentName].(ComponentTree), children)
```
sdake: this makes my eyes bleed 👍

pkg/helmreconciler/dependencies.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+}
+
+func InstallTreeString(ct ComponentTree) string {
+	var sb strings.Builder
+	buildInstallTreeString(ct, name.IstioBaseComponentName, "", &sb)
+	return sb.String()
+}
+
+func buildInstallTreeString(ct ComponentTree, componentName name.ComponentName, prefix string, sb io.StringWriter) {
```
sdake: I think this would be a little more idomatic by expanding ct->componentTree, sb -> stringWriter.

pkg/helmreconciler/dependencies.go
```diff
@@ -0,0 +1,31 @@
...
+	_, _ = sb.WriteString(prefix + string(componentName) + "\n")
+	if _, ok := ct[componentName].(ComponentTree); !ok {
+		return
+	}
+	for k := range ct[componentName].(ComponentTree) {
+		buildInstallTreeString(ct, k, prefix+"  ", sb)
```
sdake: this is fine for this PR, but I really really think you, or I, or someone else working on this code should reimplement this recursion into a non-recursive implementation.  recursion is hard to test and hard to follow.

pkg/helmreconciler/interfaces.go
```diff
@@ -77,11 +79,10 @@ type RenderingInput interface {
...
 	// (i.e. used to set Release.Namespace)
 	GetTargetNamespace() string
-	// GetProcessingOrder is a hook which allows a user to specify the order in which the generated charts
-	// should be applied.  manifests maps chart name to a list of manifests.  Examples of chart names:
-	// istio, istio/charts/security, istio/charts/galley, etc.  Subcharts will have the form:
-	// <main-chart-name>/charts/<subchart-name>
-	GetProcessingOrder(manifests ChartManifestsMap) ([]string, error)
+	// GetProcessingOrder returns a dependency tree for the given manifests. ComponentNameToListMap is a map of
+	// each component to its dependencies. DependencyWaitCh is a map of channels, indexed by name. The component with
+	// the given name must wait on the channel before starting its processing.
```
sdake: 👍

pkg/helmreconciler/reconciler.go
```diff
@@ -95,71 +100,53 @@ func (h *HelmReconciler) Reconcile() error {
 	// render charts
 	manifestMap, err := h.renderCharts(h.customizer.Input())
 	if err != nil {
-		err = errors.Wrap(err, fmt.Sprintf("error rendering charts"))
-		listenerErr := h.customizer.Listener().EndReconcile(h.instance, err)
```
sdake: I am unclear if this lgoic needs to be retained to clean up the listeners in a failed reconcile..

pkg/helmreconciler/reconciler.go
```diff
@@ -95,71 +100,53 @@ func (h *HelmReconciler) Reconcile() error {
...
-			allErrors = append(allErrors, err)
-		}
-		err = h.customizer.Listener().EndChart(chartName)
-		if err != nil {
-			allErrors = append(allErrors, err)
-		}
-	}
+	errs := h.processRecursive(manifestMap)
 
 	// delete any obsolete resources
```
sdake: this comment could use some work.  I am left wondering what `obsolete` means in this context.

pkg/helmreconciler/reconciler.go
```diff
@@ -95,71 +100,53 @@ func (h *HelmReconciler) Reconcile() error {
...
+// processRecursive processes the given manifests in an order of dependencies defined in h. Dependencies are a tree,
+// where a child must wait for the parent to complete before starting.
+func (h *HelmReconciler) processRecursive(manifests ChartManifestsMap) util.Errors {
+	deps, dch := h.customizer.Input().GetProcessingOrder(manifests)
+	var errs []error
+
+	var wg sync.WaitGroup
+	for c, m := range manifests {
+		c := c
+		m := m
```
sdake: surprised these two lines pass the linter...  Imo using the same variables during a scope change in bad practice;.

pkg/helmreconciler/reconciler.go
```diff
@@ -95,71 +100,53 @@ func (h *HelmReconciler) Reconcile() error {
...
+		wg.Add(1)
+		go func() {
+			if s := dch[name.ComponentName(c)]; s != nil {
+				log.Infof("%s is waiting on parent dependency...", c)
```
sdake: `%s is waiting on a parent.` . This might be better served at a lower log level.  A parent can't be a dependency.   The log mesage isn't very clear.

pkg/helmreconciler/reconciler.go
```diff
@@ -95,71 +100,53 @@ func (h *HelmReconciler) Reconcile() error {
...
+				<-s
+				log.Infof("Parent dependency for %s has unblocked, proceeding.", c)
```
sdake: has unblocked, proceeding->has completed"

IIRC log messages should not include periods as the logging infrastructure will add those (depending on logging library).

pkg/helmreconciler/reconciler.go
```diff
@@ -95,71 +100,53 @@ func (h *HelmReconciler) Reconcile() error {
...
+			}
+
+			if len(m) != 0 {
+				errs = util.AppendErr(errs, h.ProcessManifest(m[0]))
+			}
+
+			// Signal all the components that depend on us.
+			for _, ch := range deps[name.ComponentName(c)] {
+				log.Infof("unblocking child dependency %s.", ch)
```
sdake: can remove the word child.  the word dependency implies that the component is a dependency.

pkg/helmreconciler/rendering.go
```diff
@@ -146,41 +146,32 @@ func unmarshalAndValidateICPS(icpsYAML string) (*v1alpha2.IstioControlPlaneSpec,
...
-			h.logger.V(2).Info("Skipping rendering of manifest")
+func (h *HelmReconciler) ProcessManifest(manifest manifest.Manifest) error {
+	var errs []error
+	log.Info("Processing resources from manifest")
+	// split the manifest into individual objects
+	objects := releaseutil.SplitManifests(manifest.Content)
+	for _, raw := range objects {
+		rawJSON, err := yaml.YAMLToJSON([]byte(raw))
+		if err != nil {
+			h.logger.Error(err, "unable to convert raw data to JSON")
```
sdake: I'm unclear if a log message is necessary here.  Better would be to produce an error string and append that to errs in line 158.

pkg/helmreconciler/rendering.go
```diff
@@ -123,7 +123,7 @@ func unmarshalAndValidateICP(crYAML string) (*v1alpha2.IstioControlPlaneSpec, st
 	}
 	icps, _, err := istiomanifest.ParseK8SYAMLToIstioControlPlaneSpec(crYAML)
 	if err != nil {
-		return nil, "", fmt.Errorf("could not unmarshal the overlay file: %s\n\nOriginal YAML:\n%s", err, crYAML)
+		return nil, "", fmt.Errorf("could not parse the overlay file: %s\n\nOriginal YAML:\n%s", err, crYAML)
```
sdake: is the `\n` necessary?

sdake (commented):
LGTM although not satisfied with the recursion.  Most of my comments are nitpicks that are easy to resolve.  I feel comfortable merging this as is assuming the checks are passing.

sdake:
@ostromart I'm satisfied with the overall PR.  The checks need resolution to merge as I don't wish to break the repository with a force merge.

Cheers
-steve

pkg/helmreconciler/rendering.go
```diff
@@ -27,68 +38,112 @@ import (
...
+	// Merge base and overlay.
+	mergedYAML, err := helm.OverlayYAML(baseYAML, overlayYAML)
+	if err != nil {
+		return nil, fmt.Errorf("could not overlay user config over base: %s", err)
+	}
+	return unmarshalAndValidateICPS(mergedYAML)
+}
+
+func unmarshalAndValidateICP(crYAML string) (*v1alpha2.IstioControlPlaneSpec, string, error) {
+	// TODO: add GVK handling as appropriate.
```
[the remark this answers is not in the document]
sdake: GVK = groupVersionKind, might be better to not abbreviate here.

pkg/apis/addtoscheme_istio_v1alpha1.go
```diff
@@ -15,10 +15,10 @@
 package apis
 
 import (
-	"istio.io/operator/pkg/apis/istio/v1alpha1"
+	"istio.io/operator/pkg/apis/istio/v1alpha2"
```
linsun: @ostromart I noticed this go file is still called *_v1alpha1.go, do we need to change it?

pkg/controller/istiocontrolplane/inputs.go
```diff
@@ -1,85 +1,42 @@
...
-	"istio/charts/kiali",
-}
-
-// IstioRenderingInput is a RenderingInput specific to an IstioControlPlane instance.
+// IstioRenderingInput is a RenderingInput specific to an v1alpha2 IstioControlPlane instance.
 type IstioRenderingInput struct {
-	instance  *v1alpha1.IstioControlPlane
-	chartPath string
+	instance *v1alpha2.IstioControlPlane
+	crPath   string
```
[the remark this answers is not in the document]
linsun: chart path?  seems prior name is more intuitive.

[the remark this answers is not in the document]
linsun: istio control plane resource's path, should be whatever user specifies.

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
```diff
@@ -28,13 +30,17 @@ import (
...
-	istiov1alpha1 "istio.io/operator/pkg/apis/istio/v1alpha1"
 	"istio.io/operator/pkg/helmreconciler"
 )
 
 var log = logf.Log.WithName("controller_istiocontrolplane")
 
-const finalizer = "istio-operator"
+const (
+	finalizer = "istio-operator"
+	v1        = "v1alpha1"
```
linsun: is v1alpha1 correct here?

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
linsun: or reword is to be more meaningful, like Error during istio control plane removal, retrying

pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -83,47 +77,18 @@ func NewIstioStatusUpdater(instance *v1alpha1.IstioControlPlane) helmreconciler.
...
-			})
-		} else {
-			status.SetCondition(v1alpha1.Condition{
-				Type:    v1alpha1.ConditionTypeReconciled,
-				Reason:  v1alpha1.ConditionReasonReconcileError,
-				Status:  v1alpha1.ConditionStatusFalse,
-				Message: fmt.Sprintf("errors occurred during reconciliation: %s", err),
-			})
-		}
+		vstatus.Status = v1alpha2.InstallStatus_ERROR
```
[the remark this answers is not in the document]
linsun: +1 would be good to have a message field and user can see it

pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -82,48 +76,20 @@ func NewIstioStatusUpdater(instance *v1alpha1.IstioControlPlane) helmreconciler.
...
 	}
+	status.TrafficManagement = vstatus
+	status.ConfigManagement = vstatus
+	status.PolicyTelemetry = vstatus
+	status.Security = vstatus
+	status.IngressGateway = vstatus
+	status.EgressGateway = vstatus
```
linsun: This could be error prune because I may not ask for installing policy telemetry and I'm getting status Healthy because my pilot is healthy.

pkg/helmreconciler/reconciler.go
```diff
@@ -204,3 +191,13 @@ func (h *HelmReconciler) GetLogger() logr.Logger {
 func (h *HelmReconciler) GetClient() client.Client {
 	return h.client
 }
+
+// GetClient returns the customizer associated with this HelmReconciler
```
linsun: GetCustomizer

pkg/helmreconciler/reconciler.go
```diff
@@ -204,3 +191,13 @@ func (h *HelmReconciler) GetLogger() logr.Logger {
...
+func (h *HelmReconciler) GetCustomizer() RenderingCustomizer {
+	return h.customizer
+}
+
+// GetClient returns the instance associated with this HelmReconciler
```
linsun: GetInstance

pkg/helmreconciler/rendering.go
```diff
@@ -17,117 +17,164 @@ package helmreconciler
...
 )
 
-func (h *HelmReconciler) renderCharts(input RenderingInput) (ChartManifestsMap, error) {
-	rawVals, err := yaml.Marshal(input.GetValues())
-	if err != nil {
-		return ChartManifestsMap{}, err
+func (h *HelmReconciler) renderCharts(in RenderingInput) (ChartManifestsMap, error) {
+	icp, ok := in.GetInputConfig().(*v1alpha2.IstioControlPlaneSpec)
+	if !ok {
+		return nil, fmt.Errorf("unexpected type %T in renderCharts", in.GetInputConfig())
```
linsun: is this any related to helm charts?

linsun: ok I see, later on it is related.

pkg/helmreconciler/rendering.go
```diff
@@ -17,117 +17,164 @@ package helmreconciler
...
-		// paths always start with the root chart name and always have a template
-		// name, so we should be safe not to check length
-		if pathSegments[1] == "charts" {
-			// subcharts will have names like <root-name>/charts/<subchart-name>/...
-			chartName = strings.Join(pathSegments[:3], "/")
+func mergeICPSWithProfile(icp *v1alpha2.IstioControlPlaneSpec) (*v1alpha2.IstioControlPlaneSpec, error) {
+	profile := icp.Profile
+
+	// This contains the IstioControlPlane CR.
+	baseCRYAML, err := helm.ReadProfileYAML(profile)
```
linsun: would no profile specified in istio control plane spec means default profile?

linsun:
@ostromart thank you for picking it up, I think it is a really good starting point.

pkg/apis/istio/v1alpha2/deepcopy.go
```diff
@@ -15,13 +15,22 @@
...
+	"github.com/prometheus/common/log"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"istio.io/operator/pkg/util"
 )
 
 // DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
 func (in *IstioControlPlane) DeepCopyInto(out *IstioControlPlane) {
-	reflect.ValueOf(out).Elem().Set(reflect.ValueOf(in).Elem())
+	inj, err := util.MarshalWithJSONPB(in)
```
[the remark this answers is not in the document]
nrjpoddar: There's a better to write/generate deepcopy functions.

You should use `proto.Clone()` to do a correct deepcopy at the top level proto message.

Here's an example:
```go
package v1alpha1

import (
	"github.com/gogo/protobuf/proto"

)

// DeepCopyInto supports using Application within kubernetes types, where deepcopy-gen is used.
func (in *IstioControlPlane) DeepCopyInto(out *IstioControlPlane) {
	*out = *proto.Clone(in).(*IstioControlPlane)
}
```

We also have tools to auto-generate all deepcopy functions for you if you don't want to write this manually.

@ostromart fyi

pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -45,14 +45,19 @@ type IstioRenderingListener struct {
...
+type IstioStatusUpdater struct {
+	*helmreconciler.DefaultRenderingListener
+	instance   *v1alpha2.IstioControlPlane
+	reconciler *helmreconciler.HelmReconciler
+}
 
 // NewIstioRenderingListener returns a new IstioRenderingListener, which is a composite that includes IstioStatusUpdater
 // and IstioChartCustomizerListener.
-func NewIstioRenderingListener(instance *v1alpha1.IstioControlPlane) *IstioRenderingListener {
+func NewIstioRenderingListener(instance *v1alpha2.IstioControlPlane) *IstioRenderingListener {
```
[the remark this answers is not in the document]
ostromart: Actually there are several interfaces that either should be better documented to explain why they are needed, or removed. I'd vote for the second option since there's only one implementation right now and it's not clear when we would need to be able to handle multiple dimensions of implementations in a dynamically pluggable way - and for sure there's a price to pay for developers contributing to the code to ramp up on the complexity. Some abstractions that could potentially removed imo:
- factories like these: [1](https://github.com/istio/operator/blob/1d29ca219595b827c071d1e2506ca4a49ac3b925/pkg/controller/istiocontrolplane/customizer.go#L26), [2](https://github.com/istio/operator/blob/1d29ca219595b827c071d1e2506ca4a49ac3b925/pkg/controller/istiocontrolplane/customizer.go#L26), [3](https://github.com/istio/operator/blob/1d29ca219595b827c071d1e2506ca4a49ac3b925/pkg/helmreconciler/customizers.go#L98) and others - using these and passing them around everywhere makes code harder to follow and I'm not sure what it's buying us. 
- [RenderingInput](https://github.com/istio/operator/blob/1d29ca219595b827c071d1e2506ca4a49ac3b925/pkg/helmreconciler/interfaces.go#L72) - realistically do we need an abstraction for these? We already have an input definition from the CRD and it should cover everything we need. 

@rcernich did a great job create something that would be broad enough to accommodate future possibilities at a time when our direction was still not very clear. Now, we know a lot better what the first product will look like and I think devs will make faster progress if we roll back some of the complexity. We can certainly put these abstractions back if the need arises.

pkg/apis/istio/v1alpha2/deepcopy.go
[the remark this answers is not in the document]
ostromart: Thanks Neeraj for the proto deepcopy tip, I'm making the change you suggested. With the auto-generation, the issue I ran into a while back is that eventually you hit k8s types that don't have DeepCopyInto implemented and then you have to somehow generate zz_deepcopy files for those (in the vendor tree). I could not figure out a clean way to do it but if you know how it would be great to have.

pkg/helmreconciler/reconciler.go
ostromart: i don't think so - we're just reporting status to a listener. but it raises a valid point - if reconcile fails, what does that mean for component status? it should probably stay in "reconciling". have left a TODO for that.

pkg/helmreconciler/reconciler.go
ostromart: Done.

pkg/helmreconciler/reconciler.go
ostromart: Done.

pkg/helmreconciler/reconciler.go
ostromart: it's because a dependency is a parent in the tree. but that's an implementation detail - i've removed it.

pkg/helmreconciler/reconciler.go
ostromart: Done.

pkg/helmreconciler/reconciler.go
ostromart: Done.

pkg/helmreconciler/reconciler.go
ostromart: Done.

pkg/helmreconciler/reconciler.go
ostromart: Done.

pkg/helmreconciler/rendering.go
ostromart: yes.

pkg/helmreconciler/rendering.go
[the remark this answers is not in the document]
ostromart: Done.

pkg/helmreconciler/rendering.go
```diff
@@ -27,68 +38,112 @@ import (
...
+
+	// Merge base and overlay.
+	mergedYAML, err := helm.OverlayYAML(baseYAML, overlayYAML)
+	if err != nil {
+		return nil, fmt.Errorf("could not overlay user config over base: %s", err)
+	}
+	return unmarshalAndValidateICPS(mergedYAML)
+}
+
+func unmarshalAndValidateICP(crYAML string) (*v1alpha2.IstioControlPlaneSpec, string, error) {
```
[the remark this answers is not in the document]
ostromart: Done.

pkg/helmreconciler/rendering.go
ostromart: not absolutely, but the first line of the YAML ends up on the preceding message line without it.

pkg/helmreconciler/rendering.go
ostromart: Done.

ostromart (commented):
Thanks for the detailed review, @sdake and @linsun, I really appreciate it.

pkg/controller/istiocontrolplane/listeners.go
richardwxn: is this change added or still TODO

richardwxn: is this change added or still TODO

pkg/apis/istio/v1alpha2/deepcopy.go
```diff
@@ -15,13 +15,22 @@
 package v1alpha2
 
 import (
-	runtime "k8s.io/apimachinery/pkg/runtime"
-	"reflect"
+	"github.com/prometheus/common/log"
```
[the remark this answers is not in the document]
ostromart: I think it got auto-inserted by Goland - removed.

linsun (approved):
/approve

PR #302, opened 2019-09-25
Update golden outputs

ostromart:
Some previous tests have not updated the goldens. They are passing because some resources like configmaps are still being ignored when diffing (https://github.com/istio/istio/issues/17363).


ostromart:
Goldens were updated in another PR, obsolete.

PR #303, opened 2019-09-27
Support --set profile=x

irisdingbj:
https://github.com/istio/istio/issues/17408

### DIFF
--- a/cmd/mesh/profile-dump.go
+++ b/cmd/mesh/profile-dump.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/ghodss/yaml"
 	"github.com/spf13/cobra"
+	
 	"istio.io/operator/pkg/apis/istio/v1alpha2"
 	"istio.io/operator/pkg/component/component"
 	"istio.io/operator/pkg/helm"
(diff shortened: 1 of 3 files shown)


cmd/mesh/profile-dump.go
```diff
@@ -88,6 +89,21 @@ func profileDump(args []string, rootArgs *rootArgs, pdArgs *profileDumpArgs, l *
 
 func genProfile(helmValues bool, inFilename, profile, setOverlayYAML, configPath string) (string, error) {
 	overlayYAML := ""
+	profiles := helm.ListBuiltinProfiles()
```
ostromart: This is not quite right as it only handles "profile" field. 
If you look at manifestApply, there's this line: 
	overlayFromSet, err := makeTreeFromSetList(maArgs.set)
This is where --set processing is done. If it's not working, something has been broken here.

irisdingbj: makeTreeFromSetList has no issue, it has successfully parsed the `profile` arg and put it into overlayFromSet.

The issue here is we always use `""` as the `profile` value and ignore the values from overlayFromset.  All other values from set will be respected but only `profile` will be ignored.

ostromart: Ok, I understand the problem now. I think it would be better to work with YAML rather than text below e.g. 
- unmarshal setOverlayYAML
- if profile is unset, set it to yamlTree["profile"]

irisdingbj: Thanks Martin and Fixed

ostromart (approved):

PR #304, opened 2019-09-27
manifest diff: add resource rename flag 

### LINKED ISSUE #16826: Resource rename flag for manifest diff command
Sometimes, resources in manifests are renamed, which causes them to appear as deleted/added. In this case it also becomes difficult to diff the two resources as they need to be renamed back manually. 
The request is to have a flag like --renamed which simply takes a list of values like
kind1:namespace1:name1->newkind1:newnamespace1:newname1, kind2:namespace2...
This says that kind1:namespace1:name1 is now called newkind1:newnamespace1:newname1 in the second manifest and they are the same resource. 
Ideally wildcards would also be supported e.g. 
Service:*:istio-pilot->Service:*:istio-control, *:*:istio-citadel->*:*:istio-ca

elfinhe:
Resolve: https://github.com/istio/istio/issues/16826
manifest diff: add resource rename flag to be able to diff a name-changed resource.

Example if you have `BUILD_WITH_CONTAINER=0 make mandiff` executed locally.

``` bash
mesh manifest diff --ignore ConfigMap::istio:data.values.yaml  \
--directory /tmp/istio-mandiff-out/helm-template/istio-sds \
/tmp/istio-mandiff-out/mesh-manifest/istio-sds \
--rename 'HorizontalPodAutoscaler:istio-system:istio-telemetry->HorizontalPodAutoscaler:istio-system:istio-telemetry-renamed'
```
Output
```
Differences of manifests are:

Object HorizontalPodAutoscaler:istio-system:istio-telemetry-renamed has diffs:

metadata:
  name: istio-telemetry -> istio-telemetry-renamed
```

### DIFF
--- a/pkg/object/objects.go
+++ b/pkg/object/objects.go
@@ -385,14 +385,35 @@ func ManifestDiff(a, b string, verbose bool) (string, error) {
 
 // ManifestDiffWithSelect checks the manifest differences with selected and ignored resources.
 // The selected filter will apply before the ignored filter.
-func ManifestDiffWithSelectAndIgnore(a, b, selectResources, ignoreResources string, verbose bool) (string, error) {
+func ManifestDiffWithRenameSelectIgnore(a, b, renameResources, selectResources, ignoreResources string, verbose bool) (string, error) {
+	rnm := getKeyValueMap(renameResources)
 	sm := getObjPathMap(selectResources)
 	im := getObjPathMap(ignoreResources)
-	aosm, err := filterResourceWithSelectAndIgnore(a, sm, im)
+
+	ao, err := ParseK8sObjectsFromYAMLManifest(a)
 	if err != nil {
 		return "", err
 	}
-	bosm, err := filterResourceWithSelectAndIgnore(b, sm, im)
+	aom := ao.ToMap()
+
+	bo, err := ParseK8sObjectsFromYAMLManifest(b)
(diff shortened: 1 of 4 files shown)


pkg/object/objects.go
```diff
@@ -385,28 +385,87 @@ func ManifestDiff(a, b string, verbose bool) (string, error) {
...
+	oom := make(map[string]*K8sObject)
+	for name, obj := range iom {
+		isRenamed := false
+		for fromPat, toPat := range rnm {
+			fromRe, err := buildResourceRegexp(strings.TrimSpace(fromPat))
+			if err != nil {
+				return nil, fmt.Errorf("error building the resource regexp: %v", err)
+			}
+			if fromRe.MatchString(name) {
+				fromList := strings.Split(name, ":")
```
morvencao: shouldn't this be `fromList := strings.Split(fromPat, ":")`?

elfinhe: The following log message must be misleading. 
Clarification: the fromPat is intended to only be used for matching. Once matching is done, we use the old name to fill up missing parts just in case that newPat a wildcard.

elfinhe: I updated the log messages and comments. also added tests to cover wildcard cases.

ostromart (approved):

PR #305, opened 2019-09-29
update versions map.

morvencao:


version/versions.yaml
```diff
@@ -1,4 +1,6 @@
 - operatorVersion: 1.3.0
   supportedIstioVersions: 1.3.0
   recommendedIstioVersions: 1.3.0
```
ostromart: Since we're updating this, could you please add an entry for 1.3.1 too?

version/versions.yaml
```diff
@@ -1,4 +1,6 @@
...
-
+- operatorVersion: 1.4.0
+  supportedIstioVersions: ">=1.3.0,<=1.4"
```
ostromart: We haven't yet finalized whether we'll have forward compatibility in a minor version, so let's keep this to 1.4.0 for now.

version/versions.yaml
morvencao: done

version/versions.yaml
morvencao: removed 1.4.0 entry.

ostromart (approved):

PR #307, opened 2019-09-29
add controller target in makefile.

morvencao:


Makefile.core.mk
```diff
@@ -50,11 +54,24 @@ generate: generate-values generate-types vfsgen
...
 mesh: vfsgen
 	go build -o ${GOBIN}/mesh ./cmd/mesh.go
 
+controller: vfsgen
+	go build -o ${CONTROLLER_BUILD}/_output/bin/istio-operator ./cmd/manager
+
+controller-image: build-controller-image push-controller-image
+
+build-controller-image: controller
+	docker build -t istionightly/operator:latest -f ${CONTROLLER_BUILD}/Dockerfile .
```
sdake: please do not tag images as `:latest`.  You can use `:$(VERSION}`.

Makefile.core.mk
```diff
@@ -50,11 +54,24 @@ generate: generate-values generate-types vfsgen
...
+	docker tag istionightly/operator:latest istionightly/operator:${VERSION}
+
+push-controller-image:
+	docker push istionightly/operator:latest
```
sdake: please definitely do not push images as `:latest`.

Makefile.core.mk
sdake: what copies the Dockerfile to ${CONTROLLER_BUILD}?

morvencao: switched to `:$(VERSION}`

Makefile.core.mk
morvencao: Removed latest, only push `:${VERSION}`

Makefile.core.mk
morvencao: The `Dockerfile` is put into `${CONTROLLER_BUILD}` by default.

Makefile.core.mk
```diff
@@ -50,11 +54,22 @@ generate: generate-values generate-types vfsgen
...
+controller: vfsgen
+	go build -o ${CONTROLLER_BUILD}/_output/bin/istio-operator ./cmd/manager
+
+controller-image: build-controller-image push-controller-image
+
+build-controller-image: controller
+	docker build -t istionightly/operator:${VERSION} -f ${CONTROLLER_BUILD}/Dockerfile .
+
+push-controller-image:
+	docker push istionightly/operator:${VERSION}
```
howardjohn: please don't use istionightly

morvencao: Should we use `docker.io/istio` instead?

howardjohn:
When and why do you want the image pushed? Every commit? Every Istio
release?

Is it pushed for users to use it for developers/ci?

morvencao:
@howardjohn I think we need to cover all these cases:
- publishing image for each release
- pushing image to docker hub on each commit for CI
- also developer may override docker hub with their own for development purpose

howardjohn:
> publishing image for each release

All releases are built from istio/istio. You just need to add a target `docker.save` that outputs all docker files as `.tar` to go/out/linux_amd64_release/docker. Then add `RunMake(... operator docker.save ..)` to https://github.com/istio/release-builder/blob/5ee70657d654ab62f443ba5c15cc96a872fb51a5/pkg/build/docker.go#L34. I can do add it to the release step if you add the docker.save target

> pushing image to docker hub on each commit for CI

sgtm, we should use the `gcr.io/istio-release` hub though. We are going to do that for istio/istio as well. One thing we could do is have it published on each commit of istio/istio instead of on each commit of istio/operator. That way images will be tagged with `:istio-sha` instead of `:operator-sha`. Just depends how coupled to istio/istio you want the operator releases to be  I guess.

> also developer may override docker hub with their own for development purpose

It doesn't look like they can in the current state. Other istio repos control this with HUB and TAG variables, should we reuse that pattern here?

Makefile.core.mk
ostromart: I added docker.io/istio as a default.

ostromart:
Leaving the release changes aside for the moment, I updated this CR to support developers being able to create and push a docker image for operator. Only runs from local build environment atm.
PTAL @howardjohn

Makefile.core.mk
```diff
@@ -63,6 +64,21 @@ mesh: vfsgen
...
+controller: vfsgen
+	go build -o $(GOPATH)/bin/istio-operator ./cmd/manager
+
+docker: controller
+	mkdir -p build/out
+	cp $(GOPATH)/bin/istio-operator build/out/.
+	docker build -t $(HUB)/operator:$(TAG) -f build/Dockerfile .
+
+docker.push:
+	docker push $(HUB)/operator:$(TAG)
```
howardjohn: HUB is not set in the makefile, so this may break if a user doesn't have it set? Maybe we should either set a default or fail if it is unset. We should probably not set it to `istio` to avoid accidental pushes to the official dockerhub

deploy/operator.yaml
```diff
@@ -15,8 +15,7 @@ spec:
       serviceAccountName: istio-operator
       containers:
         - name: istio-operator
-          # Replace this with the built image name
-          image: REPLACE_IMAGE
+          image: docker.io/istio/controller
```
howardjohn: This image won't exist?

howardjohn (commented):
lgtm for local use

sdake (approved):
will address followups in a new pr

Makefile.core.mk
ostromart: It's intended to fail if HUB is unset, but it would be nice to output an error. This is also assuming there should not be a default HUB to avoid accidental pushes. @sdake is going to clean all this up to work properly in the container and address this.

PR #310, opened 2019-10-01
WIP: Operator gate test

sdake:


sdake:
operator gate is running fine - minus one non-voting check.

PR #311, opened 2019-10-01
Update values schema for sidecarInjectorWebhook

elfinhe:


pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -1333,6 +1333,14 @@ message SidecarInjectorConfig {
   google.protobuf.BoolValue selfSigned = 17;
 
   string injectLabel = 18;
+
```
ostromart: v1alpha1 is no longer used.

elfinhe:
Reopening per discussion offline.

elfinhe:
Need help: 
I used 'make gen' to generate the values_types proto, but it will use github.com/gogo/protobuf/protobuf/google/protobuf instead of github.com/gogo/protobuf/types.
Also it will replace 'map[string]interface{}' with self defined type 'TypeMapStringInterface'.

Can someone teach me how to generate the .proto.go when change values_types.proto?
@ostromart, @morvencao

PR #315, opened 2019-10-02
Remove v1alpha1 based values validation.

elfinhe:
The current values validation is based on the model from v1alpha1. Since v1alpha1 is not updated, let's remove this to not block adding new fields into config (e.g., profiles/default.yaml).

### DIFF
--- a/pkg/validate/validate_values.go
+++ b/pkg/validate/validate_values.go
@@ -15,9 +15,6 @@
 package validate
 
 import (
-	"github.com/ghodss/yaml"
-
-	"istio.io/operator/pkg/apis/istio/v1alpha1"
 	"istio.io/operator/pkg/util"
 )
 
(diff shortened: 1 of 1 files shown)


ostromart:
The latest values proto is still in v1alpha1 so we don't want to remove it at this time. We would migrate it to v1alpha2 but it's probably not worth the effort since anyway v1alpha2 will merge to MeshConfig in 1.4.

elfinhe:
> The latest values proto is still in v1alpha1 so we don't want to remove it at this time. We would migrate it to v1alpha2 but it's probably not worth the effort since anyway v1alpha2 will merge to MeshConfig in 1.4.

Those three are not in the latest v1alpha1:
    sidecarInjectorWebhook:
      rollingMaxSurge: 100%
      rollingMaxUnavailable: 25%
      tolerations: []

like I suggested yesterday in https://github.com/istio/operator/pull/311/files

We should either remove the alpha1 validation or add the those 3 fields into the model.

PR #316, opened 2019-10-02
FIx code generation logic.

geeknoid:


geeknoid:
Needs https://github.com/istio/test-infra/pull/1880 to pass the gencheck job.

geeknoid:
@ostromart Oh and this new gencheck job is highlighting the fact that the currently checked in generated files are out of date and need to be regenerated.

Makefile.core.mk
```diff
@@ -37,7 +37,7 @@ fmt: format-go
 	go mod tidy
 
 update-charts: installer.sha
-	@scripts/run_update_charts.sh `cat installer.sha`
+	@scripts/run_update_charts.sh `cat installeitr.sha`
```
ostromart: typo

ostromart:
The generated files look to be up to date - running make vfsgen on master does not update assets.

Makefile.core.mk
geeknoid: Whoops. Fixed.

geeknoid:
When I do "make gen", I end up with:

        modified:   pkg/apis/istio/v1alpha1/v1alpha1.pb.html
        modified:   pkg/apis/istio/v1alpha1/values_types.pb.go
        modified:   pkg/apis/istio/v1alpha2/v1alpha2.pb.html
        modified:   pkg/vfs/assets.gen.go

geeknoid:
I added the changed files to the PR.

geeknoid:
@ostromart Note that with the newly generated files included in the PR, a bunch of jobs are failing now...

ostromart:
Ah ok, it that's because values_types.proto was modified recently and gen was not run as part of that. Let me take care of it now. I wonder if there's some presubmit check we could add to enforce it?

geeknoid:
@ostromart That's specifically what my PR is trying to enforce :-) We will have a gate to prevent a PR from going in without updating generated files.

ostromart (approved):

PR #319, opened 2019-10-04
Fix yaml blocks in comments

ostromart:
https://github.com/istio/istio.io/issues/5091


geeknoid (commented):
Removing the indent will ruin the bullet lists I think. Each yaml block will terminate the bullet list. 

You need to run "make gen" in order to regenerate the .pb.html file. You can look at the delta in the HTML file to confirm or deny my assertion from above.

geeknoid:
See https://github.com/istio/operator/pull/320 for something that works better.

PR #331, opened 2019-10-09
Add a docker.save target

sdake:


Makefile.core.mk
```diff
@@ -78,6 +78,10 @@ docker: controller
 docker.push:
 	docker push $(HUB)/operator:$(TAG)
 
+docker.save:
+	mkdir -p /work/out/linux_amd64_release/docker
```
howardjohn: this should probably be something like gopath/out?

Makefile.core.mk
```diff
@@ -78,6 +78,10 @@ docker: controller
...
+	docker save $(HUB)/operator:$(TAG) -o /work/out/linux_amd64_release/docker/operator.tar.gz
```
howardjohn:
```suggestion
	docker save $(HUB)/operator:$(TAG) -o /work/out/linux_amd64_release/docker/operator.tar
```

Makefile.core.mk
sdake: /work is where files are written within the container.  $GOPATH=/home/go in the docker container.  Not clear how to unify the ci system with the local build system...

howardjohn: I don't think we should be hardcoding paths just because it works in the build container. istio/istio had some method to determine where out is, as does CNI, we should use the same method

sdake: agreed re hardcoding.  Something is missing from our infrastructure.  CNI does not output to the local disk - istio is not on the containers plan.  It doesn't appear to output to the local disk either, even with its logic to detect the output location.

geeknoid: OK, I screwed the output targets up. I started to put some stuff that would only work from the container. I'll fix these shortly. It'll require a new container and top-level makefile I think.

howardjohn: istio/istio works like:
```
GO_TOP := $(shell echo ${GOPATH} | cut -d ':' -f1)
export OUT_DIR=$(GO_TOP)/out
export ISTIO_OUT:=$(OUT_DIR)/$(GOOS)_$(GOARCH)/$(BUILDTYPE_DIR)
```

sdake: @howardjohn I tried that - it doesn't work in the operator implementation.  I doubt it works in istio/istio either...

sdake:
@howardjohn I don't recall why we need this target.  I believe either you or @geeknoid had requested this work and it may be part of the release workflow.  If this work is needed, please let me know and I'll finish it.

Cheers
-steve

howardjohn:
If you want this to be published as part of the istio release we need this target

sdake:
> If you want this to be published as part of the istio release we need this target

@howardjohn PTAL.  tx.

Cheers
-steve

sdake:
Looking at CNI (I was checking to determine if it had a docker.save target) I see:

```
mkdir -p /work/out/linux_amd64/release/docker
time (docker save -o /work/out/linux_amd64/release/docker/install-cni.tar istio/install-cni:5f438d5ab98dd1f97ff7c0f5e437e044c314a9c3 && gzip /work/out/linux_amd64/release/docker/install-cni.tar)
```

I think we should have consistency in our directory locations.  I am unclear if we want things in 

GOOS_GOARCH/release/docker or
GOOS_GOARCH_release/docker.

Can you clarify.

Thanks
-steve

howardjohn:
we want GOOS_GOARCH/release/docker, this ksnhow istio/istio does it as well

howardjohn (approved):

PR #338, opened 2019-10-09
couple of bug fixes for controller.

morvencao:
Now, the controller can reconcile basic IstioControlPlane object and apply istio manifests in cluster.


pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```diff
@@ -170,6 +170,8 @@ message IstioControlPlane {
     IstioControlPlaneSpec spec = 3;
     // Status reports the status of the Istio control plane.
     InstallStatus status = 4;
+    string kind = 5;
+    string apiVersion = 6;
```
morvencao: Not sure if this change is acceptable, but current the `kind` and `apiVersion` for `IstioControlPlane` can not be unmarshalled correctly, because `TypeMeta` is an inline field for kubernetes object, proto can't handle this properly. https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L41

ostromart: If this gives us enough to make the controller work then it's completely ok. We had been thinking we might need some kluge to add this field to the generated struct, but if this is sufficient then it's actually a cleaner solution. 
One thing I will do is change the field ordering to follow k8s conventions.

ostromart (approved):

PR #340, opened 2019-10-09
Operator/common files

sdake:


howardjohn (approved):

richardwxn:
@sdake maybe I didn't setup things correctly, but it seems like docker targets only work with local builds still. Is your change supposed to make it works for container build as well?

PR #342, opened 2019-10-09
Update to common-files

sdake:


sdake:
waiting to propogate local build container to test-infra.

ostromart (commented):
Could you please check that make vfsgen works for linux after this change? I worry because we had problems with GOBIN before.

sdake:
> Could you please check that make vfsgen works for linux after this change? I worry because we had problems with GOBIN before.

I tested all of the make targets (with BUILD_WITH_CONTAINER=1) prior to submission.  The gate checks BUILD_WITH_CONTAINER=0 for the most part.  I think this is GTG.  I am simply waiting on a build of the docker container that has a new prow entrypoint to push the operator container.  See PR: https://github.com/istio/tools/pull/428 https://github.com/istio/tools/pull/430.  Once these PRs regenerate a new image, I will install it in the checks.  Then this PR can merge (I will remove the hold I set) - and we should have published docker images on gcr.io (hopefully).

Cheers
-steve

howardjohn (approved):

PR #345, opened 2019-10-10
Set Owner info for Istio Objects

irisdingbj:


pkg/controller/istiocontrolplane/pruningdetails.go
```diff
@@ -104,7 +104,11 @@ var (
 
 // NewPruningDetails creates a new PruningDetails object specific to the instance.
 func NewIstioPruningDetails(instance *v1alpha2.IstioControlPlane) helmreconciler.PruningDetails {
-	gvk := instance.GetObjectKind().GroupVersionKind()
+	gvk := schema.GroupVersionKind{
+		Version: "v1alpha2",
+		Group:   "install.istio.io",
+		Kind:    "IstioControlPlane",
+	}
```
ostromart: Please define these as consts in a common package somewhere. I think Group is already defined...

pkg/helmreconciler/listeners.go
```diff
@@ -444,9 +445,14 @@ func NewOwnerReferenceDecorator(instance runtime.Object) (RenderingListener, err
 	if err != nil {
 		return nil, err
 	}
+	gvk := schema.GroupVersionKind{
+		Version: "v1alpha2",
+		Group:   "install.istio.io",
+		Kind:    "IstioControlPlane",
+	}
```
ostromart: Since gvk seems used in a couple of places, it probably makes sense to define the whole thing as an exported const in a common package e.g. IstioOperatorGVK?

irisdingbj: put it into util/common.go

ostromart (approved):

PR #346, opened 2019-10-10
fix status updater.

morvencao:


ostromart (approved):
Thanks Morven. fyi, I created an issue to track the workaround since it looks like the fix is imminent:
https://github.com/istio/istio/issues/17761

morvencao:
@ostromart Thanks

PR #350, opened 2019-10-10
Trim dependencies on kubernetes

howardjohn:
Part of https://github.com/istio/istio/issues/17245

The current set of dependencies makes it impossible to update our version of client-go in istio/istio, which is needed to support users on Kubernetes 1.16. This is due to inappropriate dependencies on k8s.io/kubernetes from some dependencies. This is part 1 of dropping k8s.io/kubernetes entirely.

Before this change:
```
$ go mod graph | wc -l
2858
```

After this change:
```
$ go mod graph | wc -l
1656
```

I have tested this change manually by deploying, and tested the leader election works by scaling multiple replicas


deploy/operator.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: istio-operator
       containers:
         - name: istio-operator
-          image: gcr.io/istio-testing/operator:master-latest-daily
+          image: gcr.io/howardjohn-istio/operator:master-latest-daily
```
ayj: intentional?

cmd/manager/server.go
```diff
@@ -123,12 +118,6 @@ func run() {
 		log.Fatalf("Could not add all controllers to operator manager: %v", err)
 	}
 
-	// Create Service object to expose the metrics port.
-	_, err = metrics.CreateMetricsService(ctx, cfg, metricsServicePort)
-	if err != nil {
-		log.Errorf("Could not create a service to expose the metrics port: %v", err.Error())
-	}
-
```
ayj: Don't we still need to expose the metrics port?

howardjohn: This is just creating a Service I think. We can just do the same in `deploy/`?

ayj: Yes, I think so.

howardjohn: Adding it to deploy

deploy/operator.yaml
howardjohn: Nope, was testing to make sure this worked, good catch. reverted.

cmd/manager/server.go
```diff
@@ -42,11 +36,8 @@ import (
 
 // Change below variables to serve metrics on different host or port.
 var (
-	metricsHost              = "0.0.0.0"
-	metricsPort        int32 = 8383
-	metricsServicePort       = []v1.ServicePort{
-		{Port: metricsPort, Name: metrics.OperatorPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: metricsPort}},
-	}
+	metricsHost       = "0.0.0.0"
```
ayj: s/var/const 

Also might be worth updating the comment to say the k8s service needs to be updated if this is changed.

howardjohn: :+1: done

ayj (approved):

PR #352, opened 2019-10-10
Run kubectl apply -k deploy to deploy operator

sdake:
This seems to work and generate a running operator, however, the operator
doesn't appear to do any work.


howardjohn (approved):

deploy/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+namespace: istio-operator
+resources:
+- crds/istio_v1alpha2_istiocontrolplane_crd.yaml
```
kramvan1: The CR would also need this `istio-operator` namespace for it to work.

PR #353, opened 2019-10-10
Add force flag to bypass validation

ostromart:
Note: there's a FIXME for a test that _should_ fail but currently doesn't because values validation still has AllowUnknownFields set.


pkg/vfs/assets.gen.go
```diff
@@ -32733,6 +33741,74 @@ func profilesSdsYaml() (*asset, error) {
...
+spec:
+  security:
+    components:
+      nodeAgent:
+        enabled: true
+<<<<<<< HEAD
+<<<<<<< HEAD
+
+=======
+>>>>>>> 0584d5d7... add back the mandiff test. (#223)
```
howardjohn: this doesn't seem right?

howardjohn (approved):
lgtm other than vfs gen, not sure if that is intended? put a hold on this

pkg/vfs/assets.gen.go
ostromart: no, it doesn't seem right - love the understatement :-)
this was just additional cruft that appeared because of some .orig files being present in the source dir, these are gitignored. thanks.

PR #354, opened 2019-10-10
Set the namesapce for the istiocontrolplane CR

sdake:


richardwxn (approved):

deploy/crds/istio_v1alpha2_istiocontrolplane_cr.yaml
```diff
@@ -1,6 +1,7 @@
 apiVersion: install.istio.io/v1alpha2
 kind: IstioControlPlane
 metadata:
+  namespace: istio-operator
```
linsun: does the namespace indicate the control plane is deployed to Istio-operator namespace?

morvencao: Yes, currently in-cluster controller will be also installed into the same namespace.

morvencao (approved):

PR #355, opened 2019-10-10
Add --prune back to manifest apply kubectl command

### LINKED ISSUE #17236: Add --prune back to manifest apply kubectl command
Currently this is removed because the manifest is applied in parallel per component, and --prune requires an aggregate of all resources for the entire manifest of all the components together. 
One possible solution is to continue to do the apply normally, in parallel as we do right now, but follow it with a cleanup step that does a kubectl apply on the whole concatenated manifest at the end with --prune. 
This will cause apply to take longer but not prohibitively so because the prior apply would be the one doing most of the work.

elfinhe:
Resolve: https://github.com/istio/istio/issues/17236

### DIFF
--- a/pkg/manifest/installer.go
+++ b/pkg/manifest/installer.go
@@ -285,8 +285,7 @@ func applyManifest(componentName name.ComponentName, manifestStr string, version
 
 	appliedObjects = append(appliedObjects, objects...)
 
-	// TODO; add "--prune" back (istio/istio#17236)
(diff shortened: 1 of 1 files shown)


ostromart (approved):

ostromart:
@elfinhe looking at the change, it looks exactly right, but could you please check that prune works as expected before merging?

elfinhe:
> @elfinhe looking at the change, it looks exactly right, but could you please check that prune works as expected before merging?

Sorry I forgot to mention that I've tested using it do both fresh install and  an upgrade. I will double check before merging.

PR #358, opened 2019-10-11
Fix Concurrent Error

irisdingbj:
Fix Concurrent error for updating status

### DIFF
--- a/pkg/helmreconciler/reconciler.go
+++ b/pkg/helmreconciler/reconciler.go
@@ -137,19 +137,22 @@ func (h *HelmReconciler) processRecursive(manifests ChartManifestsMap) *v1alpha2
 			if _, ok := out.Status[c]; !ok {
 				out.Status[c] = &v1alpha2.InstallStatus_VersionStatus{}
 			}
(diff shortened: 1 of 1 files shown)


pkg/helmreconciler/reconciler.go
```diff
@@ -132,16 +133,21 @@ func (h *HelmReconciler) processRecursive(manifests ChartManifestsMap) *v1alpha2
 				<-s
 				log.Infof("Dependency for %s has completed, proceeding.", c)
 			}
-
+			mu.Lock()
 			if _, ok := out.Status[c]; !ok {
 				out.Status[c] = &v1alpha2.InstallStatus_VersionStatus{}
 			}
 			out.Status[c].Status = v1alpha2.InstallStatus_NONE
+			mu.Unlock()
```
ostromart: Nice catch! You can reduce the amount of locking by keeping the state in a local variable and write it back to the Status map in just one place at the end.

irisdingbj: Thanks for the suggestion. Fixed!

morvencao (approved):

PR #360, opened 2019-10-11
batch of fixes for the controller.

morvencao:
- remove unused resource watch
- add more debug info

### DIFF
--- a/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
+++ b/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
@@ -70,15 +70,6 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	// TODO(user): Modify this to be the types you create that are owned by the primary resource
-	// Watch for changes to secondary resource Pods and requeue the owner IstioControlPlane
-	//err = c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{
-	//        IsController: true,
-	//        OwnerType:    &v1alpha2.IstioControlPlane{},
-	//})
-	//if err != nil {
-	//        return err
-	//}
 	log.Info("Controller added")
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/helmreconciler/listeners.go
```diff
@@ -258,25 +258,41 @@ func (l *LoggingRenderingListener) BeginResource(obj runtime.Object) (runtime.Ob
 
 // ResourceCreated logs the event
 func (l *LoggingRenderingListener) ResourceCreated(created runtime.Object) error {
-	log.Info("new resource created")
+	accessor := meta.NewAccessor()
+	kind, _ := accessor.Kind(created)
+	name, _ := accessor.Name(created)
+	version, _ := accessor.APIVersion(created)
+	log.Infof("resource created (%s, %s, %s)", version, kind, name)
```
ayj: You can get the full group/version/kind from the runtime object directly, i.e. ```created.GetObjectKind().GroupVersionKind()```

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
```diff
@@ -73,13 +72,13 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
-	if err != nil {
-		return err
-	}
+	//err = c.Watch(&source.Kind{Type: &corev1.Pod{}}, &handler.EnqueueRequestForOwner{
+	//        IsController: true,
+	//        OwnerType:    &v1alpha2.IstioControlPlane{},
+	//})
+	//if err != nil {
+	//        return err
+	//}
```
ostromart: I think you can just remove this. I don't see a use case for us wanting to watch Pods.

morvencao: removed this section.

pkg/helmreconciler/listeners.go
morvencao: Thanks, done.

ostromart (approved):

PR #362, opened 2019-10-11
minor updates to README

rvennam:


README.md
```diff
@@ -94,14 +94,13 @@ The `mesh` command supports the following flags:
 
 #### Basic default manifest
 
-The following command generates a manifest with the compiled in default profile and charts:
+The following command generates a manifest with the compiled-in profiles and charts:
```
ostromart: compiled-in is ok, but please change back to default profile, because that's what it will use (data/profiles/default.yaml).

README.md
```diff
@@ -94,14 +94,13 @@ The `mesh` command supports the following flags:
...
 
 ```bash
 mesh manifest generate
 ```
 
-You can see these sources for the compiled in profiles in the repo under `data/profiles`. `Charts/profiles` will be released separately and the
-by default the mesh command will point to a version of the released charts.
+You can see these sources for the compiled-in profiles in the repo under `data/profiles`. `Charts/profiles` will be released separately and the mesh command will point to a version of the released charts by default.
```
ostromart: You can see the sources for the compiled-in profiles and charts in the repo under data/. These profiles and charts are also included in the Istio release tar.

sdake (approved):

PR #368, opened 2019-10-11
Fix rbac issue

richardwxn:
add namespace to work

### DIFF
--- a/deploy/clusterrole.yaml
+++ b/deploy/clusterrole.yaml
@@ -59,6 +59,7 @@ rules:
   - configmaps
   - secrets
   - namespaces
+  - poddisruptionbudgets
   verbs:
   - '*'
 - apiGroups:
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: istio-operator
+  namespace: istio-operator
 spec:
   replicas: 1
   selector:
--- a/deploy/service_account.yaml
+++ b/deploy/service_account.yaml
@@ -2,3 +2,4 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: istio-operator
+  namespace: istio-operator


deploy/service_account.yaml
```diff
@@ -2,3 +2,4 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: istio-operator
+  namespace: istio-operator
```
sdake: this is redundant.  kubectl `-k` flag solves this with the kustomize.yaml code.

deploy/operator.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: istio-operator
+  namespace: istio-operator
```
sdake: also redundant.

deploy/clusterrole.yaml
```diff
@@ -59,6 +59,7 @@ rules:
   - configmaps
   - secrets
   - namespaces
+  - poddisruptionbudgets
```
sdake: this goes in `policy` not here.  This is already covered as well in the existing clusterrole.yaml:

```
- apiGroups:
  - policy
  resources:
  - poddisruptionbudgets
  verbs:
  - '*'
```

deploy/service_account.yaml
ostromart: Let's keep it since we will have an operator init command soon.

sdake: I am not opposed to keeping the redundant and parts but I do want ppl to be able to deploy directly from the internet with just kubectl and no istioctl

sdake: I am not opposed to keeping the redundant and parts but I do want ppl to be able to deploy directly from the internet with just kubectl and no istioctl

deploy/clusterrole.yaml
sdake: Please remove this from the or thx

sdake: Please remove this from the or thx

sdake:
The namespaces are not redundant.  Please add the istio-operator namespace to all objects in the repo.  See PR https://github.com/istio/istio.io/pull/5142 for why this is necessary.

Cheers
-steve

deploy/clusterrole.yaml
richardwxn: ok I didn't see it is added in policy group already. But somehow I am still seeing permission issue for pdb when testing.

sdake: can you paste the error?  I would like to see if in some way i fubared this section.. :)

sdake:
@richardwxn service.yaml also in need of a namespace.

deploy/operator.yaml
sdake: ignore comment.

deploy/service_account.yaml
sdake: ignore my comment.

sdake:
subsumed by: https://github.com/istio/operator/pull/374

sdake:
@richardwxn I can handle expanding RBAC for the operator here: https://github.com/istio/istio/issues/17841.  If you still suffer from a PDB issue, please paste the log message here or message to me on slack and I'll take a look.  If your satisfied with https://github.com/istio/operator/pull/374 - please close this issue and leave a comment.

Cheers
-steve

richardwxn:
@sdake thanks for the clarification! I am closing this PR now as yours was merged already. I would send the error log to you on slack tmr.

PR #369, opened 2019-10-11
Add readme for testing controller

richardwxn:


pkg/controller/README.md
```diff
@@ -0,0 +1,17 @@
+### Controller local test
+#### Run inside the cluster
+1.run make docker.all to push image with your HUB and TAG
+
+2.update deploy/operator.yaml to point to your image
+
+3.run kubectl apply --recursive deploy/
```
sdake: this is wrong.  instead use `kubectl apply -k deploy/` followed by an application of the CRD file.

pkg/controller/README.md
```diff
@@ -0,0 +1,17 @@
+### Controller local test
+#### Run inside the cluster
+1.run make docker.all to push image with your HUB and TAG
+
+2.update deploy/operator.yaml to point to your image
```
sdake: md formatting requires a 1 here.

pkg/controller/README.md
sdake: md formatting requires a 1 here.

pkg/controller/README.md
```diff
@@ -0,0 +1,17 @@
+### Controller local test
```
sdake: should be 1 or 2 `#` not 3.

pkg/controller/README.md
```diff
@@ -0,0 +1,17 @@
...
+#### Run inside the cluster
```
sdake: should be 2 or 3 `#` not 4.

pkg/controller/README.md
```diff
@@ -0,0 +1,17 @@
...
+1.run make docker.all to push image with your HUB and TAG
+
+2.update deploy/operator.yaml to point to your image
+
+3.run kubectl apply --recursive deploy/
+
+#### Run outside the cluster
```
sdake: should be 2 or 3 `# not 4.

pkg/controller/README.md
```diff
@@ -0,0 +1,17 @@
+### Controller local test
+#### Run inside the cluster
+1.run make docker.all to push image with your HUB and TAG
```
linsun: can we add a step on specify HUB and TAG first?

linsun:
Thanks for adding this @richardwxn

pkg/controller/README.md
linsun: also whether user needs to export BUILD_WITH_CONTAINER

pkg/controller/README.md
richardwxn: done

pkg/controller/README.md
richardwxn: done

pkg/controller/README.md
richardwxn: done

pkg/controller/README.md
richardwxn: ok, done

pkg/controller/README.md
richardwxn: done

pkg/controller/README.md
sdake: BUILD_WITH_CONTAINER is default.  I am not sure HUB nor TAG are passed into the make core file.  I will investigate this and fix it if it is busted.

richardwxn: @sdake yes actually I did not use BUILD_WITH_CONTAINER for this step, but looks like you already added passing HUB and TAG in common files. 

Also another reason I failed to run make docker.all with container was some docker permission issue, but other make target works fine.

ostromart:
@sdake and @linsun could we get this merged? We have a volunteer tester (@nrjpoddar ) who will try out these docs.

ostromart:
Approving so @nrjpoddar can try it, please open an issue if anything still needs updating.

ostromart (approved):

PR #370, opened 2019-10-12
Add anonymous synthetic types for Go struct generation

ostromart:
This creates synthetic types for where we need anonymous members in generated Go structs.


geeknoid (approved):
The gencheck job is still not happy for some reason. Do you need one more `make clean gen` to update all the files?

pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto
```diff
@@ -166,6 +166,11 @@ package v1alpha2;
 
 // IstioControlPlane is a CustomResourceDefinition (CRD) describing an Istio control plane.
 message IstioControlPlane {
+    // ObjectMeta is v11.ObjectMeta.
+    TypeObjectMeta object_meta = 1;
```
morvencao: Do we need to set `json_name` option for `ObjectMeta`? because its json key is `metadata` instead of `objectMeta`.

ostromart: yes, i think that also explains the problem you're seeing with unmarshaling the finalizer field.

ostromart:
$ make clean gen 
-> modified:   pkg/apis/istio/v1alpha1/values_types.pb.go
(this is due to fixup_structs not running and running the substitution on gogo proto include)
$ touch pkg/apis/istio/v1alpha1/values_types.proto
$ make gen
-> modified file disappears i.e. substitution runs correctly.

Any ideas what could be causing this @geeknoid ?

geeknoid:
@ostromart That's strange. Go ahead and ignore the job failure (it's not required) for now, and once this change goes in I'll try to figure out what's going wrong. It's probably something silly.

rcernich:
Hey guys, it's probably a bit late for this comment, but why are we using .proto files instead of k8s/code-generator go-to-protobuf?  I think using that would solve a lot of problems, as the type would be defined in go, as it's used by the controller, while making protobuf available for use with the apiserver.

geeknoid:
@rcernich We have infrastructure built around using proto as source of truth (docs, validation, schema generation, etc). I don't want operator to be an outlier as compared to the rest of Istio.

rcernich:
@geeknoid, understood.  However, these types are specific to the operator, which only runs in k8s, thus not the same.  I think it's fine to have what is essentially the 'spec' portion in istio/api, but have the wrapping k8s type fully declared in this project, or at least in istio/client-go.  That said, the type used by the operator will have a status field, which is currently not supported with the k8s api generation toolset (i.e. kubetype-gen).  I think we're being pedantic forcing the k8s type into istio/api.

ostromart:
There's an ongoing discussion [elsewhere](https://github.com/istio/operator/pull/382) about ongoing problems and history with using proto as a source of truth. 
With considerable effort, I think we've managed to make it work. It's certainly not a perfect solution. I propose revisiting this discussion after 1.4 - this is a very deep and involved topic with lots of surprises and caveats where the rubber hits the road. 
@sdake @rcernich could one of you guys please approve this for 1.4? We should create an issue to tackle the bigger issue of source of truth, use of jsonpb, gogo etc. and fix this or at least agree on the direction at the start of 1.5.

rcernich:
I'll defer to @sdake, as I think he has a PR dealing with removing gogo libraries.  That said, I'm okay with this.  If it's not going to disrupt that effort, let me know and I'll approve this.

sdake (approved):
This looks ok for now.   We need to take a hard look at parsing protos after generation in the future.  Feels wrong :)

PR #371, opened 2019-10-12
Fix DeadLock problem for reconciler

irisdingbj:


morvencao (commented):
how does the dead lock problem happen? Would you provide more details?
I thought `componentDependencies` and `dependencyWaitCh` are inited once and we don't need recompute them since they are static data for now.

irisdingbj:
> how does the dead lock problem happen? Would you provide more details?
> I thought `componentDependencies` and `dependencyWaitCh` are inited once and we don't need recompute them since they are static data for now.

```	// Signal all the components that depend on us.
			for _, ch := range deps[cn] {
				log.Infof("Unblocking dependency %s.", ch)
				dch[ch] <- struct{}{}
			}```  This is reason why deadlock happens as we will read out from channels for those existed only components.

morvencao (approved):

PR #372, opened 2019-10-12
Prune resources when IstioControlPlane changes

irisdingbj:
https://github.com/istio/istio/issues/17825


morvencao:
@irisdingbj The resources can be pruned when I switch profile with latest master code;
The controller will prune resources with different generation annotation, the current logic is pretty tricky, but it is working in my env. 

See my comments for the original issue: https://github.com/istio/istio/issues/17825#issuecomment-541320696

pkg/helmreconciler/pruning.go
```diff
@@ -28,15 +30,18 @@ import (
...
 // Prune removes any resources not specified in manifests generated by HelmReconciler h. If all is set to true, this
 // function prunes all resources.
-func (h *HelmReconciler) Prune(all bool) error {
+func (h *HelmReconciler) Prune(all bool, objMap map[string]bool) error {
 	allErrors := []error{}
 	namespacedResources, nonNamespacedResources := h.customizer.PruningDetails().GetResourceTypes()
 	targetNamespace := h.customizer.Input().GetTargetNamespace()
-	err := h.PruneResources(namespacedResources, all, targetNamespace)
+	if targetNamespace == "" {
+		targetNamespace, _ = os.LookupEnv("WATCH_NAMESPACE")
```
ostromart: I think this namespace is for the operator itself rather than Istio.

irisdingbj:
> @irisdingbj The resources can be pruned in my env when I switch profile with latest master code;
> The controller will prune resources with different generation annotation, the current logic is pretty tricky, but it is working in my env.
> 
> See my comments for the original issue: [istio/istio#17825 (comment)](https://github.com/istio/istio/issues/17825#issuecomment-541320696)

Tried using latest master code and this works.

PR #373, opened 2019-10-12
add leader election namespace.

morvencao:
Add dedicated namespace for leader election namespace(default value is the namespace in which controller pod is running);
Because `WATCH_NAMESPACE` can be `""`, which means watching all namespaces for IstioControlPlane CR, while `LEADER_ELECTION_NAMESPACE` can not be `""`.


howardjohn (approved):
Nice

PR #375, opened 2019-10-13
Print error messages to stderr.

elfinhe:
Print error messages to stderr to support CLI usage like:
`istioctl manifest generate --log-stderr > file.yaml `

Resolve: https://github.com/istio/istio/issues/17743

### DIFF
--- a/cmd/mesh/shared.go
+++ b/cmd/mesh/shared.go
@@ -67,11 +67,12 @@ func (l *logger) logAndPrint(v ...interface{}) {
 		return
 	}
 	s := fmt.Sprint(v...)
-	if !l.logToStdErr {
-		l.print(s)
-		l.print("\n")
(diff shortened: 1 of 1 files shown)


ostromart:
I saw that issue but some errors we do want to show up in stdout as well. I don't think we should redirect everything to stderr like this - need to think this through a bit more.

elfinhe:
> I saw that issue but some errors we do want to show up in stdout as well. I don't think we should redirect everything to stderr like this - need to think this through a bit more.

The current print is fine. But not feeling comfortable for having a hidden log file (.mesh_cli.log) in the current folder. We can talk about this later.

ostromart:
Yeah we should definitely take out the hidden log file now. Feel free to do that, I totally agree.

PR #378, opened 2019-10-14
Fix NPE

irisdingbj:
when CreatePatch fails, the mutatedObj will be nil. So use `obj` instead to prevent panic:
```
panic: runtime error: invalid memory address or nil pointer dereference

```

### DIFF
--- a/pkg/helmreconciler/rendering.go
+++ b/pkg/helmreconciler/rendering.go
@@ -240,7 +240,7 @@ func (h *HelmReconciler) ProcessObject(obj *unstructured.Unstructured) error {
 				log.Errorf("unexpected error occurred during postprocessing of updated resource: %s", err)
 			}
 		} else {
-			listenerErr := h.customizer.Listener().ResourceError(mutatedObj, err)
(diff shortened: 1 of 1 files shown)


richardwxn (approved):

pkg/helmreconciler/rendering.go
```diff
@@ -240,7 +240,7 @@ func (h *HelmReconciler) ProcessObject(obj *unstructured.Unstructured) error {
 				log.Errorf("unexpected error occurred during postprocessing of updated resource: %s", err)
 			}
 		} else {
-			listenerErr := h.customizer.Listener().ResourceError(mutatedObj, err)
+			listenerErr := h.customizer.Listener().ResourceError(obj, err)
```
ostromart: Thanks for looking into this Iris. I find this function (and especially this block) quite difficult to read and I think it could be structured better to avoid having so much nesting. I wouldn't be surprised if we find more errors of this type in here without some cleanup.

ostromart (approved):

PR #379, opened 2019-10-14
cross-repo e2e gating for the operator

sdake:
This work runs e2e testing using end to end testing against the
istio/istio repository.

To run - use `make e2e`.

Depends-On: https://github.com/istio/istio/pull/17864

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -69,6 +69,7 @@ require (
 	k8s.io/kube-proxy v0.0.0-00010101000000-000000000000 // indirect
 	k8s.io/kubectl v0.0.0-20191003004222-1f3c0cd90ca9
 	sigs.k8s.io/controller-runtime v0.2.2
+	sigs.k8s.io/kind v0.5.1 // indirect
 	sigs.k8s.io/yaml v1.1.0
 )
 
--- a/go.mod
+++ b/go.mod
@@ -69,7 +69,6 @@ require (
 	k8s.io/kube-proxy v0.0.0-00010101000000-000000000000 // indirect
 	k8s.io/kubectl v0.0.0-20191003004222-1f3c0cd90ca9
 	sigs.k8s.io/controller-runtime v0.2.2
-	sigs.k8s.io/kind v0.5.1 // indirect
 	sigs.k8s.io/yaml v1.1.0
 )
 
--- a/tests/e2e/e2e.sh
+++ b/tests/e2e/e2e.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+
+# Copyright 2019 Istio Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eux
+
+export ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
+
+function cleanup_kind_cluster() {
+    kind export logs --name istio-testing "${ARTIFACTS}/kind"
+#    kind delete cluster --name=istio-testing
+}
+
+function setup_kind_cluster() {
+  # Delete any previous e2e KinD cluster
+  echo "Deleting previous KinD cluster with name=istio-testing"
+  if ! (kind delete cluster --name=istio-testing) > /dev/null; then
+    echo "No existing kind cluster with name istio-testing. Continue..."
+  fi
+
+  trap cleanup_kind_cluster EXIT
+
+  # Create KinD cluster
+  if ! (kind create cluster --name=istio-testing --loglevel debug --retain --wait 30s); then
+    echo "Could not setup KinD environment. Something wrong with KinD setup. Exporting logs."
+    exit 1
+  fi
+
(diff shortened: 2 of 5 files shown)


tests/e2e/e2e.sh
```diff
@@ -0,0 +1,97 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eux
+
+export ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
+
+function cleanup_kind_cluster() {
+    kind export logs --name istio-testing "${ARTIFACTS}/kind"
+#    kind delete cluster --name=istio-testing
```
howardjohn: This is a must have! see https://github.com/kubernetes-sigs/kind/issues/759. We should have a consolidated kind library. Maybe in common files? Opened https://github.com/istio/istio/issues/17948 to track this

tests/e2e/e2e.sh
```diff
@@ -0,0 +1,97 @@
...
+metadata:
+  namespace: metallb-system
+  name: config
+data:
+  config: |
+    address-pools:
+    - name: default
+      protocol: layer2
+      addresses:
+      - 172.17.255.1-172.17.255.250
```
howardjohn: will this work in all environments? An alternative used in installer is a nodeport

tests/e2e/e2e.sh
```diff
@@ -0,0 +1,97 @@
...
+EOF
+
+  kubectl apply -f https://raw.githubusercontent.com/google/metallb/v0.8.1/manifests/metallb.yaml
+  kubectl apply -f ./metallb-config.yaml
+}
+
+function setup_docker() {
+#  HUB=istio-testing TAG=istio-testing make -f Makefile.core.mk docker
```
howardjohn: don't we need to build?

tests/e2e/e2e.sh
```diff
@@ -0,0 +1,97 @@
...
+do
+	cat "${manifest_file}" >> "${ISTIO_DIR}"/install/kubernetes/istio-operator.yaml
+	echo "---" >> "${ISTIO_DIR}"/install/kubernetes/istio-operator.yaml
+done
+
+kubectl get pods --all-namespaces -o wide
+
+if [[ ! -d "${ISTIO_DIR}" ]]
+then
+  git clone https://github.com/istio/istio.git "${ISTIO_DIR}"
```
howardjohn: shouldn't this be done before we write the istio-operator.yaml?

tests/e2e/e2e.sh
sdake: I agree a consolidated setup library would be great.  When I come up for air after 1.4, I had planned to work a bit more on cross-repo gating and make things generic enough so that different components can use the same set of scripts...

sdake: comment removed from line 23.

tests/e2e/e2e.sh
sdake: I am not sure.  It works where we need it the most - which is automated testing.  I have read in the past this may not work well on mac, but I have no reason to think it may not.  I'd suggest a little latitude here, so I can prove this out, and if things are not working, I can setup the call to --use-local or whatever it is for e2e.  We have a project goal of getting rid of nodeport testing...

tests/e2e/e2e.sh
sdake: done.

tests/e2e/e2e.sh
sdake: done.  Not sure why things worked well for me prior.

sdake:
@howardjohn PTAL thanks!

go.mod
```diff
@@ -69,6 +69,7 @@ require (
 	k8s.io/kube-proxy v0.0.0-00010101000000-000000000000 // indirect
 	k8s.io/kubectl v0.0.0-20191003004222-1f3c0cd90ca9
 	sigs.k8s.io/controller-runtime v0.2.2
+	sigs.k8s.io/kind v0.5.1 // indirect
```
howardjohn: this should not be here and pulls in dependencies that may not be ok for istio/istio

howardjohn (commented):
looks good except that one comment

go.mod
sdake: Ya not sure why this is present will fix this after dinner tonight.

sdake: Ya not sure why this is present will fix this after dinner tonight.

morvencao:
@sdake Can we addd license `CC-BY-4.0` to https://github.com/istio/operator/blob/master/common/config/license-lint.yml to pass the license linter?

morvencao (commented):
@sdake Can we addd license CC-BY-4.0 to https://github.com/istio/operator/blob/master/common/config/license-lint.yml to pass the license linter?

sdake:
> @sdake Can we addd license CC-BY-4.0 to https://github.com/istio/operator/blob/master/common/config/license-lint.yml to pass the license linter?

I reverted the gomod changes so the liniter should be gtg now.

tests/e2e/e2e.sh
```diff
@@ -0,0 +1,99 @@
...
+setup_docker
+
+mkdir -p "${ARTIFACTS}/out"
+
+ISTIO_DIR="${GOPATH}/src/istio.io/istio"
+
+# Create a clone of the Istio repository
+if [[ ! -d "${ISTIO_DIR}" ]]
+then
+  git clone https://github.com/sdake/istio.git "${ISTIO_DIR}"
```
linsun: is this correct? pointing to your personal repo?

linsun: also, not sure why this cmd is needed?  Seems you need to clone the istio-operator repo to get the deploy dir.

howardjohn: please follow up on this

howardjohn (approved):

PR #380, opened 2019-10-15
Test Switch Profile for IstioControlPlane via controller

irisdingbj:


pkg/controller/istiocontrolplane/istiocontrolplane_controller_test.go
```diff
@@ -0,0 +1,196 @@
...
+func checkICPStatus(cl client.Client, key client.ObjectKey, profile string) (bool, error) {
+	minimalStatus := map[string]*v1alpha2.InstallStatus_VersionStatus{
+		"Pilot": {
+			Status: v1alpha2.InstallStatus_HEALTHY,
+		},
+		"crds": {
+			Status: v1alpha2.InstallStatus_HEALTHY,
+		},
+	}
+	defaultStatus := map[string]*v1alpha2.InstallStatus_VersionStatus{
```
morvencao: Maybe we can make these variables be constant in follow up PR as we need to more profile switch cases.

morvencao (approved):

PR #381, opened 2019-10-15
fix istiocontrolplane cr deletion.

### LINKED ISSUE #17883: The deletionTimestamp and finalizers fields are always nil in the IstioControlPlane instance
The controller can’t get `deletionTimestamp` and `finalizers` from the IstioControlPlane instance, while they do exist, because I can get the instance from api-server with kubectl:
```
root@mcdev1:~/go/src/istio.io/operator# kc get istiocontrolplane example-istiocontrolplane -o yaml
kind: IstioControlPlane
metadata:
  creationTimestamp: "2019-10-12T12:45:39Z"
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2019-10-12T12:46:52Z"
  finalizers:
  - istio-operator
  generation: 2
  namespace: default
  resourceVersion: "1218"
  selfLink: /apis/install.istio.io/v1alpha2/namespaces/default/istiocontrolplanes/example-istiocontrolplane
  uid: 85a3dc1e-fa49-47c2-8210-247f1d39e1a3
spec:
  profile: default
status:
  status:
    Citadel:
      status: 2
    Galley:
      status: 2
    IngressGateway:
      status: 2
    Injector:
      status: 2
    Pilot:
      status: 2
    Policy:
      status: 2
    Prometheus:
      status: 2
    Telemetry:
      status: 2
    crds:
      status: 2
```

Generally, after the end-users delete the IstiocontrolPlane CR with kubectl, the `deletionTimeStamp` will be added to the IstiocontrolPlane CR, and the controller will start to handle the `finalizers`.
The current situation is that, the controller can’t get deletionTimeStamp field(it’s always nil in my env) after I delete the IstiocontrolPlane CR with kubectl.

morvencao:
resolve: https://github.com/istio/istio/issues/17883

### DIFF
--- a/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
+++ b/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
@@ -18,6 +18,7 @@ import (
 	"context"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -28,6 +29,7 @@ import (
 
 	"istio.io/operator/pkg/apis/istio/v1alpha2"
 	"istio.io/operator/pkg/helmreconciler"
+	"istio.io/operator/pkg/util"
 	"istio.io/pkg/log"
 )
 
@@ -109,8 +111,16 @@ func (r *ReconcileIstioControlPlane) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
-	deleted := instance.GetDeletionTimestamp() != nil
-	finalizers := instance.GetFinalizers()
+	// Workaroud for issue: https://github.com/istio/istio/issues/17883
+	// Using an unstructured object to get deletionTimestamp and finalizers fields.
+	u := &unstructured.Unstructured{}
+	u.SetGroupVersionKind(util.IstioOperatorGVK)
+	err = r.client.Get(context.TODO(), request.NamespacedName, u)
+	if err != nil {
+		log.Errorf("error getting the unstructured of IstioControlPlane instance: %s", err)
+	}
+	deleted := u.GetDeletionTimestamp() != nil
+	finalizers := u.GetFinalizers()
 	finalizerIndex := indexOf(finalizers, finalizer)
 
 	if deleted {


irisdingbj:
This workaround works!

irisdingbj (approved):

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
```diff
@@ -109,8 +111,16 @@ func (r *ReconcileIstioControlPlane) Reconcile(request reconcile.Request) (recon
 		return reconcile.Result{}, err
 	}
 
-	deleted := instance.GetDeletionTimestamp() != nil
-	finalizers := instance.GetFinalizers()
+	// Workaroud for issue: https://github.com/istio/istio/issues/17883
+	// Using an unstructured object to get deletionTimestamp and finalizers fields.
+	u := &unstructured.Unstructured{}
```
sdake: @morvencao 

I appreciate the work.  The operator should be using versioned client libraries for reading and writing IstioControlPlane's CR (as well as listening for changes, etc).  I think we can merge this as is.  The code base is in a need of overhaul to properly provide a clientset/informers/listeners that are compatible with `controller-runtime`.  I've been banging on it for about 6 hours, and have something that nearly almost compiles..., but it is very disruptive and I am not keen to merge this work prior to the release of 1.4.

I will however post my work a a nonmergeable PR so folks can see what I am talking about in this comment :)

We should probably do a more thorough audit of the operator code base for these problems.

Cheers
-steve

morvencao: @sdake Understood.
As a workaround, I really hope this can be merged so that we can continue to test more cases and also make e2e tests(coming soon I think) work as expected, the original issue https://github.com/istio/istio/issues/17883 is a blocker for us to proceed.

Maybe I should keep the original github issue https://github.com/istio/istio/issues/17883 in open state until the API marshal/unmarshal issue is completely resolved.

/cc @ostromart

sdake:
placing a short hold on this issue. Will followup with a PR that I believe fixes this problem correctly (but introduces other errors).

Really really nice job on the root cause @morvencao !!

sdake:
@morvencao when I'm resting, can you take on finishing out the remainder of this PR https://github.com/istio/operator/pull/382?  I will make more progress today hopefully, but if I am not done, I'd like to see 382 land so we can unblock e2e.

Cheers
-steve

ostromart (approved):

ostromart:
Let's go ahead with this workaround. A lot of our headaches are caused by having to use proto as a source of truth - protoc generates structs that are not fully compatible with k8s and we have to use jsonpb and gogo to marshal/unmarshal. K8s themselves use structs as the source of truth and generate protos from those but there are some issues with doing that for Istio. This is a very involved problem and a perfect one to tackle at the start of 1.5; therefore I suggest we pause this effort until we feel confident operator is fundamentally working well. We only have a couple of days before code freeze...

rcernich:
@ostromart, fwiw, the client-go generation that I did incorporates deepcopy gen into the api build, for specific types used in k8s.  This handles some of that magic.  If you go that route, it would be better to generate the json shims in api.  That said, there is still some work to do, as the names for json fields are not correct.  (The shim only generates marshalers for types with oneof fields, but because of the json issue, it should really generate the marshaler for all types.)

pkg/controller/istiocontrolplane/istiocontrolplane_controller.go
sdake: yes please keep the issue open.  I think we need to make an exit from gogo - should simplify the implementation.

sdake:
> @morvencao when I'm resting, can you take on finishing out the remainder of this PR #382? I will make more progress today hopefully, but if I am not done, I'd like to see 382 land so we can unblock e2e.
> 
> Cheers
> -steve

We will try unblocking e2e first and see if the e2e test cases work.  I have 3 PRs outstanding to implement *very basic* e2e testing for the operator (runs the simple test).  Will sort out getting them merged, then we can take a checkpoint on what is next during wednesday's environments meeting.

PR #348, opened 2019-10-10
istioctl x upgrade

elfinhe:
This is a WIP `istioctl x upgrade` command to automate the Istio Control Plane In-place Upgrade.
[[design doc]](https://docs.google.com/document/d/1OdnfyBai5ytjQy6lwU1KWOVgRHf-RSIRDLx_SFpOXGE/edit#)

I will continue to update the code with more comments, re-organize the code structure and add more unit tests. 

Comments are welcome!

Sample Output:
```
$ mesh upgrade --wait
Control Plane - citadel pod - version: 1.3.0
Control Plane - galley pod - version: 1.3.0
Control Plane - ingressgateway pod - version: 1.3.0
Control Plane - pilot pod - version: 1.3.0
Control Plane - policy pod - version: 1.3.0
Control Plane - sidecar-injector pod - version: 1.3.0
Control Plane - telemetry pod - version: 1.3.0
Client - istioctl version: 1.3.1

Version check passed: 1.3.0 -> 1.3.1.

Upgrade check: values will be changed during the upgrade:
citadel: map[dnsCerts:map[istio-pilot-service-account.istio-system:istio-pilot.istio-system]
  enableNamespacesByDefault:true image:citadel rollingMaxSurge:100% rollingMaxUnavailable:25%
  selfSigned:true trustDomain:cluster.local] ->
cni: -> map[enabled:false namespace:istio-system]
galley:
  rollingMaxSurge: 100% ->
  rollingMaxUnavailable: 25% ->
gateways:
  istio-egressgateway:
    connectTimeout: 10s ->
    drainDuration: 45s ->
    rollingMaxSurge: 100% ->
    rollingMaxUnavailable: 25% ->
  istio-ingressgateway:
    rollingMaxSurge: 100% ->
    rollingMaxUnavailable: 25% ->
    telemetry_addon_gateways: map[grafana_gateway:map[desPort:3000 enabled:false name:grafana
      port:15031 tls:false] kiali_gateway:map[desPort:20001 enabled:false name:kiali
      port:15029 tls:false] prometheus_gateway:map[desPort:9090 enabled:false name:prometheus
      port:15030 tls:false] tracing_gateway:map[desPort:80 enabled:false name:tracing
      port:15032 tls:false]] ->
    telemetry_domain_name: ' ->'
global:
  proxy:
    protocolDetectionTimeout: -> 1s
  tag: 1.3.0 -> 1.3.1
grafana:
  namespace: -> istio-system
istio_cni: map[enabled:false] ->
[...]

PR #385, opened 2019-10-15
Add package for upgrade hooks

ostromart:
This package for istioctl upgrade command: https://github.com/istio/operator/pull/348.

### DIFF
--- a/pkg/hooks/upgrade_hooks.go
+++ b/pkg/hooks/upgrade_hooks.go
@@ -1,0 +1,184 @@
+// Copyright 2019 Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


pkg/hooks/upgrade_hooks.go
```diff
@@ -0,0 +1,184 @@
...
+
+var (
+	// preUpgradeHooks is a list of hook version constraint pairs mapping to a slide of corresponding hooks to run
+	// before upgrade.
+	preUpgradeHooks = []hookVersionMapping{
+		{
+			sourceVersionConstraint: "1.3",
+			targetVersionConstraint: "",
+			hooks:                   []hook{checkInitCrdJobs},
+		},
```
elfinhe: I am thinking abstract this into an interface, so the hook can be a self-contained type and easy to implement by others without understanding the code here.

elfinhe (approved):
LGTM. We can leave the comment as a follow-up action item when hooks have more instances.

pkg/hooks/upgrade_hooks.go
ostromart: yes, it could be done like this. my only concern would be if we allow hooks to be registered from anywhere, the code may become quite scattered. keeping all this private (except runPreUpgradeHooks and runPostUpgradeHooks, which should be public) has the advantage that all these hacks are contained to the hooks package.

morvencao (approved):

PR #389, opened 2019-10-16
Remove unused codes from inputs

irisdingbj:


ostromart:
I was surprised that this code could be removed because it's capturing the dependency tree relationship. Then I saw that the logic was rewritten in https://github.com/istio/operator/pull/371, but I don't understand why there was a deadlock and why it was fixed like this - could you explain?

irisdingbj:
> I was surprised that this code could be removed because it's capturing the dependency tree relationship. Then I saw that the logic was rewritten in #371, but I don't understand why there was a deadlock and why it was fixed like this - could you explain?

@ostromart We now will calculate the dependency dynamically according to the generated manifest instead of the static data here. The reason is there will be a dead lock for the waiting group : pkg/helmreconciler/reconciler.go:L164. 

Because only those components which existed in the manifest will read from the dependency channel while originally with the static dependency data we will write to the channel for all components.

This will cause the wg.wait() always hangs there because below lines will be blocked there:
```
// Signal all the components that depend on us.
			for _, ch := range deps[cn] {
				log.Infof("Unblocking dependency %s.", ch)
				dch[ch] <- struct{}{}
			}
```

PR #390, opened 2019-10-16
replace --yes flag with --skip-confirmation flag for better usage

carolynhu:
Here is the current `--skip-confirmation` behavior to replace the confusing `--yes` flag:
When `--skip-confirmation` flag is true, then, do not ask for confirmation; otherwise, ask for confirmation.
(1) `mesh manifest apply`
Ask for confirmation:
```
mesh manifest apply
Do you want to proceed? (y/N)
```
Since `--skip-confirmation` by default is `false`. If the user just type `enter`, then it will showing `Cancelled`. Since, the default answer is `No`.
(2) `mesh manifest apply --skip-confirmation` or  `mesh manifest apply --ask-confirm=true`
Do not ask for confirmation.
(3) `mesh manifest apply --skip-confirmation=false`
Ask for confirmation:
 ```
mesh manifest apply
Do you want to proceed? (y/N)
```

### DIFF
--- a/cmd/mesh/manifest-apply.go
+++ b/cmd/mesh/manifest-apply.go
@@ -35,8 +35,9 @@ type manifestApplyArgs struct {
 	readinessTimeout time.Duration
 	// wait is flag that indicates whether to wait resources ready before exiting.
 	wait bool
-	// yes means don't ask for confirmation (asking for confirmation not implemented)
-	yes bool
+	// skipConfirmation determines whether the user is prompted for confirmation.
+	// If set to true, the user is not prompted and a Yes response is assumed in all cases.
+	skipConfirmation bool
 	// force proceeds even if there are validation errors
 	force bool
 	// set is a string with element format "path=value" where path is an IstioControlPlane path and the value is a
@@ -48,7 +49,7 @@ func addManifestApplyFlags(cmd *cobra.Command, args *manifestApplyArgs) {
 	cmd.PersistentFlags().StringVarP(&args.inFilename, "filename", "f", "", filenameFlagHelpStr)
 	cmd.PersistentFlags().StringVarP(&args.kubeConfigPath, "kubeconfig", "c", "", "Path to kube config")
 	cmd.PersistentFlags().StringVar(&args.context, "context", "", "The name of the kubeconfig context to use")
-	cmd.PersistentFlags().BoolVarP(&args.yes, "yes", "y", false, "Do not ask for confirmation")
+	cmd.PersistentFlags().BoolVar(&args.skipConfirmation, "skip-confirmation", false, skipConfirmationFlagHelpStr)
 	cmd.PersistentFlags().BoolVar(&args.force, "force", false, "Proceed even with validation errors")
 	cmd.PersistentFlags().DurationVar(&args.readinessTimeout, "readiness-timeout", 300*time.Second, "Maximum seconds to wait for all Istio resources to be ready."+
 		" The --wait flag must be set for this flag to apply")
@@ -65,8 +66,8 @@ func manifestApplyCmd(rootArgs *rootArgs, maArgs *manifestApplyArgs) *cobra.Comm
 		Args:  cobra.ExactArgs(0),
 		Run: func(cmd *cobra.Command, args []string) {
 			l := newLogger(rootArgs.logToStdErr, cmd.OutOrStdout(), cmd.OutOrStderr())
-			if !maArgs.yes && maArgs.kubeConfigPath == "" && maArgs.context == "" {
-				if !confirm("Are you sure?", cmd.OutOrStdout()) {
+			if !maArgs.skipConfirmation && maArgs.kubeConfigPath == "" && maArgs.context == "" {
+				if !confirm("Do you want to proceed? (y/N)", cmd.OutOrStdout()) {
 					cmd.Print("Cancelled.\n")
 					os.Exit(1)
 				}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cmd/mesh/manifest-apply.go [hunk 1]
```diff
@@ -38,7 +38,7 @@ type manifestApplyArgs struct {
 	readinessTimeout time.Duration
 	// wait is flag that indicates whether to wait resources ready before exiting.
 	wait bool
-	// yes means don't ask for confirmation (asking for confirmation not implemented)
+	// yes means don't ask for confirmation when manifest apply
```
ayj: "yes means don't..." is a little confusing. What about "If true, apply the manifest without asking for confirmation. If false, apply will ask for an explicit confirmation before modifying the cluster state."?

cmd/mesh/root.go
```diff
@@ -28,6 +28,8 @@ const (
 Overrides the corresponding path value in the selected profile or passed through IstioControlPlane CR
 customization file`
 	filenameFlagHelpStr = `Path to file containing IstioControlPlane CustomResource`
+	yesFlagHelpStr      = `Do not ask for confirmation when manifest apply. If --yes or -y flag is not set, its default value is false. 
+If --yes or -y flag is set without any options, then its default value is true`
 )
```
carolynhu: @ayj Here I explained how --yes flag being used and what's the default value with or without --yes flag. I agree, this flag is really confusing.  Let me think how to make it better for usage.

carolynhu:
@ayj please take another look about this PR. Thank you.

ayj:
Looking better. cc @sdake @ostromart since I think they discussed the use of `-y` previously.

cmd/mesh/manifest-apply.go
```diff
@@ -51,7 +52,7 @@ func addManifestApplyFlags(cmd *cobra.Command, args *manifestApplyArgs) {
 	cmd.PersistentFlags().StringVarP(&args.inFilename, "filename", "f", "", filenameFlagHelpStr)
 	cmd.PersistentFlags().StringVarP(&args.kubeConfigPath, "kubeconfig", "c", "", "Path to kube config")
 	cmd.PersistentFlags().StringVar(&args.context, "context", "", "The name of the kubeconfig context to use")
-	cmd.PersistentFlags().BoolVarP(&args.yes, "yes", "y", false, "Do not ask for confirmation")
+	cmd.PersistentFlags().BoolVar(&args.askForConfirm, "ask-confirm", true, "Whether ask for confirmation or not when manifest apply.")
```
ostromart: It's better to make the default be false if possible. In this case I suggest calling this "skip-confirmation" instead and make the description:
If set to true, the command will skip prompts and always assume a Yes response.

carolynhu: @ostromart done 
cc @ayj

cmd/mesh/manifest-apply.go
```diff
@@ -35,8 +35,9 @@ type manifestApplyArgs struct {
 	readinessTimeout time.Duration
 	// wait is flag that indicates whether to wait resources ready before exiting.
 	wait bool
-	// yes means don't ask for confirmation (asking for confirmation not implemented)
-	yes bool
+	// skipConfirmation means whether skip for confirmation when manifest apply or not
+	// if skipConfirmation=true, it means skip for confirmation; otherwise, ask for confirmation
```
ostromart: skipConfirmation determines whether the user is prompted for confirmation. If set to true, the user is not prompted and a Yes response is assumed in all cases.

cmd/mesh/manifest-apply.go
```diff
@@ -65,8 +66,8 @@ func manifestApplyCmd(rootArgs *rootArgs, maArgs *manifestApplyArgs) *cobra.Comm
 		Args:  cobra.ExactArgs(0),
 		Run: func(cmd *cobra.Command, args []string) {
 			l := newLogger(rootArgs.logToStdErr, cmd.OutOrStdout(), cmd.OutOrStderr())
-			if !maArgs.yes && maArgs.kubeConfigPath == "" && maArgs.context == "" {
-				if !confirm("Are you sure?", cmd.OutOrStdout()) {
+			if !maArgs.skipConfirmation && maArgs.kubeConfigPath == "" && maArgs.context == "" {
+				if !confirm("Do you wish to run this command? (Y/N)", cmd.OutOrStdout()) {
```
ostromart: Sorry, this is real nitpicking but I suggest "Proceed?" instead.
"Do you wish to run this command" suggests in some sense that the command might be optional.

cmd/mesh/manifest-apply.go
```diff
@@ -48,7 +49,7 @@ func addManifestApplyFlags(cmd *cobra.Command, args *manifestApplyArgs) {
 	cmd.PersistentFlags().StringVarP(&args.inFilename, "filename", "f", "", filenameFlagHelpStr)
 	cmd.PersistentFlags().StringVarP(&args.kubeConfigPath, "kubeconfig", "c", "", "Path to kube config")
 	cmd.PersistentFlags().StringVar(&args.context, "context", "", "The name of the kubeconfig context to use")
-	cmd.PersistentFlags().BoolVarP(&args.yes, "yes", "y", false, "Do not ask for confirmation")
+	cmd.PersistentFlags().BoolVar(&args.skipConfirmation, "skip-confirmation", false, "Whether skip for confirmation or not when manifest apply.")
```
ostromart: Use the same text as the variable comment.

ostromart (approved):

PR #395, opened 2019-10-16
Make versions.yaml also compiled-in

ostromart:
Also removing some lingering versions related info which is unused and just causes confusion.


data/versions.yaml
```diff
@@ -0,0 +1,15 @@
...
+  supportedIstioVersions: ">=1.3.0,<=1.3.1"
+  recommendedIstioVersions: 1.3.1
+- operatorVersion: 1.3.2
+  supportedIstioVersions: ">=1.3.0,<=1.3.2"
+  recommendedIstioVersions: 1.3.2
+- operatorVersion: 1.3.3
+  supportedIstioVersions: ">=1.3.0,<=1.3.3"
+  recommendedIstioVersions: 1.3.3
+- operatorVersion: 1.4.0
+  supportedIstioVersions: ">=1.3.3, <1.6"
```
elfinhe: Sorry, I missed this part. Are we planning to use the ICP v1alpha2 API for 1.4 and 1.5? Any doc I can learn more about the plan?

elfinhe (approved):
LGTM.

data/versions.yaml
ostromart: For 1.4 yes, for 1.5 it's TBD but probably not. The user would have to keep their old ICP CR around if they wish to downgrade. 
For upgrade, we will provide some tool or guides, but likely 1.4->1.5 will require two sets of configs to be present side by side.

morvencao (approved):

PR #396, opened 2019-10-16
Start to use the upgrade-hooks pkg

elfinhe:
Put the work together from https://github.com/istio/operator/pull/385 and https://github.com/istio/operator/pull/348.

Using the new upgrade-hooks pkg in upgrade command, remove the old one.


cmd/mesh/upgrade.go
```diff
@@ -157,21 +160,43 @@ func upgrade(rootArgs *rootArgs, args *upgradeArgs, l *logger) (err error) {
...
+	// Create a new values translator
+	tr, err := translate.NewReverseTranslator(opversion.OperatorBinaryVersion.MinorVersion)
+	if err != nil && !args.force {
+		return fmt.Errorf("fail to create values translator, error: %v", err)
+	}
+
+	// Translate the current values into ICPS
+	var currentICPS *v1alpha2.IstioControlPlaneSpec
+	if tr != nil {
+		currentICPS, err = tr.TranslateFromValueToSpec([]byte(currentValues))
```
ostromart: I don't think we should be doing this inside of upgrade. It's better to use manifest migrate and then run upgrade on the resulting ICP.

elfinhe: talked offline, we will pass in the only icps for both parameters. We will consider API changes maybe in 1.5 or later.

ostromart (approved):

PR #399, opened 2019-10-17
add test for manager.

morvencao:


morvencao:
@sdake @geeknoid I have added tests for the controller manager which needs kubebuilder binaries to get started, would you help to add kubebuilder binaries to the `build-tools` image.

elfinhe (approved):
LGTM, @sdake @howardjohn please consider adding kubebuilder into build-tools.

PR #400, opened 2019-10-17
Add more cluster role rules for operator

linsun:
and organized this file a bit.

I found a lot of error in my operator log, like below, so I took a pass at giving more rules to the operator:

```
2019-10-17T02:02:13.627842Z	warn	retrieving resources to prune type security.istio.io/v1beta1, Kind=AuthorizationPolicy: authorizationpolicies.security.istio.io is forbidden: User "system:serviceaccount:istio-operator:istio-operator" cannot list resource "authorizationpolicies" in API group "security.istio.io"

2019-10-17T02:02:13.651208Z	warn	retrieving resources to prune type admissionregistration.k8s.io/v1beta1, Kind=ValidatingWebhookConfiguration: validatingwebhookconfigurations.admissionregistration.k8s.io is forbidden: User "system:serviceaccount:istio-operator:istio-operator" cannot list resource "validatingwebhookconfigurations" in API group "admissionregistration.k8s.io"
```


deploy/clusterrole.yaml
```diff
@@ -85,17 +82,32 @@ rules:
...
   - '*'
-  - istiocontrolplanes
+- apiGroups:
+  - ""
+  resources:
+  - configmaps  
+  - endpoints
+  - events
+  - namespaces
+  - jobs
```
rcernich: Is the operator installing any jobs?  I suspect this should be removed.  If it is, we should move the work being done by the job into the operator itself.

linsun: Good point. I saw it in the log to query `jobs.batch is forbidden`

My env is a bit tossed, let me test it a bit more.

linsun: not seeing jobs after installing the default cr.  going to remove it for now.

rcernich (approved):

ostromart:
@linsun the label doesn't work (can we fix it or remove it @howardjohn ?)
/cherrypick release-1.4

howardjohn:
The label does work if its applied before the PR merges. I do have an open
issue https://github.com/kubernetes/test-infra/issues/14244

PR #404, opened 2019-10-17
WIP: test of the operator check for e2e

sdake:


ostromart:
Hi @sdake, will you be able to make progress on this? Otherwise we should find another owner since it's quite critical.

sdake:
@ostromart looks like @irisdingbj will take this over.  Note my original work was passing in my local environment and also prow.  There is some minor issue - possibly related to the usage of metallb - but possibly something else.

I'll take a look today after my morning meetings finish up.

Cheers
-steve

sdake:
subsumed by work from @irisdingbj .

PR #414, opened 2019-10-18
fix controller panic.

morvencao:
resolve: https://github.com/istio/istio/issues/18041

PR #415, opened 2019-10-18
prune citadel resources

### LINKED ISSUE #18031: deleting ICP cr left a few resources
**Bug description**

After I delete my ICP (istio control plane) cr, all pods/deployments/services in Istio-system are removed.  However, the istio-security ConfigMap remains.  Also, all secrets remain, plus the default service account for istio-system namespace.   All other configmaps and service accounts do get removed.

I noticed what is left don't have the proper labels needed for operator to remove them.

For example, I'm not seeing these labels on the resources that remain which I believe caused the operator to not prune them.

```
 labels:
    install.operator.istio.io/owner-group: install.istio.io
    install.operator.istio.io/owner-kind: IstioControlPlane
    install.operator.istio.io/owner-name: example-istiocontrolplane

```

@morvencao could you take a look at this prob if you have time?  

**Affected product area (please put an X in all that apply)**

[ ] Docs
[ ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Expected behavior**

**Version (include the output of `istioctl version --remote` and `kubectl version`)**

**How was Istio installed?**

Additionally, please consider attaching a [cluster state archive](http://istio.io/help/bugs/#generating-a-cluster-state-archive) by attaching
the dump file to this issue.

morvencao:
depends on: https://github.com/istio/operator/pull/414
resolve: https://github.com/istio/istio/issues/18031


pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -243,6 +246,77 @@ func (c *IstioDefaultChartCustomizer) waitForDeployment(object runtime.Object) {
...
+	if err := client.List(context.TODO(), secretList); err != nil {
+		return fmt.Errorf("error getting secret list: %s", err)
+	}
+
+	for retryCount := 1; retryCount <= cleanCitadelResourcesMaxRetries; retryCount++ {
+		if err := client.Delete(context.TODO(), cm); err != nil && retryCount == cleanCitadelResourcesMaxRetries {
+			return fmt.Errorf("error deleting ConfigMap %s/%s: %s", cmNamespace, cmName, err)
+		}
+
+		for _, secret := range secretList.Items {
```
howardjohn: today, I'd you delete istio your sidecars will still function properly. If you delete the secrets I don't think they will

linsun: I assume this is only in mTLS strict mode perhaps?  I just did a test, when i wipe out my istio-systems ns, my bookinfo still function okay, with sidecars.

When all istio CRDs are removed in our case when the removal of icp cr, i don't see how you enforce istio mTLS strict mode...

morvencao: https://github.com/istio/istio/issues/18031#issuecomment-543502295
does it mean that we only need to delete the secrets during upgrading istio? For istio uninstallation, we should keep these secrets?

howardjohn: Do not delete them during upgrade or you will get a cluster wide traffic loss. Once the certs are deleted envoy will hot restart without the certs and start failing mtls calls.

Maybe we don't care about this in deletion, but definitely in upgrade

linsun: Agree on not deleting during upgrade.

We should delete the secrets when removal of the icp CR when users have decided they don't want to use istio and we return a cleanup for them.

linsun:
@morvencao great work, thank you!!!

I'm good with the PR, will wait till a few others have eye on it.

Just confirming, this will delete the secrets and config map upon the icp cr removal.   With this PR, The namespace and service account associated with the namespace will continue to remain.  Is that a good summary?

pkg/controller/istiocontrolplane/listeners.go
morvencao: Got it, I need to make sure the `ResourceDeleted` will not invoked during upgrade.

morvencao:
@linsun 
Current situation is that if you don't have the `istio-system` namespace, then the controller e will help you to create one and after the icp cr is removed, the `istio-system` namespace together with all resources(including the secrets and configmap) in this namespace will be pruned.

However, if you do have the `istio-system` namespace before creating the icp CR, then this PR will help to clean all the secrets and configmap generated by citadel after the icp CR is removed, but keep the `istio-system` namespace.

linsun:
@morvencao that is awesome, that is the desired behavior I think makes sense to users.

pkg/controller/istiocontrolplane/listeners.go
sdake: and keep in mind if we install a new operator version (for z stream updates) - nothing should be deleted - instead the existing resources should just be reconciled.

pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -243,6 +248,77 @@ func (c *IstioDefaultChartCustomizer) waitForDeployment(object runtime.Object) {
...
+	cleanCitadelResourcesMaxRetries := 10
+	objAccessor, err := meta.Accessor(obj)
+	if err != nil {
+		return fmt.Errorf("get error to get namespace of %s: %s", obj.GetObjectKind().GroupVersionKind(), err)
+	}
+
+	client := c.Reconciler.GetClient()
+	cmNamespace := objAccessor.GetNamespace()
+	cmName := "istio-security"
+	prunedSecretType := "istio.io/key-and-cert"
```
linsun: What about these secrets in istio-system ns?

```
dns.istio-galley-service-account               istio.io/dns-key-and-cert             3      19h
dns.istio-sidecar-injector-service-account     istio.io/dns-key-and-cert             3      19h
istio-ca-secret                                istio.io/ca-root 
```

linsun: Also, I noticed you listed the cm name and secret type here... curious why we aren't adding the created by istio operator label and use that to determine if it should be cleaned vs hard coded name/secret type?

pkg/helmreconciler/interfaces.go
```diff
@@ -106,7 +106,7 @@ type RenderingListener interface {
 	// other charts; etc.
 	// obj represents a resource created from a manifest.
 	// The returned runtime.Object is the object that will be reconciled (created/updated).
```
linsun: add the new chart param in comment?

linsun:
@morvencao I think this looks great, I have a few minor comment.

One thing I could not figure out is where the namespace created by the operator is deleted? I don't see the ICP deployednamespace in pruningdetails.

@sdake @howardjohn ptal before this is merged.

morvencao:
@linsun I can add the namespace to the pruning list, but in that case, the istio-system namespace will be deleted in either case.

linsun:
@morvencao thank you - in that case, it is probably good to keep the namespace around for now until the namespace deletion topic has more discussion.  I'm not a fan of removing it if user has created it.

linsun (approved):

PR #416, opened 2019-10-18
Add vars for hub and tag that can be populated during build time

ostromart:
This addresses the issue of hub and tag being compiled in and committed before being pulled in before operator is imported as a module into istioctl.

### DIFF
--- a/cmd/mesh/profile-dump.go
+++ b/cmd/mesh/profile-dump.go
@@ -15,8 +15,10 @@
 package mesh
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
+	"text/template"
 
 	"github.com/ghodss/yaml"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


howardjohn (commented):
This looks like if the ld flag is not set then this will result in `tag: ""`? If so, we should set defaults here? Or does the overlay not apply if they are empty

ostromart:
See the change in profile-dump.go. It's only overlaying if these are set.

howardjohn (approved):

ostromart:
@howardjohn before we merge let's discuss how we can test this.

sdake (approved):
looks good.

howardjohn:
@ostromart 

if you mean how to manually test this 1 time: run `go build -ldflags "-X istio.io/operator/cmd/mesh.HubValueFromBuild=myhub"` and then do a `mesh generate` and inspect. (syntax here maybe isn't 100% accurate

if you mean how to add automated testing for this: in istio/release-builder we have a test that sanity checks a release, and right now that includes looking at all the charts for the right hub and tag, as well as running `istioctl version` and making sure info is populated, etc. It makes sense to add the same check there?

ostromart:
Manual test works and I added an automated test for the same.

PR #419, opened 2019-10-18
Add offline support for version maps of `istioctl x upgrade`

### LINKED ISSUE #17912: Upgrade version maps should be available offline
Version maps specify which versions are supported from which operator. The source of truth is master but we need a local copy built into operator for offline upgrade cases.

elfinhe:
Fall back to using the built-in version map if the remote one is not available. + some minor fixes on output.

Version maps specify which versions are supported from which operator. The source of truth is master but we need a local copy built into operator for offline upgrade cases.
Resolve: https://github.com/istio/istio/issues/17912

### DIFF
--- a/cmd/mesh/manifest-versions.go
+++ b/cmd/mesh/manifest-versions.go
@@ -104,7 +104,7 @@ func getVersionCompatibleMap(versionsURI string, binVersion *goversion.Version,
 		}
 	}
 	if myVersionMap == nil {
-		return nil, fmt.Errorf("this operator version %v was not found in the global manifestVersions map", binVersion.String())
+		return nil, fmt.Errorf("this operator version %s was not found in the global manifestVersions map", binVersion.String())
 	}
 	return myVersionMap, nil
 }
@@ -116,14 +116,14 @@ func loadCompatibleMapFile(versionsURI string, l *logger) (b []byte, err error) 
 			return
 		}
 	}
-	l.logAndPrintf("Warning. Failed to retrieve the version map from the remote url: %v, error: %v."+
-		" Now fall back to the local file system.", versionsURI, err)
+	l.logAndPrintf("Warning: failed to retrieve the version map from the remote URL (%s): %s. " +
+		"Falling back to the local file: %s", versionsURI, err, versionsURI)
 
 	b, err = ioutil.ReadFile(versionsURI)
 	if err == nil {
 		return
 	}
-	l.logAndPrintf("Warning. Failed to retrieve the version map from the local file: %v, error: %v."+
-		" Now fall back to the built-in version map.", versionsURI, err)
+	l.logAndPrintf("Warning: failed to retrieve the version map from the local file (%s): %s. " +
+			"Falling back to the internal version map.", versionsURI, err)
 	return vfs.ReadFile("versions.yaml")
 }
--- a/cmd/mesh/manifest-versions.go
+++ b/cmd/mesh/manifest-versions.go
@@ -17,16 +17,15 @@ package mesh
 import (
 	"fmt"
 	"io/ioutil"
-	"net/url"
 
 	goversion "github.com/hashicorp/go-version"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
 
-	"istio.io/operator/pkg/vfs"
-
 	"istio.io/operator/pkg/httprequest"
+	"istio.io/operator/pkg/util"
 	"istio.io/operator/pkg/version"
+	"istio.io/operator/pkg/vfs"
 	binversion "istio.io/operator/version"
 )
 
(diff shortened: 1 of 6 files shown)


cmd/mesh/manifest-versions.go
```diff
@@ -105,8 +104,26 @@ func getVersionCompatibleMap(versionsURI string, binVersion *goversion.Version,
 		}
 	}
 	if myVersionMap == nil {
-		l.logAndFatal("This operator version ", binVersion.String(),
-			" was not found in the global manifestVersions map.")
+		return nil, fmt.Errorf("this operator version %v was not found in the global manifestVersions map", binVersion.String())
```
ostromart: nit: use %s for string. %v is ok but not idiomatic.

cmd/mesh/manifest-versions.go
```diff
@@ -105,8 +104,26 @@ func getVersionCompatibleMap(versionsURI string, binVersion *goversion.Version,
...
+
+func loadCompatibleMapFile(versionsURI string, l *logger) (b []byte, err error) {
+	if _, err = url.ParseRequestURI(versionsURI); err == nil {
+		b, err = httprequest.Get(versionsURI)
+		if err == nil {
+			return
+		}
+	}
+	l.logAndPrintf("Warning. Failed to retrieve the version map from the remote url: %v, error: %v."+
+		" Now fall back to the local file system.", versionsURI, err)
```
ostromart: nit: "Warning: failed to retrieve the version map from the remote URL (%s ): %s. Falling back to internal version map."
Here and below.

cmd/mesh/manifest-versions.go
```diff
@@ -105,8 +104,26 @@ func getVersionCompatibleMap(versionsURI string, binVersion *goversion.Version,
...
 	}
 	if myVersionMap == nil {
-		l.logAndFatal("This operator version ", binVersion.String(),
-			" was not found in the global manifestVersions map.")
+		return nil, fmt.Errorf("this operator version %v was not found in the global manifestVersions map", binVersion.String())
+	}
+	return myVersionMap, nil
+}
+
+func loadCompatibleMapFile(versionsURI string, l *logger) (b []byte, err error) {
```
ostromart: The logic here is not quite right. We should try versionsURI as a get if it has a http: prefix. Otherwise try it as a local file path. 
Depending on which way we try it, emit a warning about either remote URL or file.

cmd/mesh/manifest-versions_test.go
```diff
@@ -0,0 +1,101 @@
...
+			args: args{
+				versionsURI: "__nonexistent-versions.yaml",
+				binVersion:  goVer132,
+				l:           l,
+			},
+			want:    vm132,
+			wantErr: nil,
+		},
+		{
+			name: "read from github",
```
ostromart: I suggest skipping the read from github unless you feel like implementing a server in the test so that it's hermetic. We don't want to be calling out to a real URL in a test.
Instead, I suggest pointing it to a file in testdata. This will allow you also to test the negative case if path is incorrect, should get an error.

cmd/mesh/manifest-versions_test.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	goVer131000, _ := goversion.NewVersion("1.3.1000")
+	goVer132, _ := goversion.NewVersion("1.3.2")
+	scVer132, _ := goversion.NewConstraint(">=1.3.0,<=1.3.2")
+	rcVer132, _ := goversion.NewConstraint("1.3.2")
+	vm132 := &version.CompatibilityMapping{
+		OperatorVersion:          goVer132,
+		SupportedIstioVersions:   scVer132,
+		RecommendedIstioVersions: rcVer132,
+	}
```
ostromart: this should probably come from the real versions map.
see version/version_test.go to see how it's read.

cmd/mesh/manifest-versions.go
elfinhe: Fixed.

cmd/mesh/manifest-versions.go
elfinhe: Sure, fixed.

cmd/mesh/manifest-versions.go
elfinhe: Yes, I fixed the logic here. Need a double-check. @ostromart

cmd/mesh/manifest-versions_test.go
elfinhe: Sure, will use test data instead.

cmd/mesh/manifest-versions_test.go
elfinhe: sure, fixed.

cmd/mesh/manifest-versions.go
ostromart: lgtm

cmd/mesh/manifest-versions.go
```diff
@@ -105,8 +103,25 @@ func getVersionCompatibleMap(versionsURI string, binVersion *goversion.Version,
...
-		l.logAndFatal("This operator version ", binVersion.String(),
-			" was not found in the global manifestVersions map.")
+		return nil, fmt.Errorf("this operator version %s was not found in the version map", binVersion.String())
+	}
+	return myVersionMap, nil
+}
+
+func loadCompatibleMapFile(versionsURI string, l *logger) (b []byte, err error) {
+	if util.IsHTTPURL(versionsURI) {
+		b, err = httprequest.Get(versionsURI)
```
ostromart: nit: it's best to avoid named return values where possible for improved readability through the smallest possible scoping, and also to remain consistent with project style. also, it's normal to redefine err as a local var where possible. 
here, i propose changing to:
if b, err :=  httprequest.Get(versionsURI); err == nil {
  return b, nil
}

elfinhe: Sure, fixed.

ostromart (approved):

PR #422, opened 2019-10-19
Bump k8s dependency to 1.16.1

hzxuzhonghu:
Bump to same dependency version as istio.io/istio


hzxuzhonghu:
This is the first step to bump k8s.io/client-go for `istio.io/istio`

hzxuzhonghu:
The err is because https://github.com/kubernetes-sigs/controller-runtime/ latest version is 0.3.0 and depend on k8s 1.15.4

hzxuzhonghu:
Solved

PR #433, opened 2019-10-22
Add upgrade short description in ARCHITECTURE.

elfinhe:
Add `upgrade` short description in ARCHITECTURE.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -248,7 +248,7 @@ subdirectory as a Cobra command with the following subcommands:
     - [diff](cmd/mesh/profile-diff.go): the diff subcommand is used to display the difference between two Istio configuration profiles.
     - [dump](cmd/mesh/profile-dump.go): the dump subcommand is used to dump the values in an Istio configuration profile.
     - [list](cmd/mesh/profile-list.go): the list subcommand is used to list available Istio configuration profiles.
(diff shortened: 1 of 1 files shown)


ARCHITECTURE.md
```diff
@@ -248,6 +248,7 @@ subdirectory as a Cobra command with the following subcommands:
     - [diff](cmd/mesh/profile-diff.go): the diff subcommand is used to display the difference between two Istio configuration profiles.
     - [dump](cmd/mesh/profile-dump.go): the dump subcommand is used to dump the values in an Istio configuration profile.
     - [list](cmd/mesh/profile-list.go): the list subcommand is used to list available Istio configuration profiles.
+- [upgrade](cmd/mesh/upgrade.go): performs an in-place upgrade of the Istio control plane. This subcommand reuses functions in [manifest](cmd/mesh/manifest.go) to generate manifests for the upgrade. Prior to executing the upgrade actions, it checks that the eligibility criteria for upgrade are met. The upgrade subcommand also checks for any changes in the profile default values between versions and alerts the user to any such changes. After applying this upgrade manifests, this subcommand has a flag (--wait) to wait for all pods to be running at the same upgraded version.
```
richardwxn: The last sentence starting from "after applying" seems to be implementation details, not sure whether it is necessary. what's the default value of wait flag, if it is true, maybe just saying it would wait for all components to run at the upgraded version? @ostromart for feedback?

ostromart: i agree, this is probably too much detail here - a one sentence summary would be sufficient. 
you could add some of this text in the upgrade file code as a top level comment if you think it's appropriate.

elfinhe: Updated. Thanks!

ostromart (approved):

PR #437, opened 2019-10-22
Use 1.5-dev tag for master

sdake:


morvencao (approved):

howardjohn:
Would suggest the `latest` tag so you don't need to update this every 3 months (which will probably forget to do)

PR #438, opened 2019-10-22
fix bool unmarshal issue.

### LINKED ISSUE #18113: istio operator fails to reconcile for some cr changes
Isti-operator will fail to reconcile if we apply below crs:

```
kind: IstioControlPlane
metadata:
spec:
  profile: demo
  trafficManagement:
    components:
      pilot:
       enabled: true
        k8s:
          hpa_spec:
            maxReplicas: 4
            metrics:
              - resource:
                  name: cpu
                  targetAverageUtilization: 80
                type: Resource
            minReplicas: 1
            scaleTargetRef:
              apiVersion: apps/v1
              kind: Deployment
              name: istio-pilot
```

2019-10-21T08:33:47.619842Z	error	Cannot unmarshal: json: cannot unmarshal object into Go value of type bool

The reason is we will change the spec as below: 
```
spec:
  profile: demo
  trafficManagement:
    components:
      pilot:
        enabled:
          value: true
        k8s:
          hpa_spec:
            maxReplicas: 4
            metrics:
              - resource:
                  name: cpu
                  targetAverageUtilization: 80
                type: Resource
            minReplicas: 1
            scaleTargetRef:
              apiVersion: apps/v1
              kind: Deployment
              name: istio-pilot
```

morvencao:
Closes: https://github.com/istio/istio/issues/18113


sdake (approved):
/lgtm

PR #441, opened 2019-10-22
Initial implementation of validation webhook

richardwxn:
Changes include:
1. Setup webhook server and register the validation path with self defined handler.
2. Implemented the validation logic towards CR admission request, which is using the validation logic of IstioControlPlaneSpec
3..Create configs for ValidatingWebhookConfiguration, Certificate/Issuer and update the corresponding operator service deployment

Update at 11/12:
Thanks for the help of @martonsereg , we update the matching rules of webhook configuration to be plural form of IstioControlPlane, otherwise the admission request would not trigger the validation call.


pkg/webhook/istiocontrolplane/server.go
```diff
@@ -0,0 +1,35 @@
...
+func Add(mgr manager.Manager) error {
+	return add(mgr)
+}
+
+func add(mgr manager.Manager) error {
+	log.Info("setting up webhook server")
+	webhookServer := mgr.GetWebhookServer()
+
+	webhookServer.Register("/validate-v1alpha2-istiocontrolplane",
+		&webhook.Admission{Handler: &iscpValidator{}})
```
rcernich: We should be setting all the webhook fileds, e.g.:
* Rules
* FailurePolicy
* Type

deploy/webhook.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: admissionregistration.k8s.io/v1beta1
```
rcernich: I don't think this is required, as it will be created automatically.  Also, I think whatever is specified here will be overridden by the webhooks registered through the code.

deploy/service.yaml
```diff
@@ -4,9 +4,12 @@ metadata:
   namespace: istio-operator
   labels:
     name: istio-operator
-  name: istio-operator-metrics
+  name: istio-operator
 spec:
   ports:
+  - port: 443
+    name: webhook
+    targetPort: 443
```
rcernich: Using `8443` will ensure the operator is capable of running without using an elevated security context.

deploy/cert.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: certmanager.k8s.io/v1alpha1
```
rcernich: Is this required?  Is there any drawback to using the built-in certificate generator?  Fewer moving parts.

pkg/webhook/istiocontrolplane/server.go
richardwxn: seems like there is no place to specify Rules/Types during registration. Can you point me to an example, I think API of controller-runtime changes quite a bit

richardwxn: this is doc I found from kubebuilder that might be related: https://book.kubebuilder.io/migration/v1vsv2.html

deploy/cert.yaml
richardwxn: I think there is no more auto generating certs support now

deploy/webhook.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: validating-webhook-configuration
+  annotations:
+    certmanager.k8s.io/inject-ca-from: istio-operator/webhook-server-cert
+webhooks:
+  - clientConfig:
+      caBundle: Cg==
+      service:
+        name: istio-operator
+        namespace: istio-operator
+        path: /validate-install-istio-io-v1alpha2-istiocontrolplane
```
morvencao: Is this correct? since you have register validate path `/validate` in webhook server.

richardwxn: because I have done multiple trials in webhook server with different combination, actually I tried using the /validate path here or change the path during registration but seems no luck

rcernich: As I mentioned previously, the webhook defined in the yaml is irrelevant.  All of this gets defined when the webhook is configured in the code.

deploy/service.yaml
richardwxn: done

deploy/webhook.yaml
richardwxn: no more self registration supported in v2

howardjohn (commented):
What is the motivation for making a Webhook? We can do open API schema validation which gets us everything we want here, with way less complexity.

Two 2 webhooks we already have (galley/injector) already cause us a ton of pain, I am very hesitant to add another unless it is absolutely needed

deploy/webhook.yaml
richardwxn: for controller-runtime v2 we have to create this configuration to replace path of previous registration

richardwxn:
@howardjohn I think the motivation here is that validation webhook can do more than openAPI schema check does(basic type, pattern, range checking). So now our custom validation logic is not doing much more than that, but it is being added more custom logic(I would add more details). Also this webhook is managed by controller-runtime manager, similar with controller, so we don't need to have extra deployment or so for it, making it simpler to maintain.

howardjohn:
@richardwxn I am not suggesting the openAPI check can do everything a webhook can do. Of course the webhook can do whatever validation it wants so its much more flexible. It is a huge pain to manage the certs and other issues involved though, so we shouldn't take it lightly to add a webhook.

OpenAPI and webhook can coexist too. In my opinion, we the best option is to start with openAPI schema (which is still useful even if we add a webhook later) and only add a webhook if there are clearly defined benefits that openAPI does not provide

ayj:
+1 to @howardjohn's suggestion to start with OpenAPI schema first and only add the webhook later if the schema isn't sufficient.

richardwxn:
@howardjohn @ayj Yep, that makes sense. I would put this PR on hold for now and move the basic schema checking part to openAPI first.

PR #447, opened 2019-10-23
update CRD correct links and add desp

linsun:
One of our users ask us what does status mean, thus triggered the PR. :)  Note the few existing links lead to 404.

```
status:
  status:
    Citadel:
      status: 2
    EgressGateway:
      status: 2
    Galley:
      status: 2
    Grafana:
      status: 2
    IngressGateway:
      status: 2
    Injector:
      status: 2
    Kiali:
      status: 2
    Pilot:
      status: 2
    Policy:
      status: 2
    Prometheus:
      status: 2
    Telemetry:
      status: 2
    Tracing:
      status: 2
    crds:
      status: 2
```

### DIFF
--- a/deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml
+++ b/deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml
@@ -38,7 +38,7 @@ spec:
             More info: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status'
           type: object
         status:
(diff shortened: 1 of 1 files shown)


morvencao (approved):

ostromart:
Thanks @linsun I've also been thinking that it may be better to change the proto field from enum (which renders as an int) into a string. Either that, or add an additional human readable status field as a string so that we satisfy both human and machine readers.

linsun:
@ostromart I think that is a good idea, I will open an issue for that, hopefully I can get to that soon.

PR #449, opened 2019-10-23
[WIP]Switch to use jsonpb for bool value

irisdingbj:
@ostromart  ^^^


ostromart:
I tried to make this work but after thinking about it a bit it's probably not possible because in any case to use jsonpb.Marshal correctly we'd need to start with an IstioControlPlane with struct tags. Using the alias type doesn't buy us anything in terms of handling things like BoolValue deeper in the struct.
I verified that we are able to unmarshal a more complex input (I used default.yaml as an overlay to the demo profile) and it worked so I think we should be ok.

PR #455, opened 2019-10-24
Fix e2e

howardjohn:


howardjohn:
I don't have time to fix this, but if some one else wants to you just need to set the correct istioctl location and fix an extra )

PR #456, opened 2019-10-24
Operator e2e Fix

irisdingbj:
bases on https://github.com/istio/operator/pull/404

@sdake I am trying to fix the e2e failure for operator controller here based on your above PR.

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -20,7 +20,7 @@ spec:
           command:
           - istio-operator
           - server
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
           resources:
             limits:
               cpu: 200m
--- a/tests/e2e/e2e.sh
+++ b/tests/e2e/e2e.sh
@@ -24,7 +24,7 @@ function setup_docker() {
   HUB=istio-testing TAG=1.5-dev make controller docker
   kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:1.5-dev
   kind --loglevel debug --name istio-testing load docker-image istio-testing/app:istio-testing
-  }
+}
 
 
 
@@ -63,6 +63,7 @@ export KUBECONFIG
 pushd "${ISTIO_DIR}" || exit
   HUB=istio-testing TAG=istio-testing make docker.app
 popd
+
 setup_docker
 
 
--- a/tests/e2e/e2e.sh
+++ b/tests/e2e/e2e.sh
@@ -17,15 +17,14 @@
 set -eux
 
 export ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
-HUB="istio-testing"
-TAG="istio-testing"
+HUB="gcr.io/istio-testing"
+TAG="latest"
 ROOT=$(cd ../../../; pwd)
 
 
 function setup_docker() {
-  HUB=gcr.io/istio-testing TAG=1.5-dev make controller docker
-  kind --loglevel debug --name istio-testing load docker-image gcr.io/istio-testing/operator:1.5-dev
-  kind --loglevel debug --name istio-testing load docker-image istio-testing/app:istio-testing
+  HUB=istio-testing TAG=1.5-dev make controller docker
+  kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:1.5-dev
 }
 
 
@@ -42,7 +41,7 @@ fi
 
 # Create an operator manifest from the default control plane configuration
 cd "${ROOT}/src/istio.io/operator"
-operator_manifest_files=( "deploy/namespace.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml" "deploy/service_account.yaml" "deploy/clusterrole.yaml" "deploy/clusterrole_binding.yaml" "deploy/service.yaml" "deploy/operator.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_cr.yaml" )
+operator_manifest_files=( "deploy/namespace.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml" "deploy/service_account.yaml" "deploy/clusterrole.yaml" "deploy/clusterrole_binding.yaml" "deploy/service.yaml" "tests/e2e/testdata/operator.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_cr.yaml" )
 
 # Generate the main manifest
 rm -f "${ISTIO_DIR}"/install/kubernetes/istio-operator.yaml
@@ -57,11 +56,10 @@ pushd "${ISTIO_DIR}"
 # shellcheck disable=SC1091
 source "./prow/lib.sh"
 setup_kind_cluster ""
-KUBECONFIG=$(kind get kubeconfig-path --name="istio-testing")
-export KUBECONFIG
-HUB="${HUB}" TAG="${TAG}" make docker.app
 popd
 
+KUBECONFIG=$(kind get kubeconfig-path --name="istio-testing")
+export KUBECONFIG
 setup_docker
 
 pushd "${ISTIO_DIR}" || exit
--- a/tests/e2e/e2e.sh
+++ b/tests/e2e/e2e.sh
@@ -18,11 +18,28 @@ set -eux
 
 export ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
 ROOT=$(cd ../../../; pwd)
+HUB=istio-testing
+TAG=istio-testing
 
 function setup_docker() {
   HUB=istio-testing TAG=1.5-dev make controller docker
   kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:1.5-dev
   }
+
+function build_kind_images() {
+  # Build just the images needed for the tests
+  for image in pilot proxyv2 app test_policybackend mixer citadel galley sidecar_injector kubectl node-agent-k8s; do
+     make docker.${image}
+  done
+
+  time load_kind_images
+}
+
+function load_kind_images() {
+	# Archived local images and load it into KinD's docker daemon
+	# Kubernetes in KinD can only access local images from its docker daemon.
+	docker images "${HUB}/*:${TAG}" --format '{{.Repository}}:{{.Tag}}' | xargs -n1 kind --loglevel debug --name istio-testing load docker-image
+}
 
 mkdir -p "${ARTIFACTS}/out"
 
@@ -60,6 +77,7 @@ setup_docker
 
 
 pushd "${ISTIO_DIR}" || exit
+  build_kind_images
   make istioctl
   HUB=istio-testing TAG=istio-testing E2E_ARGS="--use_operator --test_logs_path=${ARTIFACTS}" make e2e_simple_noauth_run
 popd
--- a/tests/e2e/e2e.sh
+++ b/tests/e2e/e2e.sh
@@ -35,8 +35,7 @@ then
 fi
 
 # Create an operator manifest from the default control plane configuration
-#cd "${ROOT}/src/istio.io/operator"
-cd /Users/iris/oti/goProjects/istio.io/operator
+cd "${ROOT}/src/istio.io/operator"
 operator_manifest_files=( "deploy/namespace.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml" "deploy/service_account.yaml" "deploy/clusterrole.yaml" "deploy/clusterrole_binding.yaml" "deploy/service.yaml" "deploy/operator.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_cr.yaml" )
 
 # Generate the main manifest
--- a/tests/e2e/e2e.sh
+++ b/tests/e2e/e2e.sh
@@ -69,5 +69,5 @@ setup_docker
 
 pushd "${ISTIO_DIR}" || exit
   make istioctl
-  HUB=istio-testing TAG=istio-testing E2E_ARGS="--use_operator --test_logs_path=${ARTIFACTS}" make e2e_simple_noauth_run
+  HUB=istio-testing TAG=istio-testing E2E_ARGS="--use_operator --use_local_cluster=true --test_logs_path=${ARTIFACTS}" make e2e_simple_noauth_run
 popd
--- a/tests/e2e/e2e.sh
+++ b/tests/e2e/e2e.sh
@@ -17,83 +17,50 @@
 set -eux
 
 export ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
-
-function cleanup_kind_cluster() {
-    kind export logs --name istio-testing "${ARTIFACTS}/kind"
-    kind delete cluster --name=istio-testing
-}
-
-function setup_kind_cluster() {
-  # Delete any previous e2e KinD cluster
-  echo "Deleting previous KinD cluster with name=istio-testing"
-  if ! (kind delete cluster --name=istio-testing) > /dev/null; then
-    echo "No existing kind cluster with name istio-testing. Continue..."
-  fi
-
-  trap cleanup_kind_cluster EXIT
-
-  # Create KinD cluster
-  if ! (kind create cluster --name=istio-testing --loglevel debug --retain --wait 30s); then
-    echo "Could not setup KinD environment. Something wrong with KinD setup. Exporting logs."
-    exit 1
-  fi
-
-  KUBECONFIG="$(kind get kubeconfig-path --name="istio-testing")"
-  export KUBECONFIG
-
-cat <<EOF > metallb-config.yaml
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  namespace: metallb-system
-  name: config
-data:
-  config: |
-    address-pools:
-    - name: default
-      protocol: layer2
-      addresses:
-      - 172.17.255.1-172.17.255.250
-EOF
-
-  kubectl apply -f https://raw.githubusercontent.com/google/metallb/v0.8.1/manifests/metallb.yaml
-  kubectl apply -f ./metallb-config.yaml
-}
+ROOT=$(cd ../../../; pwd)
 
 function setup_docker() {
-  HUB=istio-testing TAG=latest make -f Makefile.core.mk controller docker
-  kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:istio-testing
-}
-
-setup_kind_cluster
-setup_docker
+  HUB=istio-testing TAG=1.5-dev make controller docker
+  kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:1.5-dev
+  }
 
 mkdir -p "${ARTIFACTS}/out"
 
-ISTIO_DIR="${GOPATH}/src/istio.io/istio"
+ISTIO_DIR="${ROOT}/src/istio.io/istio"
 
 # Create a clone of the Istio repository
 if [[ ! -d "${ISTIO_DIR}" ]]
 then
-  git clone https://github.com/sdake/istio.git "${ISTIO_DIR}"
+  git clone https://github.com/istio/istio.git "${ISTIO_DIR}"
 fi
 
 # Create an operator manifest from the default control plane configuration
-
-operator_manifest_files=( "deploy/namespace.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_cr.yaml" "deploy/service_account.yaml" "deploy/clusterrole.yaml" "deploy/clusterrole_binding.yaml" "deploy/service.yaml" "deploy/operator.yaml" )
+#cd "${ROOT}/src/istio.io/operator"
+cd /Users/iris/oti/goProjects/istio.io/operator
+operator_manifest_files=( "deploy/namespace.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml" "deploy/service_account.yaml" "deploy/clusterrole.yaml" "deploy/clusterrole_binding.yaml" "deploy/service.yaml" "deploy/operator.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_cr.yaml" )
 
 # Generate the main manifest
 rm -f "${ISTIO_DIR}"/install/kubernetes/istio-operator.yaml
 for manifest_file in "${operator_manifest_files[@]}"
 do
 	cat "${manifest_file}" >> "${ISTIO_DIR}"/install/kubernetes/istio-operator.yaml
-	echo "---" >> "${ISTIO_DIR}"/install/kubernetes/istio-operator.yaml
 done
 
-kubectl get pods --all-namespaces -o wide
+
+#kind cluster setup
+pushd "${ISTIO_DIR}"
+# shellcheck disable=SC1091
+source "./prow/lib.sh"
+setup_kind_cluster kindest/node:v1.15.3
+popd
+
+KUBECONFIG=$(kind get kubeconfig-path --name="istio-testing")
+export KUBECONFIG
+
+setup_docker
+
 
 pushd "${ISTIO_DIR}" || exit
   make istioctl
-
-  HUB=gcr.io/istio-testing TAG=latest E2E_ARGS="--use_operator --test_logs_path=${ARTIFACTS}" make e2e_simple_run
+  HUB=istio-testing TAG=istio-testing E2E_ARGS="--use_operator --test_logs_path=${ARTIFACTS}" make e2e_simple_noauth_run
 popd
--- a/tests/e2e/e2e.sh
+++ b/tests/e2e/e2e.sh
@@ -17,12 +17,14 @@
 set -eux
 
 export ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
+HUB="istio-testing"
+TAG="istio-testing"
 ROOT=$(cd ../../../; pwd)
 
 
 function setup_docker() {
-  HUB=istio-testing TAG=1.5-dev make controller docker
-  kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:1.5-dev
+  HUB=gcr.io/istio-testing TAG=1.5-dev make controller docker
+  kind --loglevel debug --name istio-testing load docker-image gcr.io/istio-testing/operator:1.5-dev
   kind --loglevel debug --name istio-testing load docker-image istio-testing/app:istio-testing
 }
 
@@ -54,20 +56,15 @@ done
 pushd "${ISTIO_DIR}"
 # shellcheck disable=SC1091
 source "./prow/lib.sh"
-setup_kind_cluster kindest/node:v1.15.3
-popd
-
+setup_kind_cluster ""
 KUBECONFIG=$(kind get kubeconfig-path --name="istio-testing")
 export KUBECONFIG
-
-pushd "${ISTIO_DIR}" || exit
-  HUB=istio-testing TAG=istio-testing make docker.app
+HUB="${HUB}" TAG="${TAG}" make docker.app
 popd
 
 setup_docker
 
-
 pushd "${ISTIO_DIR}" || exit
   make istioctl
-  HUB=istio-testing TAG=istio-testing E2E_ARGS="--use_operator --use_local_cluster=true --test_logs_path=${ARTIFACTS}" make e2e_simple_noauth_run
+  HUB="${HUB}" TAG="${TAG}" E2E_ARGS="--use_operator --use_local_cluster=true --test_logs_path=${ARTIFACTS}" make e2e_simple_noauth_run
 popd
--- a/tests/e2e/e2e.sh
+++ b/tests/e2e/e2e.sh
@@ -18,28 +18,15 @@ set -eux
 
 export ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
 ROOT=$(cd ../../../; pwd)
-HUB=istio-testing
-TAG=istio-testing
+
 
 function setup_docker() {
   HUB=istio-testing TAG=1.5-dev make controller docker
   kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:1.5-dev
+  kind --loglevel debug --name istio-testing load docker-image istio-testing/app:istio-testing
   }
 
-function build_kind_images() {
-  # Build just the images needed for the tests
-  for image in pilot proxyv2 app test_policybackend mixer citadel galley sidecar_injector kubectl node-agent-k8s; do
-     make docker.${image}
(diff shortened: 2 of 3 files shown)


tests/e2e/e2e.sh
```diff
@@ -17,69 +17,25 @@
...
-data:
-  config: |
-    address-pools:
-    - name: default
-      protocol: layer2
-      addresses:
-      - 172.17.255.1-172.17.255.250
-EOF
-
-  kubectl apply -f https://raw.githubusercontent.com/google/metallb/v0.8.1/manifests/metallb.yaml
```
sdake: this part is suspect for prow.  I'd suggest taking a look at removal of this for the prow case (but keep for the e2e case).  Additionally I have read metallb does nt work on Mac,but does work on Linux, so we may need some host checking before using metallb and then use `-use-local` or whatever the test flag is for the actual e2e run based upon that conditional.

tests/e2e/e2e.sh
```diff
@@ -17,69 +17,25 @@
...
-  kubectl apply -f ./metallb-config.yaml
-}
+export GOPATH=$(cd ../../../; pwd)
 
 function setup_docker() {
-  HUB=istio-testing TAG=latest make -f Makefile.core.mk controller docker
-  kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:istio-testing
+  HUB=gcr.io/istio-testing TAG=1.5-dev make -f Makefile.core.mk controller docker
```
sdake: this part looks correct.

tests/e2e/e2e.sh
```diff
@@ -17,69 +17,25 @@
...
-
 mkdir -p "${ARTIFACTS}/out"
 
 ISTIO_DIR="${GOPATH}/src/istio.io/istio"
 
 # Create a clone of the Istio repository
 if [[ ! -d "${ISTIO_DIR}" ]]
 then
-  git clone https://github.com/sdake/istio.git "${ISTIO_DIR}"
+  git clone https://github.com/istio/istio.git "${ISTIO_DIR}"
```
sdake: surprised this merged with sdake as my repo.  I was using that for testing only and have since overwritten my master.  github.com/istio/istio.git contains all of the changes needed for this PR to work - so this line looks correct.

tests/e2e/e2e.sh
```diff
@@ -90,10 +46,18 @@ do
 	echo "---" >> "${ISTIO_DIR}"/install/kubernetes/istio-operator.yaml
```
sdake: please delete this echo line - it is no longer needed as I added YAML terminators in a prior PR to all of the yaml files.  I am not sure if my master PR was backported - if not it needs to be.  I will look at this now and make it so.

tests/e2e/e2e.sh
```diff
@@ -90,10 +46,18 @@ do
...
-  make istioctl
+setup_docker
 
-  HUB=gcr.io/istio-testing TAG=latest E2E_ARGS="--use_operator --test_logs_path=${ARTIFACTS}" make e2e_simple_run
-popd
+KUBECONFIG=$(kind get kubeconfig-path --name="e2e")
+export KUBECONFIG
+
+pushd "${ISTIO_DIR}" || exit
+  HUB=gcr.io/istio-testing TAG=1.5-dev E2E_ARGS="--use_operator --use_local_cluster=true --test_logs_path=${ARTIFACTS}" make e2e_simple_run
```
sdake: `--use-local-cluster=true` is fine for some conditions (not using metallb) - but I really do want to test load-balancer cases in the 2e test cases locally.  I am good with this minor change (adding metallb into the system) after we get a good baseline).

sdake: iirc make e2e_simple_run does not build istioctl - which is mandatory for the test case to run - but I could be mistaken (this is why I build it in the original work).

tests/e2e/e2e.sh
```diff
@@ -90,10 +46,18 @@ do
 	echo "---" >> "${ISTIO_DIR}"/install/kubernetes/istio-operator.yaml
 done
 
-kubectl get pods --all-namespaces -o wide
+#kind cluster setup
+pushd "${ISTIO_DIR}"
+# shellcheck disable=SC1091
+source "./prow/lib.sh"
```
sdake: not sure what this is or does.  Did you add prow/lib.sh to the repo?  If its already in repo - make sure it does what you expect it to. @howardjohn and I were speaking of the need to generalize this work so it can be reused around all satellite repos (cni and operator to be specific at first) - although that generalization work can follow on making it work for the specific case of operator - which is essential for the release.  If your struggling, I should be available for a conversation tonight on webex.

sdake (commented):
looks good - minus the loss of metallb functionality.  If removal of metallb gets things moving for e2e, can you follow up this work with a metallb implementation for folks running on Linux systems?  After that, a further generalization of this script so that it works correctly for cni is in order - perhaps you can take that work on as a followup.

Cheers
-steve

Cheers
-steve

tests/e2e/e2e.sh
sdake: Please be aware - `make e2e` should work locally on your system (if your on Linux)l.  I didn't test mac local e2e, but that must work, although Linux dev environments take priority.  `make e2e` was working locally and in prow for me prior to this submission of this PR, however, I noticed I was using master of sdake/istio - and may have had some minor changes in istio/istio that were not pushed (and are now sady lost as I have force pushed to sdake/istio)...

tests/e2e/e2e.sh
sdake: @istio/wg-environments-maintainers please take care when reviewing PRs to not **approve** PRs that contain personal git repositories.  Thewse personal git repositories may be used for testing if they are from trusted developers, however, they should not be merged and testing should not be permitted for folks new to the project (secrets can be exposed...)

Cheers
-steve

sdake (commented):
looks pretty good, some more comments inline.

tests/e2e/e2e.sh
irisdingbj: It is part of istio repo. So we can resue all the functions there.

tests/e2e/e2e.sh
irisdingbj: THis can work on my mac locally

tests/e2e/e2e.sh
```diff
@@ -17,69 +17,25 @@
 set -eux
 
 export ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
-
-function cleanup_kind_cluster() {
```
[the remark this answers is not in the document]
irisdingbj: setup_kind_cluster will take care all of this including cleanup existed kind cluster.

tests/e2e/e2e.sh
irisdingbj: Fixed!

tests/e2e/e2e.sh
```diff
@@ -17,56 +17,12 @@
...
+GOPATH=$(cd ../../../; pwd)
 
 function setup_docker() {
-  HUB=istio-testing TAG=latest make -f Makefile.core.mk controller docker
-  kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:istio-testing
-}
-
-setup_kind_cluster
-setup_docker
+  HUB=istio-testing TAG=1.5-dev make -f Makefile.core.mk controller docker
```
sdake: this should just be `make`.  Make determines if it is running in a container environment or not, and this model would break container builds.

tests/e2e/e2e.sh
```diff
@@ -75,25 +31,36 @@ ISTIO_DIR="${GOPATH}/src/istio.io/istio"
...
 if [[ ! -d "${ISTIO_DIR}" ]]
 then
-  git clone https://github.com/sdake/istio.git "${ISTIO_DIR}"
+  git clone https://github.com/istio/istio.git "${ISTIO_DIR}"
 fi
 
 # Create an operator manifest from the default control plane configuration
-
-operator_manifest_files=( "deploy/namespace.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_cr.yaml" "deploy/service_account.yaml" "deploy/clusterrole.yaml" "deploy/clusterrole_binding.yaml" "deploy/service.yaml" "deploy/operator.yaml" )
+cd "${GOPATH}/src/istio.io/operator"
```
sdake: can you explain this in more detail.  This doesn't make immediate sense.

tests/e2e/e2e.sh
```diff
@@ -75,25 +31,36 @@ ISTIO_DIR="${GOPATH}/src/istio.io/istio"
...
+operator_manifest_files=( "deploy/namespace.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_crd.yaml" "deploy/service_account.yaml" "deploy/clusterrole.yaml" "deploy/clusterrole_binding.yaml" "deploy/service.yaml" "deploy/operator.yaml" "deploy/crds/istio_v1alpha2_istiocontrolplane_cr.yaml" )
```
sdake: the ordering shouldn't matter - not sure this change is correct.

tests/e2e/e2e.sh
sdake: disagree - operator should stand alone in its dependencies on building the repository.  If we need a common files prow.ib, we can do that, but it should be merged as a separate PR in common-files repo (see link above).

tests/e2e/e2e.sh
```diff
@@ -75,25 +31,36 @@ ISTIO_DIR="${GOPATH}/src/istio.io/istio"
...
+setup_docker
+
+
 
 pushd "${ISTIO_DIR}" || exit
   make istioctl
-
-  HUB=gcr.io/istio-testing TAG=latest E2E_ARGS="--use_operator --test_logs_path=${ARTIFACTS}" make e2e_simple_run
-popd
+  HUB=istio-testing TAG=1.5-dev E2E_ARGS="--use_operator --use_local_cluster=true --test.run TestSimple1 --test_logs_path=${ARTIFACTS}" make e2e_simple_noauth_run
```
sdake: `--test.run . TestSimple1` is a subset of the test cases that should be run.  Recommend removing and fixing the deficiencies in istio/istio or this repo.

tests/e2e/e2e.sh
irisdingbj: It will fails if we deploy the icp cr a bit ahead. It complains that no crd found. By making cr deploy the last one, we can make sure no such problem happens.

tests/e2e/e2e.sh
irisdingbj: Let's first make sure our e2e can run.  This will not make the operator build dependent on istio. This is just a test dependency as we will depend on istio for e2e test anyway.

tests/e2e/e2e.sh
irisdingbj: t will fails if we deploy the icp cr a bit ahead. It complains that no crd found. By making cr deploy the last one, we can make sure no such problem happens

tests/e2e/e2e.sh
```diff
@@ -17,56 +17,12 @@
...
-    - name: default
-      protocol: layer2
-      addresses:
-      - 172.17.255.1-172.17.255.250
-EOF
-
-  kubectl apply -f https://raw.githubusercontent.com/google/metallb/v0.8.1/manifests/metallb.yaml
-  kubectl apply -f ./metallb-config.yaml
-}
+GOPATH=$(cd ../../../; pwd)
```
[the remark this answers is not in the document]
irisdingbj: rename it as `ROOT ` to prevent confusion

tests/e2e/e2e.sh
irisdingbj: Changed to `make controller docker` as we only need these two targets.

sdake: `make docker controller` are or should be dependencies of the top level `Makefile.core.mk` file so they build whenever e2e runs.  You may need to specify docker and docker.save as `PHONY` target if not already done.

tests/e2e/e2e.sh
```diff
@@ -18,12 +18,29 @@ set -eux
 
 export ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
 ROOT=$(cd ../../../; pwd)
+HUB=istio-testing
```
sdake: it is better to accept these from the `Makeile.core.mk` file directly rather than have to maintain them in two places.  On the call to `e2e.sh` i'd recommend HUB=$(HUB) TAG=$(TAG) tests/e2e/e2e.sh such that user overrides of HUB or TAG are used during the build process.

tests/e2e/e2e.sh
```diff
@@ -18,12 +18,29 @@ set -eux
...
+TAG=istio-testing
 
 function setup_docker() {
   HUB=istio-testing TAG=1.5-dev make controller docker
```
sdake: try with `HUB=$(HUB)

tests/e2e/e2e.sh
```diff
@@ -18,12 +18,29 @@ set -eux
...
   kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:1.5-dev
```
sdake: replace istio-testing/oeprator:1.5-dev with `$(HUB)/operator:$(TAG)

tests/e2e/e2e.sh
```diff
@@ -18,12 +18,29 @@ set -eux
...
   }
 
+function build_kind_images() {
```
sdake: rename function to `build_istio_images()`

tests/e2e/e2e.sh
```diff
@@ -18,12 +18,29 @@ set -eux
...
+     make docker.${image}
+  done
+
+  time load_kind_images
+}
+
+function load_kind_images() {
+	# Archived local images and load it into KinD's docker daemon
+	# Kubernetes in KinD can only access local images from its docker daemon.
+	docker images "${HUB}/*:${TAG}" --format '{{.Repository}}:{{.Tag}}' | xargs -n1 kind --loglevel debug --name istio-testing load docker-image
```
sdake: I believe this may not be necessary as it was not necessary prior for e2e testing to work.  Perhaps this was a side-effect of running the `make e2e` test target in the istio repository - but there should be no strong rationale to duplicate this functionality between repositories (it will make maintenance harder).

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,57 @@
 set -eux
 
 export ARTIFACTS="${ARTIFACTS:-$(mktemp -d)}"
+ROOT=$(cd ../../../; pwd)
```
sdake: I continue not to like this use of (cd ../../../; pwd).  There is an environment variable (IIRC) that contains `PWD` in the top level Makefile.  Can you see if that can be used?  If its not being passed to `Makefile.core.mk` - it can be easily added.

tests/e2e/e2e.sh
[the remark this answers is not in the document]
sdake: I don't see a setup_kind_cluster in this file.  It seems to have been removed.  Is this being imported from the istio repository in some way?  If so, please duplicate the work here.  While duplication is generally bad, we have a process for managing duplication consistently across repositories using https://github.com/istio/common-files.  I would far prefer this common-files approach to distribute a prow library then calling one prow library on one place and having delta code in a different place (reference cni e2e where this was copied from).

I'd like to get strong baselines (in this PR) for all of the various repositories before working on the unification of the kind handling (in a later PR).

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,57 @@
...
 
-function cleanup_kind_cluster() {
-    kind export logs --name istio-testing "${ARTIFACTS}/kind"
-    kind delete cluster --name=istio-testing
-}
-
-function setup_kind_cluster() {
-  # Delete any previous e2e KinD cluster
-  echo "Deleting previous KinD cluster with name=istio-testing"
```
sdake: imo setup shouldn't delete, it should setup.  If we need to call delete to get the system progressing, thats fine.  i recommend not confusing the function names with what the functions do.

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,57 @@
...
-  # Create KinD cluster
-  if ! (kind create cluster --name=istio-testing --loglevel debug --retain --wait 30s); then
-    echo "Could not setup KinD environment. Something wrong with KinD setup. Exporting logs."
-    exit 1
-  fi
-
-  KUBECONFIG="$(kind get kubeconfig-path --name="istio-testing")"
-  export KUBECONFIG
-
-cat <<EOF > metallb-config.yaml
```
sdake: getting metallb to work on Linux was hard, but I can revive this work later if need be.  Please open a separate PR and place a hold on it to add metlalb back to the system and I can take on reimplementation from there within the PR queue.

tests/e2e/e2e.sh
sdake: if metallb works locally on your mac without ``--use-local-cluster=true` that would be ideal.  Can you clarify your statement?

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,57 @@
...
 }
 
-setup_kind_cluster
-setup_docker
+
 
 mkdir -p "${ARTIFACTS}/out"
 
-ISTIO_DIR="${GOPATH}/src/istio.io/istio"
+ISTIO_DIR="${ROOT}/src/istio.io/istio"
```
sdake: this is definately wrong - although it may be my original work where it was wrong.

I'd recommend trying `ISTIO_DIR=$(mktemp -d)`.  If that fails because of problems in Istio, we will need to work through that problem (we can do that after this PR merges) because hardcode paths in the build system are undesireable.

sdake: Further following this up - the git clone below could negatively interact with a local copy of istio (I recognize a test for the directory is done) but this would result in $ISTIO/istio being used from the local filesystem if the developer has work in that location.  This is highly undesirable for local test modes.

tests/e2e/e2e.sh
sdake: thats bad news - but not unexpected.  Thanks.

tests/e2e/e2e.sh
sdake: see review for more information about why using a prow library is an undesirable approach.  As painful as it may be, please copy the parts you need into this e2e.sh script.  As followup work a refactor of the e2e for the various repos (istio/operator, istio/cni, istio/istio) can be done in https://github.com/istio/common-files.

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,57 @@
...
+source "./prow/lib.sh"
+setup_kind_cluster kindest/node:v1.15.3
+popd
+
+KUBECONFIG=$(kind get kubeconfig-path --name="istio-testing")
+export KUBECONFIG
 
 pushd "${ISTIO_DIR}" || exit
-  make istioctl
+  HUB=istio-testing TAG=istio-testing make docker.app
```
sdake: this should use the same environment variables as the rest of the build system `HUB=$(HUB) TAG=$(TAG) make docker.app`

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,57 @@
...
+popd
+
+setup_docker
+
 
-  HUB=gcr.io/istio-testing TAG=latest E2E_ARGS="--use_operator --test_logs_path=${ARTIFACTS}" make e2e_simple_run
+pushd "${ISTIO_DIR}" || exit
+  make istioctl
+  HUB=istio-testing TAG=istio-testing E2E_ARGS="--use_operator --use_local_cluster=true --test_logs_path=${ARTIFACTS}" make e2e_simple_noauth_run
```
sdake: see comment on line 64 about hub and tag usage.

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,57 @@
...
+setup_kind_cluster kindest/node:v1.15.3
+popd
+
+KUBECONFIG=$(kind get kubeconfig-path --name="istio-testing")
+export KUBECONFIG
 
 pushd "${ISTIO_DIR}" || exit
-  make istioctl
+  HUB=istio-testing TAG=istio-testing make docker.app
+popd
```
sdake: why pop to push again?  Is the need to call setup_docker?

sdake (commented):
implementation looks ok.  There are some problems around HUB,TAG variable usage and the git cloing specifically.  If you tell me it wont work with a git clone in a temp dir, we can merge without that.

I'd really strongly like to avoid the usage of the prow library - but instead copy the relevant bits into this script so we can make a unified e2e script in common-tools as a followup PR.

sdake:
@howardjohn would you mind reviewing?

tests/e2e/e2e.sh
howardjohn: It needs to be `src/istio.io/istio`, since prow ~~will~~ should check out istio/istio for us instead of doing the clone here. 
And to your other concern, I think the clone only happens if the folder doesn't exist

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,57 @@
...
-	echo "---" >> "${ISTIO_DIR}"/install/kubernetes/istio-operator.yaml
 done
 
-kubectl get pods --all-namespaces -o wide
+
+#kind cluster setup
+pushd "${ISTIO_DIR}"
+# shellcheck disable=SC1091
+source "./prow/lib.sh"
+setup_kind_cluster kindest/node:v1.15.3
```
howardjohn: I don't think we should explicitly set the node here, or this will break when we update kind. Really, any of this can break if we update ./prow/lib.sh so its already sort of at risk

howardjohn (commented):
This is pretty misleading, you are not actually testing the PR

You set the HUB and TAG, build docker images, but they are **not** used.

When you pass HUB and TAG to the test, it is not used because the operator part of the test doesn't support it. 

When you render the operator install yaml, it has the hardcoded hub and tag: https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_operator/456/e2e_operator/115/artifacts/simple-auth-test-a9a5b04cf9db4c37a5fe/yaml/istio-operator.yaml

So anything you are doing with docker and hubs and tags is entirely unused as far as I can tell

tests/e2e/e2e.sh
sdake: ok thats fair.

tests/e2e/e2e.sh
sdake: @howardjohn right, but if the repo does exist, then the e2e is done based upon the local version of Istio?  I agree with your other points.  Not sure what to do with the prefix `$(ROOT}`

tests/e2e/e2e.sh
sdake: right, my suggestion was to keep this per repo - and then follow this work up with a common-files replication.  agree re node, we should bake node into the buildtools image imo.

howardjohn:
+1 We should bake the default node image into the build. I tried
unsuccessfully

irisdingbj:
> This is pretty misleading, you are not actually testing the PR
> 
> You set the HUB and TAG, build docker images, but they are **not** used.
> 
> When you pass HUB and TAG to the test, it is not used because the operator part of the test doesn't support it.
> 
> When you render the operator install yaml, it has the hardcoded hub and tag: https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_operator/456/e2e_operator/115/artifacts/simple-auth-test-a9a5b04cf9db4c37a5fe/yaml/istio-operator.yaml
> 
> So anything you are doing with docker and hubs and tags is entirely unused as far as I can tell

good catch! Thanks. I will fix this.

tests/e2e/e2e.sh
irisdingbj: removed already

tests/e2e/e2e.sh
irisdingbj: removed the node images here

tests/e2e/e2e.sh
irisdingbj: revised

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,54 @@
...
-  kubectl apply -f https://raw.githubusercontent.com/google/metallb/v0.8.1/manifests/metallb.yaml
-  kubectl apply -f ./metallb-config.yaml
-}
 
 function setup_docker() {
-  HUB=istio-testing TAG=latest make -f Makefile.core.mk controller docker
-  kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:istio-testing
+  HUB=gcr.io/istio-testing TAG=1.5-dev make controller docker
+  kind --loglevel debug --name istio-testing load docker-image gcr.io/istio-testing/operator:1.5-dev
+  kind --loglevel debug --name istio-testing load docker-image istio-testing/app:istio-testing
```
howardjohn: where is app coming from?

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,54 @@
...
-EOF
-
-  kubectl apply -f https://raw.githubusercontent.com/google/metallb/v0.8.1/manifests/metallb.yaml
-  kubectl apply -f ./metallb-config.yaml
-}
 
 function setup_docker() {
-  HUB=istio-testing TAG=latest make -f Makefile.core.mk controller docker
-  kind --loglevel debug --name istio-testing load docker-image istio-testing/operator:istio-testing
+  HUB=gcr.io/istio-testing TAG=1.5-dev make controller docker
```
howardjohn: the hub should not be a real hub, or jts impossible to tell whether we are using the local image or accidentally using a published image

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,54 @@
...
+HUB="${HUB}" TAG="${TAG}" make docker.app
+popd
+
+setup_docker
 
 pushd "${ISTIO_DIR}" || exit
   make istioctl
-
-  HUB=gcr.io/istio-testing TAG=latest E2E_ARGS="--use_operator --test_logs_path=${ARTIFACTS}" make e2e_simple_run
+  HUB="${HUB}" TAG="${TAG}" E2E_ARGS="--use_operator --use_local_cluster=true --test_logs_path=${ARTIFACTS}" make e2e_simple_noauth_run
```
howardjohn: nit - we should probably using gcr for the hub for all Istio components, and then the local image for the operator. We aren't testing istio/istio here so no need to build those images

tests/e2e/e2e.sh
howardjohn: nevermind, I see we build it later, but probably don't need to

howardjohn (commented):
looks good, just a couple minor comments. I think this is in a good state though

tests/e2e/e2e.sh
irisdingbj: we need this for simple testing, especially the testNgix part

tests/e2e/e2e.sh
irisdingbj: just need to build istio-testing/app:istio-testing as we will not build this images during normal istio build

tests/e2e/e2e.sh
irisdingbj: Need to set this as a real hub otherwise we need to modify deploy/operator.yaml..

irisdingbj:
> looks good, just a couple minor comments. I think this is in a good state though

Thanks @howardjohn seems it fails for `istioctl kube-inject this time..` not sure the reason yet

tests/e2e/e2e.sh
howardjohn: I think We should not be building it, we should use gcr.io/istio-testing/app:latest. We aren't building pilot or galley, so why build the test app?

tests/e2e/e2e.sh
howardjohn: I think it's better to modify that. Being able to change the hub/tag is a requirement for users, so we need to support it some way or another.

Otherwise, it's likely someone changes the image pull policy of something else that causes us to not use the local image and we don't notice?

howardjohn:
the istioctl inject is not related to this pr, I have a PR on istio/istio to fix it

tests/e2e/e2e.sh
irisdingbj: agree!  Let me put the operator.yaml one copy into testdata

tests/e2e/e2e.sh
irisdingbj: > the istioctl inject is not related to this pr, I have a PR on istio/istio to fix it

Thanks for the info!

irisdingbj: changed to use gcr.io/istio-testing/app:latest and skip building istio-testing/app

irisdingbj:
https://github.com/istio/istio/pull/18352

deploy/operator.yaml
```diff
@@ -20,7 +20,7 @@ spec:
           command:
           - istio-operator
           - server
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
```
sdake: @ostromart I believe I like Always as a pull policy better.  I understand why this PR changes the default pull policy.  If Always is specified, docker will pull the image in all cases, even though it may not be rermotely pushed.

A sed operation on this line may make sense.  Do you have any thoughts on what is bettter?

tests/e2e/e2e.sh
sdake: will address in a followon PR if we can get this PR merged today.

tests/e2e/e2e.sh
sdake: @howardjohn can you clarify your suggestion here?

tests/e2e/e2e.sh
```diff
@@ -17,83 +17,52 @@
...
+
+KUBECONFIG=$(kind get kubeconfig-path --name="istio-testing")
+export KUBECONFIG
+setup_docker
 
 pushd "${ISTIO_DIR}" || exit
   make istioctl
-
-  HUB=gcr.io/istio-testing TAG=latest E2E_ARGS="--use_operator --test_logs_path=${ARTIFACTS}" make e2e_simple_run
+  HUB="${HUB}" TAG="${TAG}" E2E_ARGS="--use_operator --use_local_cluster=true --test_logs_path=${ARTIFACTS}" make e2e_simple_noauth_run
```
sdake: this HUB in use doesn't match the HUB at line 26.

tests/e2e/testdata/operator.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: apps/v1
```
sdake: why not use sed to generate this file on the fly as to avoid duplication of the manifests which may become inconsistent over time?

sdake (approved):
This is pretty close and I feel comfortable merging assuming the unresolved issues are addressed in a followup iterative PR.  Assuming this PR can pass the checks, we can merge it now and follow up with fixes for the outstanding issues.  If this PR still blocks in the checks, @irisdingbj can you get it passing as soon as viable?

sdake:
groan - I had forgot the e2e tests were non-voting.  I'll take a look at fixing this up today - if not @irisdingbj - can you take a look at fixing when you start your day.

Cheers
-steve

PR #457, opened 2019-10-24
Clean up manifest apply output

ostromart:
This cleans up a bunch of output that should only go to the log, not be visible to the user during normal operation.


cmd/mesh/manifest-apply.go
```diff
@@ -67,12 +67,6 @@ func manifestApplyCmd(rootArgs *rootArgs, maArgs *manifestApplyArgs) *cobra.Comm
 		Run: func(cmd *cobra.Command, args []string) {
 			// Passing cmd.OutOrStdXXX() allows capturing command output for e2e tests.
 			l := newLogger(rootArgs.logToStdErr, cmd.OutOrStdout(), cmd.OutOrStderr())
-			if !maArgs.skipConfirmation && maArgs.kubeConfigPath == "" && maArgs.context == "" {
-				if !confirm("Do you want to proceed? (y/N)", cmd.OutOrStdout()) {
-					cmd.Print("Cancelled.\n")
-					os.Exit(1)
-				}
-			}
```
ayj: Intentional?

ostromart: Yes, it's odd (and annoying) to have this prompt before apply. It was added as an interim step when our confidence level in it was not very high. 
istioctl manifest apply should work just like kubectl apply imo - we should reserve confirmation prompts for cases like upgrade default parameters changing.

ayj: I mostly agree. To work just like kubectl I think we need something like the following:

```bash
$ kubectl apply
error: must specify one of -f and -k

$ istioctl manifest apply
error: must specify one of -f or --use-default-profile
```

Otherwise the implicit _what_ that is being applied is not completely clear to new users.

ostromart: I agree with your point, but --use-default-profile seems excessive. 
How about, if -f is not specified, then we put up a prompt: 
"This will install the default Istio profile into the cluster. Proceed? [y|N]"

ayj: SGTM

ostromart: Done.

ayj (approved):

PR #458, opened 2019-10-24
Show reconcile status as string

ostromart:
Enums render as int, this gives a more human readable output:

status:
  status:
    Citadel:
      status: 2
      statusString: HEALTHY
    EgressGateway:
      status: 2
      statusString: HEALTHY
...


linsun (approved):
Nice @ostromart !  didn't realize this would be rendered with an extra string

sdake:
@ostromart would approve this PR although the checks are failing badly.

ostromart:
There's something messed up with lint...

PR #474, opened 2019-10-28
Fix reverse translate issue

richardwxn:
Fix for https://github.com/istio/istio/issues/18289
Also as a TODO, create https://github.com/istio/istio/issues/18392 to track update of the reverse translation table

### DIFF
--- a/cmd/mesh/testdata/manifest-generate/output/pilot_override_kubernetes.yaml
+++ b/cmd/mesh/testdata/manifest-generate/output/pilot_override_kubernetes.yaml
@@ -18,6 +18,29 @@
 
 # NodeAgent component is disabled.
 
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: pilot
+    istio: pilot
+    release: istio
+  name: istio-pilot
+  namespace: istio-control
+spec:
+  ports:
+  - name: grpc-xds
+    port: 11111
+  - name: https-xds
+    port: 15011
+  - name: http-legacy-discovery
+    port: 8080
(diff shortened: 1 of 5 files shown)


cmd/mesh/testdata/manifest-generate/output/pilot_override_kubernetes.yaml
```diff
@@ -18,6 +18,29 @@
 
 # NodeAgent component is disabled.
 
+apiVersion: v1
```
ostromart: why are there such huge diffs here? it seems like some resources are reordered.

richardwxn: yep I think it doesn't makes sense, but as I said here: https://github.com/istio/operator/pull/482#pullrequestreview-308998258, seems like the output order is not stable

pkg/translate/translate_value.go
```diff
@@ -532,10 +532,17 @@ func (t *ReverseTranslator) isEnablementPath(path util.Path) bool {
 	if len(path) < 2 || path[len(path)-1] != "enabled" {
 		return false
 	}
+
+	pstr := path.String()
+	if pstr == "mixer.policy.enabled" || pstr == "mixer.telemetry.enebaled" ||
```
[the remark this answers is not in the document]
richardwxn: done

richardwxn:
This one needs to get in soon as it fixes the breaking migration tools, can you check again plz @ostromart

ostromart (approved):

PR #475, opened 2019-10-28
use hub and tag from istio/pkg DockerBuildInfo

carolynhu:
https://github.com/istio/istio/issues/18173


howardjohn (commented):
the code change looks good, but you added 3000 junk files. We don't use vendor, shouldn't check in .DS_store

howardjohn (commented):
wasn't there a test for this, that would need to be update?

carolynhu:
yes, that test was not merged. We should update that test also. I'm going to make that commit

howardjohn (approved):

carolynhu:
replace `github.com/spf13/viper` to `github.com/istio/viper` for fixing: 
```
# istio.io/pkg/collateral
../../../pkg/mod/istio.io/pkg@v0.0.0-20191029184635-5c2f5ef63692/collateral/control.go:357:35: v.GetAliases undefined (type *viper.Viper has no field or method GetAliases)
```

carolynhu:
remember to cherry-pick this PR to 1.4 branch

PR #476, opened 2019-10-28
Exclude the overridden values when comparing the old with the new ones.

### LINKED ISSUE #17916: Istioctl x upgrade: exclude the overridden values when comparing the old with the new ones
In `func checkUpgradeValues(curValues string, tarValues string, l *logger)`, before we run `compare.YAMLCmp(curValues, tarValues)` we should exclude the overridden values specified in profile yaml / customized yaml, only showing up the unexpected changes.

elfinhe:
Resolve: https://github.com/istio/istio/issues/17916

When user explicitly inputs an IstioControlPlaneSpec CR file, do not output the overridden value changes by default. Also add --show-overrides flag to show all changed values.

### DIFF
--- a/cmd/mesh/profile-dump.go
+++ b/cmd/mesh/profile-dump.go
@@ -177,7 +177,7 @@ func genICPS(inFilename, profile, setOverlayYAML string, force bool, l *logger) 
 	return finalYAML, finalICPS, nil
 }
 
-func genOverlayICPS(filename string) (string, *v1alpha2.IstioControlPlaneSpec, error) {
+func genOverlayICPS(filename string, force bool) (string, *v1alpha2.IstioControlPlaneSpec, error) {
 	if filename == "" {
 		return "", nil, nil
 	}
@@ -186,7 +186,7 @@ func genOverlayICPS(filename string) (string, *v1alpha2.IstioControlPlaneSpec, e
 	if err != nil {
 		return "", nil, fmt.Errorf("could not read from file %s: %s", filename, err)
 	}
-	overlayICPS, _, err := unmarshalAndValidateICP(string(b))
+	overlayICPS, _, err := unmarshalAndValidateICP(string(b), force)
 	if err != nil {
 		return "", nil, err
 	}
--- a/cmd/mesh/profile-dump.go
+++ b/cmd/mesh/profile-dump.go
@@ -18,7 +18,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
-
+	"strings"
 	"text/template"
 
 	"github.com/ghodss/yaml"
@@ -175,6 +175,33 @@ func genICPS(inFilename, profile, setOverlayYAML string, force bool, l *logger) 
 		return "", nil, err
 	}
 	return finalYAML, finalICPS, nil
+}
+
+func genOverlayICPS(filename string) (string, *v1alpha2.IstioControlPlaneSpec, error) {
+	if strings.TrimSpace(filename) == "" {
+		return "", nil, nil
+	}
+
+	b, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return "", nil, fmt.Errorf("could not read from file %s: %s", filename, err)
+	}
+	overlayICPS, _, err := unmarshalAndValidateICP(string(b))
+	if err != nil {
+		return "", nil, err
+	}
+
+	t, err := translate.NewTranslator(binversion.OperatorBinaryVersion.MinorVersion)
+	if err != nil {
+		return "", nil, err
+	}
+
+	overlayYAML, err := component.TranslateHelmValues(overlayICPS, t, "")
+	if err != nil {
+		return "", nil, err
+	}
+
+	return overlayYAML, overlayICPS, nil
 }
 
 func genProfile(helmValues bool, inFilename, profile, setOverlayYAML, configPath string, force bool, l *logger) (string, error) {
--- a/cmd/mesh/profile-dump.go
+++ b/cmd/mesh/profile-dump.go
@@ -18,7 +18,6 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
-	"strings"
 	"text/template"
 
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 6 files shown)


cmd/mesh/profile-dump.go
```diff
@@ -173,6 +174,33 @@ func genICPS(inFilename, profile, setOverlayYAML string, force bool, l *logger)
 	return finalYAML, finalICPS, nil
 }
 
+func genOverlayICPS(filename string) (string, *v1alpha2.IstioControlPlaneSpec, error) {
+	if strings.TrimSpace(filename) == "" {
```
ostromart: Good idea to do this but could you move it higher in the call stack so all functions get the trimmed version?

cmd/mesh/profile-dump.go
```diff
@@ -173,6 +174,33 @@ func genICPS(inFilename, profile, setOverlayYAML string, force bool, l *logger)
...
+	if err != nil {
+		return "", nil, err
+	}
+
+	t, err := translate.NewTranslator(binversion.OperatorBinaryVersion.MinorVersion)
+	if err != nil {
+		return "", nil, err
+	}
+
+	overlayYAML, err := component.TranslateHelmValues(overlayICPS, t, "")
```
ostromart: i think this may cause more problems than help. I suggest just using the untranslated input as a mask.

cmd/mesh/upgrade.go
```diff
@@ -54,20 +55,24 @@ type upgradeArgs struct {
 	skipConfirmation bool
 	// force means directly applying the upgrade without eligibility checks.
 	force bool
-	// versionsURI is a URI pointing to a YAML formatted versions mapping.
-	versionsURI string
+	// showOverrides shows all changed values even if they are specified in inFilename.
```
ostromart: I don't think we need this, it's not useful enough to warrant an extra option. If users are interested in how defaults are changing between versions, they should use profile diff.

pkg/compare/compare.go
```diff
@@ -146,6 +151,59 @@ func UnmarshalInlineYaml(obj map[string]interface{}, targetPath string) (err err
...
+		return nil, err
+	}
+	return cmp.FilterPath(func(curPath cmp.Path) bool {
+		up := pathToStringList(curPath)
+		found, _ := IsPathInTree(tree, up)
+		return found
+	}, cmp.Ignore()), nil
+}
+
+func IsPathInTree(tree map[string]interface{}, path []string) (bool, error) {
```
ostromart: does it need to be public?

pkg/compare/compare.go
```diff
@@ -146,6 +151,59 @@ func UnmarshalInlineYaml(obj map[string]interface{}, targetPath string) (err err
...
+	// Clone a slice: https://github.com/go101/go101/wiki
+	remainPath := append(path[:0:0], path...)
+	return isPathInTree(tree, remainPath)
+}
+
+func isPathInTree(tree map[string]interface{}, remainPath []string) (bool, error) {
+	if len(remainPath) == 0 {
+		return false, nil
+	}
+	for key, val := range tree {
```
ostromart: could you repurpose one of the existing tree traversals (e.g. read node) rather than copying the code here?

pkg/compare/compare_test.go
```diff
@@ -656,7 +656,105 @@ metadata:
...
+	}
+}
+
+func TestYAMLCmpWithIgnoreTree(t *testing.T) {
+	tests := []struct {
+		desc string
+		a    string
+		b    string
+		i    []string
+		y    string
```
ostromart: y is not very intuitive. maybe call it mask or something?

pkg/compare/compare_test.go
```diff
@@ -656,7 +656,105 @@ metadata:
...
+kind: IstioControlPlane`,
+			want: `spec:
+  hub: gcr.io/istio-testing -> docker.io/istio
+  tag: latest -> 1.3.3
+`,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			if got, want := YAMLCmpWithIgnore(tt.a, tt.b, tt.i, tt.y), tt.want; !(got == want) {
```
ostromart: since you're using YAMLCmpWithIgnore you don't need to use ICP structs as input. the tests would be more readable if you make the inputs much simpler e.g. a:b:c

cmd/mesh/profile-dump.go
elfinhe: Fixed.

cmd/mesh/profile-dump.go
elfinhe: I am not 100% sure this is the right thing to do. Let me take a look at how the translator works.

elfinhe: You are right. I will fix this.

cmd/mesh/upgrade.go
elfinhe: This option provides an overview of all changes will be applied by upgrade. And the main reason of adding this option is to let users check unexpected changes made even by their CR yaml (mostly by using CLI).

pkg/compare/compare.go
elfinhe: Removed this func since this algorithm does not mutate path.

elfinhe: And change it into private.

pkg/compare/compare.go
elfinhe: Because the potential impact of repurposing tree traversal func is big. We need to do that in a separate round of refactoring and we already had an issue to track it: https://github.com/istio/istio/issues/15494.

pkg/compare/compare_test.go
elfinhe: fixed, thanks!

pkg/compare/compare_test.go
elfinhe: Renamed the input, we are not using ICP as input, this one is ignore. Misleading names.
thanks.

cmd/mesh/upgrade.go
elfinhe: Discussed offline, decide to remove it as it is not required for now, and there is an existing tool (profile diff) can do the same job.

ostromart (approved):
LG, thanks!

elfinhe:
cannot reproduce error in my up-to-date local repo

PR #478, opened 2019-10-29
Add multicluster remote gateway install YAML for mesh command

ostromart:
This is used in the instructions for setting up remote clusters.


data/examples/multicluster/istio-gateway.yaml
```diff
@@ -0,0 +1,107 @@
...
+    enabled: false
+
+  policy:
+    enabled: false
+
+  telemetry:
+    enabled: false
+
+  security:
+    enabled: false
```
ayj: Citadel is required for mTLS.

ostromart: This was actually totally wrong due to too many tasks going on in parallel. 
PTAL at the updated values which should correspond to directions here:
https://preliminary.istio.io/docs/setup/install/multicluster/gateways/#deploy-the-istio-control-plane-in-each-cluster
and here:
https://preliminary.istio.io/docs/setup/install/multicluster/shared-gateways/#setup-cluster-1-primary

ostromart:
@ayj could you ptal?

ayj (approved):

PR #480, opened 2019-10-29
Properly format yaml messages

sdake:
As per YAML specifications, YAML documents should begin with `---`
and conclude with `...`. Kubernetes accepts pretty sloppy yaml and
just requires a `---` although other tools, such as `yq` require
properly formatted YAML in order to operate correctly.


morvencao (approved):
/lgtm

PR #481, opened 2019-10-29
update sha for installer to include automtls change

incfly:
istio/istio#14524

some test in ci is blocked on this.


incfly:
https://github.com/istio/installer/commit/e8b6e534c78ed846d35ac46d1d88529e827aa7f0

howardjohn (commented):
not super familiar with this, but I think there should be something about auto mtls here...

incfly:
i did a run for the `make gen` and this asset.go is the only diff i got...

PR #482, opened 2019-10-29
update sha for installer to include automtls change

incfly:
istio/istio#14524


incfly:
@howardjohn @ostromart I manually modified the default.yaml now. All tests passed, PTAL, thanks.

richardwxn (commented):
About common file changes, just curious what is the current workflow to update the LICENSE files, so we need to manually update common occasionally from different repos? @geeknoid 

also we might need to generate more consistent output of the manifest cmd test, otherwise even one field's change would lead to a really large PR because of golden file updates, which make it hard to verify. @ostromart @sdake

incfly:
They are the outcome of the make gen.

Steve updated the CI test automation to make it aligned CI and dev env.

ostromart (approved):

PR #499, opened 2019-11-01
Fix cryptic messages.

### LINKED ISSUE #18521: 'istioctl x upgrade' fails with cryptic message when using 1.4 beta 1
Steps:
- Start with cluster with Istio 1.3
- Run `'istioctl x upgrade` using Istio 1.4 Beta 1

Result: it fails with `Error: the target version 1.4-dev in  is not supported by istioctl 1.4.0, please download istioctl 1.4-dev and run upgrade again`.

A few issues:

- Ideally, it should work
- If it really cannot, it should give a clear message why. Right now, the message seems to be missing a word  (`in  is`), and as a user I have no idea what to do next

### LINKED ISSUE #18523: 'istioctl x upgrade --help' refers to the 'mesh' command
Using 1.4 Beta 1, `istioctl x upgrade --help` includes:

    Examples:
    mesh upgrade

It's referring to a `mesh` command instead of the `istioctl` that users actually use.

elfinhe:
Fix: https://github.com/istio/istio/issues/18521
Fix: https://github.com/istio/istio/issues/18523

Also add some messages of how to upgrade sidecars at the end of a successful upgrade.

### DIFF
--- a/cmd/mesh/upgrade.go
+++ b/cmd/mesh/upgrade.go
@@ -39,6 +39,14 @@ const (
 	// The maximum number of attempts that the command will check for the upgrade completion,
 	// which means only the target version exist and the old version pods have been terminated.
 	upgradeWaitCheckVerMaxAttempts = 60
+
+	// This message provide the guide of how to upgrade Istio data plane
+	upgradeSidecarMessage = "To upgrade the Istio data plane, you will need to re-inject it.\n" +
+		"If you’re using automatic sidecar injection, you can upgrade the sidecar by doing a rolling" +
+		" update for all the pods:\n" +
+		"    kubectl rollout restart deployment --namespace $NAMESPACE_WITH_AUTO_INJECTION\n" +
+		"If you’re using manual injection, you can upgrade the sidecar by executing:\n" +
+		"    kubectl apply -f <(istioctl kube-inject -f $ORIGINAL_DEPLOYMENT_YAML)"
 )
 
 type upgradeArgs struct {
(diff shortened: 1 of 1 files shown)


cmd/mesh/upgrade.go
```diff
@@ -39,6 +39,14 @@ const (
 	// The maximum number of attempts that the command will check for the upgrade completion,
 	// which means only the target version exist and the old version pods have been terminated.
 	upgradeWaitCheckVerMaxAttempts = 60
+
+	// This message provide the guide of how to upgrade Istio data plane
+	upgradeSidecarMessage = "To upgrade the Istio data plane, you will need to re-inject it.\n" +
+		"If you’re using automatic sidecar injection, you can upgrade the sidecar by doing a rolling" +
+		" update for all the pods:\n" +
+		"    kubectl rollout restart deployment --namespace $NAMESPACE_WITH_AUTO_INJECTION\n" +
```
ostromart: $NAMESPACE_WITH_AUTO_INJECTION is a bit confusing - better to represent this as <namespace with auto injection>. Same for $ORIGINAL_DEPLOYMENT_YAML (e.g. <customization YAML file you used to install Istio>

ostromart (approved):

cmd/mesh/upgrade.go
```diff
@@ -39,6 +39,14 @@ const (
 	// The maximum number of attempts that the command will check for the upgrade completion,
 	// which means only the target version exist and the old version pods have been terminated.
 	upgradeWaitCheckVerMaxAttempts = 60
+
+	// This message provide the guide of how to upgrade Istio data plane
+	upgradeSidecarMessage = "To upgrade the Istio data plane, you will need to re-inject it.\n" +
+		"If you’re using automatic sidecar injection, you can upgrade the sidecar by doing a rolling" +
+		" update for all the pods:\n" +
+		"    kubectl rollout restart deployment --namespace <namespace with auto injection>\n" +
```
davidebbo: `kubectl rollout restart deployment` does seem to exist for me, and I have the latest kubectl that comes with `gcloud components install kubectl`. What am I doing wrong?

davidebbo: Ok, if I download the latest non-gcloud, it has the command so it must be very new. It might confused some users if many still have older versions.

PR #502, opened 2019-11-01
cleanup unused kiali chart customizer.

morvencao:
Currently we don't need the kiali chart customizer in the rendering lifecycle, clean this up to reduce the reconcile time.

### DIFF
--- a/pkg/controller/istiocontrolplane/listeners.go
+++ b/pkg/controller/istiocontrolplane/listeners.go
@@ -116,13 +116,11 @@ type IstioChartCustomizerFactory struct {
 var _ helmreconciler.ChartCustomizerFactory = &IstioChartCustomizerFactory{}
 
 // NewChartCustomizer returns a new ChartCustomizer for the specific chart.
-// Currently, an IstioDefaultChartCustomizer is returned for all charts except: kiali
+// Currently, an IstioDefaultChartCustomizer is returned for all charts except: citadel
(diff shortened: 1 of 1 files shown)


ostromart:
@rcernich PTAL if this is ok.

pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -327,49 +325,3 @@ func (c *CitadelChartCustomizer) cleanCitadelResources(obj runtime.Object) error
 	}
 	return nil
 }
-
-// KialiChartCustomizer is a ChartCustomizer for the kiali chart
```
sdake: we probably no longer want to customize KialiChart- but KialiOperator may desire customization...

sdake (commented):
not sure this PR is needed.  The Kiali chart is pending deletion - instead this PR may want to focus on the Kiali operator implementation.

rcernich:
The intent when I initially added these hooks was to populate the endpoint details for jaeger and grafana, so they would be configured automatically, without user intervention, e.g. https://github.com/Maistra/istio-operator/blob/maistra-1.1/pkg/controller/servicemesh/controlplane/hooks.go#L62-L181.  The same should apply when using the kiali CR (which is what the linked code modifies).

morvencao:
thanks @rcernich Let's keep it.
/close

PR #500, opened 2019-11-01
Turn off DNS certificate provisioning

lei-tang:
This PR is for the issue: https://github.com/istio/istio/issues/18456

PR #505, opened 2019-11-01
Cherrypick "Turn off DNS certificate provisioning" (#500)

lei-tang:
Cherrypick "Turn off DNS certificate provisioning" (https://github.com/istio/operator/pull/500).

BUILD_WITH_CONTAINER=0
make update-goldens


installer.sha
```diff
@@ -1 +1 @@
-6bf329f3ff088fa56e537afecb31938f49800978
\ No newline at end of file
+a4241599ca243b2cc6bdfd28fb45c26709143438
```
ericvn: This sha is a _master_ sha. You'll want to use a release-1.4 sha. I and @sdake tried using a6e4a4b3087ffd1bce5b186c0e3bd49cf2ee40a0 yesterday (latest one), but ran into mandiff errors we hoped this cherry pick would fix.

ericvn (changes requested):

installer.sha
lei-tang: I pushed a new commit with sha changed to a6e4a4b3087ffd1bce5b186c0e3bd49cf2ee40a0 and
"make update-goldens".

ericvn: OK. So that looks good.

ericvn (approved):

PR #506, opened 2019-11-01
Add force flag for set override and update CNI schema

### LINKED ISSUE #18561: `istioctl manifest` does not work with CNI
```
$ istioctl manifest generate --set profile=demo --set cni.enabled=true --set cni.components.cni.enabled=true --set values.cni.cniBinDir=/var/lib/cni/bin --set values.cni.cniConfDir=/var/run/multus/cni/net.d

bad path=value (values.cni.cniBinDir=/var/lib/cni/bin): unknown field "cni" in v1alpha1.Values
```

This is istioctl from 1.4:
```
$ istioctl version --short=false --remote=false
version.BuildInfo{Version:"69dffe00dec378e2a9a24e42bc36746f26f3c9a6", GitRevision:"69dffe00dec378e2a9a24e42bc36746f26f3c9a6", GolangVersion:"go1.13.3", BuildStatus:"Clean", GitTag:"1.4.0-beta.1-19-g69dffe00d"}
```

richardwxn:
resolve: https://github.com/istio/istio/issues/18561

changes include:
1. add force flag in set override logic, so user can apply --force to ignore validation error for values schema.
2. update outdated values schema for CNI part.


pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -52,6 +52,26 @@ message ArchConfig {
 message CNIConfig {
   // Controls whether CNI is enabled.
   google.protobuf.BoolValue enabled = 1;
+
```
ostromart: Add a test to check that bad CNI field causes error, good one passes?

richardwxn: added unit test here, would create separate PR for fully validation e2e test

richardwxn:
not sure why the gen check keeps failing even though there is no local update when run make gen..

jmazzitelli:
@richardwxn now you have conflicts :-/

ostromart:
@richardwxn gencheck is probably failing because you have to do make clean gen. Give that a try.
Otherwise lgtm.

ostromart (approved):

jmazzitelli:
Are these messages about conflicts an indication of something wrong in the 1.4 branch?

richardwxn:
@jmazzitelli actually not, the conflicts come from test data output, see https://github.com/istio/istio/issues/18564.  I am fixing that, hopefully the output manifests order would be stable after that.

PR #509, opened 2019-11-02
Stabilize k8s overlay output order

richardwxn:
To solve: https://github.com/istio/istio/issues/18564

Try to stabilize the k8s overlay output order when generating manifests, the overlay part would be put at the front, and the base manifests would be sorted by key, the same as when process the other manifests.

### DIFF
--- a/cmd/mesh/testdata/manifest-generate/output/pilot_override_kubernetes.yaml
+++ b/cmd/mesh/testdata/manifest-generate/output/pilot_override_kubernetes.yaml
@@ -209,39 +209,50 @@ spec:
     istio: pilot
 
 ---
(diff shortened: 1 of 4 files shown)


morvencao (approved):

ostromart:
@richardwxn there are some conflicts - could you please merge manually?
For future PRs, if you set the cherrypick-1.4 label *before* the PR merges, it will auto-cherry pick for you. Otherwise you have to issue the bot command manually.

richardwxn:
@ostromart ok will do, actually I found out that the overlay part order needed to be included in the sorted order as well, otherwise the order still not stable, I would create another PR to get it in master, and cherry pick those two.

PR #517, opened 2019-11-04
Try add a hint to use force flag

richardwxn:
Got couple issues complaining about the validation of values pass through part failed the command, it is hard to make sure the values schema proto is always updated. Try to let user know the existence of flag so they can choose to proceed.

### DIFF
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -43,7 +43,7 @@ var (
 func CheckIstioControlPlaneSpec(is *v1alpha2.IstioControlPlaneSpec, checkRequired bool) (errs util.Errors) {
 	errs = CheckValues(is.Values)
 	if errs != nil {
-		fmt.Println("Try run command with --force flag if you want to proceed with the validation error")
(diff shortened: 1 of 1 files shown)


pkg/validate/validate.go
```diff
@@ -41,7 +41,10 @@ var (
 // CheckIstioControlPlaneSpec validates the values in the given Installer spec, using the field map defaultValidations to
 // call the appropriate validation function.
 func CheckIstioControlPlaneSpec(is *v1alpha2.IstioControlPlaneSpec, checkRequired bool) (errs util.Errors) {
-	errs = util.AppendErrs(errs, CheckValues(is.Values))
+	errs = CheckValues(is.Values)
+	if errs != nil {
+		fmt.Println("Try run command with --force flag if you want to proceed with the validation error")
```
ostromart: Run the command with the --force flag if you want to ignore the validation error and proceed.

richardwxn: done

ostromart (approved):

PR #533, opened 2019-11-06
Sync installer charts

jwendell:


howardjohn (approved):

pitlv2109:
Ok so if I sync to this PR and `make mesh` then run `./mesh manifest generate --set profile=demo-auth` or any profile, there will be an error `render error in "istio-discovery/templates/telemetryv2.yaml": template: istio-discovery/templates/telemetryv2.yaml:1:50: executing "istio-discovery/templates/telemetryv2.yaml" at <.Values.pilot.telemetry.v2.enabled>: nil pointer evaluating interface {}.enabled`.

Syncing to current `master`(2e56cf404a00cf778ba3a98d42e43d8be1b262b5) is fine.

PR #534, opened 2019-11-06
Delete SDS profile

ostromart:
According to the security team, SDS profile does not work with the new installer charts. 
I'll also create a documentation PR to highlight this.


ostromart:
Just kidding, we're going to try to make SDS work.

pitlv2109:
Do you think we can make it work in 1.4? Otherwise I need to change the installation guide for the sds tasks.

howardjohn:
I think I have found the issue, will reach out to you if not so we can
debug it together maybe

PR #549, opened 2019-11-07
Update reverse translation for coredns

richardwxn:


pkg/translate/translate_value_test.go
```diff
@@ -326,6 +337,11 @@ telemetry:
    telemetry:
      enabled: false
  enabled: false
+coreDNS:
+ components:
+   coreDNS:
```
howardjohn: This feels weird

richardwxn: can you clarify?

richardwxn: in this case, the source values.yaml didn't enable coreDNS so it is expected.

howardjohn: Not necessarily and issue with this pr, but it makes no sense to do `--set coreDNS.enabled=true --set coreDNS.components.coreDNS.enabled=true`. Its so confusing for users that we have the same thing enabled in two places.

I don't know how our users will ever figure out the magic settings to get it to work

ostromart: That should not be necessary. Just --set coreDNS.enabled=true, the reason both were needed before was due to a bug.

richardwxn: yes the component enablement would inherit the feature one, unless you need to override

richardwxn:
@howardjohn PTAL

ostromart (approved):

PR #551, opened 2019-11-07
Enable ingress in remote profile

ostromart:


data/profiles/remote.yaml
```diff
@@ -17,7 +17,10 @@ spec:
     enabled: false
 
   gateways:
-    enabled: false
+    enabled: true
+    components:
+      egressGateway:
+        enabled: false
```
richardwxn: do you need to specify ingressgateway enabled true?

howardjohn (approved):

data/profiles/remote.yaml
ostromart: no, because it will inherit the feature setting.

PR #559, opened 2019-11-08
Fix aud for sds

howardjohn:
This was out of sync with the values.yaml from installer


richardwxn (approved):

ostromart:
I think this should've been updated through the migration script? Could you investigate what happened @richardwxn ? I also saw that the istioRemote values were not migrated (that was ok because they are all nil).

ostromart (approved):
Approved, with caveat that we need to find why auto-migrate failed.

PR #561, opened 2019-11-08
Fix podDNSSearchNamespaces validation issue

elfinhe:


pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -495,7 +495,7 @@ message GlobalConfig {
 
   google.protobuf.BoolValue operatorManageWebhooks = 41;
 
-  // The next available key is 43
+  repeated string podDNSSearchNamespaces = 43;
```
ostromart: could you pls move this up and add comment, like this:

  // Specifies the Configuration for Istio mesh across multiple clusters through Istio gateways.
  MultiClusterConfig multiCluster = 22;

  // Custom DNS config for the pod to resolve names of services in other
  // clusters. Use this to add additional search domains, and other settings.
  // see https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#dns-config
  // This does not apply to gateway pods as they typically need a different
  // set of DNS settings than the normal application pods (e.g. in multicluster scenarios).
  repeated string podDnsSearchNamespaces = 43;

... and update the next key to 44.

elfinhe: sure

ostromart (approved):

elfinhe:
fixing tests

elfinhe:
I will cherry pick this manually.

PR #578, opened 2019-11-11
clean local charts before update

richardwxn:
seems like the make update-charts has no effect because of previously cached charts.

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -45,7 +45,7 @@ gen-check: clean gen check-clean-repo
 
 clean: clean-values clean-types clean-vfs clean-charts
 
-update-charts: installer.sha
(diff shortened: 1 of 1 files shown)


ostromart:
is this required? make update-charts would not do anything if the charts were unchanged with the sha change i think. can you explain the case this is fixing?

richardwxn:
> is this required? make update-charts would not do anything if the charts were unchanged with the sha change i think. can you explain the case this is fixing?

My local charts never got updated with existing command, but I have not checked why. My guess is that it only works if I update my local installer.sha manually, but if I just sync with master this doesn't do the update work

sdake:
istio/operator is out of the support window. If you wish to reopen this PR, feel free.

PR #584, opened 2019-11-11
Turn on control plane security for sds

pitlv2109:


howardjohn (approved):
This matches istio/istio, lgtm

PR #590, opened 2019-11-11
[WIP]sync default profile

richardwxn:
Process: I used the manifest migrate command to generate migrated default profile as a base, and did some manual update for existing default profile( The process still cannot be fully automated based on couple reasons, still WIP for that)

Changes include:
1. I added some k8s settings such as podaffinity to the profile even though they are default as empty, trying to keep it more consistent with the values.yaml format we use.
2. Some of the settings such as env or readinessprobe are not exposed because they are just hardcoded in the charts, remove those from profile.
3.[WIP part] The values pass through part is quite out of sync, I am still working to check and sync them if needed.

### DIFF
--- a/data/profiles/default.yaml
+++ b/data/profiles/default.yaml
@@ -12,17 +12,11 @@ spec:
       pilot:
         enabled: true
         k8s:
+          affinity:
+            podAntiAffinity:
+              preferredDuringSchedulingIgnoredDuringExecution: []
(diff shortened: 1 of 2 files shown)


ostromart:
@richardwxn this is quite old and I'm not sure what the status is. If the default profile in sync now and can this be closed or is there any further work to be done?

sdake:
istio/operator is out of support window. If you wish to reopen this PR, feel free.

PR #591, opened 2019-11-11
update manifest migrate commands

richardwxn:
changes include:
1. Update the manifest migrate command output to be IstioControlPlane format instead of IstioControlPlaneSpec for easier comparison
2. Update scaleTargetRef reverse translation to work for gateways and mixer part.
3. Update run_migrate_profile.sh to use manifest profile diff command to generate the difference instead of using manifest diff
4. Update UnmarshalValuesWithJSONPB to take in params to decide whether to allowUnknownFields, previously it is hardcoded as true or false.

### DIFF
--- a/pkg/translate/translate_value.go
+++ b/pkg/translate/translate_value.go
@@ -80,6 +80,9 @@ var (
 	// Component enablement mapping. Ex "{{.ValueComponent}}.enabled": {"{{.FeatureName}}.Components.{{.ComponentName}}.enabled}", nil},
 	// Feature enablement mapping. Ex: "{{.ValueComponent}}.enabled": {"{{.FeatureName}}.enabled}", nil},
 	componentEnablementPattern = "{{.FeatureName}}.Components.{{.ComponentName}}.Enabled"
+	// specialComponentPath lists cases of component path of values.yaml we need to have special treatment.
+	specialComponentPath = []string{"mixer.policy", "mixer.telemetry", "gateways.istio-ingressgateway", "gateways.istio-egressgateway",
+		"mixer", "gateways"}
 )
 
 // initAPIMapping generate the reverse mapping from original translator apiMapping.
@@ -326,10 +329,11 @@ kind: Deployment
 name: istio-%s`
 
 	// need to do special handling for gateways and mixer
-	// ex. deployment name should be istio-telemetry instead of istio-mixer.telemetry
-	if newPS == "mixer.policy" || newPS == "mixer.telemetry" ||
-		newPS == "gateways.istio-ingressgateway" || newPS == "gateways.istio-egressgateway" {
-		newPS = newP[1 : len(newP)-1].String()
+	// ex. because deployment name should be istio-telemetry instead of istio-mixer.telemetry, we need to get rid of the prefix mixer part.
+	for _, cp := range specialComponentPath {
+		if cp == newPS && len(newP) > 2 {
+			newPS = newP[1 : len(newP)-1].String()
+		}
 	}
 
 	stString := fmt.Sprintf(stVal, newPS)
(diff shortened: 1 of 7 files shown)


pkg/translate/translate_value.go
```diff
@@ -325,6 +325,13 @@ apiVersion: apps/v1
 kind: Deployment
 name: istio-%s`
 
+	// need to do special handling for gateways and mixer
+	// ex. deployment name should be istio-telemetry instead of istio-mixer.telemetry
+	if newPS == "mixer.policy" || newPS == "mixer.telemetry" ||
+		newPS == "gateways.istio-ingressgateway" || newPS == "gateways.istio-egressgateway" {
+		newPS = newP[1 : len(newP)-1].String()
```
ostromart: comment above doesn't agree with what the code is doing - here were're just truncating the path, example shows translation.

pkg/translate/translate_value.go
```diff
@@ -535,7 +542,7 @@ func (t *ReverseTranslator) isEnablementPath(path util.Path) bool {
 
 	pstr := path.String()
 	if pstr == "mixer.policy.enabled" || pstr == "mixer.telemetry.enabled" ||
-		pstr == "gateways.enabled" ||
+		pstr == "gateways.enabled" || pstr == "mixer.enabled" ||
```
ostromart: there are enough of these that you might want to define as a map at the top of the file. it will be hard to keep it updated here.

pkg/translate/translate_value.go
richardwxn: added more clear explanation

pkg/translate/translate_value.go
richardwxn: added a var at the top of the file

richardwxn:
@ostromart can you PTAL again?

pkg/translate/translate_value.go
```diff
@@ -80,6 +80,9 @@ var (
 	// Component enablement mapping. Ex "{{.ValueComponent}}.enabled": {"{{.FeatureName}}.Components.{{.ComponentName}}.enabled}", nil},
 	// Feature enablement mapping. Ex: "{{.ValueComponent}}.enabled": {"{{.FeatureName}}.enabled}", nil},
 	componentEnablementPattern = "{{.FeatureName}}.Components.{{.ComponentName}}.Enabled"
+	// specialComponentPath lists cases of component path of values.yaml we need to have special treatment.
+	specialComponentPath = []string{"mixer.policy", "mixer.telemetry", "gateways.istio-ingressgateway", "gateways.istio-egressgateway",
```
ostromart: It's more idiomatic to use a map for this so you can avoid the for loops below and just do specialComponentPath[key] to test.

richardwxn: done

pkg/translate/translate_value.go
```diff
@@ -330,10 +336,8 @@ name: istio-%s`
 
 	// need to do special handling for gateways and mixer
 	// ex. because deployment name should be istio-telemetry instead of istio-mixer.telemetry, we need to get rid of the prefix mixer part.
-	for _, cp := range specialComponentPath {
-		if cp == newPS && len(newP) > 2 {
-			newPS = newP[1 : len(newP)-1].String()
-		}
+	if _, exist := specialComponentPath[newPS]; exist && len(newP) > 2 {
```
ostromart: Better to use a map of [string]bool like this: https://goplay.googleplex.com/p/Kl2ngRlsnR3
Then you can write if specialComponentPath[newPS] && len(newP) > 2
Note that although it's idiomatic in Go to check for existence of map keys separately as you have, it's also ok and encouraged to use the default value being returned in the absence of a key e.g. for myMap map[string]string if myMap[key] == ""

richardwxn: done

ostromart (approved):

PR #592, opened 2019-11-12
Make apply errors more visible

ostromart:
This prints an error message at the end of apply if any errors showed up for any component. Currently it's easy to miss errors as they scroll by.

### DIFF
--- a/cmd/mesh/manifest.go
+++ b/cmd/mesh/manifest.go
@@ -134,11 +134,13 @@ func genApplyManifests(setOverlay []string, inFilename string, force bool, dryRu
 	if err != nil {
 		return fmt.Errorf("failed to apply manifest with kubectl client: %v", err)
 	}
+	gotError := false
 	for cn := range manifests {
 		if out[cn].Err != nil {
 			cs := fmt.Sprintf("Component %s failed install:", cn)
 			l.logAndPrintf("\n%s\n%s", cs, strings.Repeat("=", len(cs)))
 			l.logAndPrint("Error: ", out[cn].Err, "\n")
+			gotError = true
 		} else {
 			cs := fmt.Sprintf("Component %s installed successfully:", cn)
 			l.logAndPrintf("\n%s\n%s", cs, strings.Repeat("=", len(cs)))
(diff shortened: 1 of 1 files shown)


cmd/mesh/manifest.go
```diff
@@ -134,23 +134,31 @@ func genApplyManifests(setOverlay []string, inFilename string, force bool, dryRu
...
 		}
 		if strings.TrimSpace(out[cn].Stdout) != "" {
 			l.logAndPrint(out[cn].Stdout, "\n")
 		}
 	}
+
+	if gotError {
+		l.logAndPrint("\n\n*** Errors were logged during apply operation. Please check component installation logs above. ***\n")
+	}
+
```
ayj: We return a non-zero code to the user if the command failed. Will returning an error here make its way back to the exit code status?

ostromart: This is a bit tricky because showing error outputs does not necessarily mean failed install. That's because we use kubectl under the hood which unfortunately sends warnings to stderr. 
This is the reason for saying "inspect the logs" rather than "your installation failed".
To assert the latter, we will need to do more e.g. check for the actual installed state upon completion.
For the reasons above, I propose not returning error code for this particular path (we do elsewhere, in the case of a clear error).

ericvn: Hopefully, the message lower to the command prompt will be caught if not returning an error code. I know the failure I was seeing scrolled off the screen to fast to see the failure.

ostromart:
@howardjohn could you please approve?

elfinhe (approved):

ostromart:
Since it didn't make it in last night, could you please take another look @ayj and see if you are ok with this as is?

howardjohn (approved):

cmd/mesh/manifest.go
ayj: This PR is an improvement. We do need to get to the point where uses don't need to inspect logs to confirm a successful install. This should be one of the main responsibilities off an operator (CLI or controlled based).

One of the error paths above says "component failed to install". I wouldn't expect to see successful install  if one of the components couldn't be installed. Are you concerned about false positives? If anything, I think we'd want to error on the side of caution and accept more false negatives (and require the user to inspect the logs to see if it really was a failure) vs. proceed with errors (and rely on the user to inspect logs when the command returned success).

ostromart: The "component failed to install" message has also been changed to "component install returned the following errors". 
Yes, agreed, we do need to be able to declare whether Istio installed successfully - an item for 1.5.

PR #594, opened 2019-11-12
Remove original cr yaml from error message

richardwxn:
The reason why I propose this change is because usually the original cr yaml is really large, making it hard to see the actual error message from the output, also user can just refer to input file instead of looking into the output for the cr content.

ok there has been a recent change to dump the output to stderr instead, but still I think it increases complexity for debugging


richardwxn:
oh there has been a recent change to dump the output to stderr instead, so looks like this change is not that necessary

ostromart (approved):

ostromart:
@richardwxn lint is still failing.

PR #597, opened 2019-11-12
Sync installer

elfinhe:


ostromart (approved):

ostromart:
mandiff error is most likely due to default values changing. 
to see the exact diff, you can run it locally with BUILD_WITH_CONTAINER=0 and look at the diff files in /tmp

elfinhe:
> mandiff error is most likely due to default values changing.
> to see the exact diff, you can run it locally with BUILD_WITH_CONTAINER=0 and look at the diff files in /tmp

thanks.

jwendell (approved):

PR #605, opened 2019-11-13
Add vm for multi network support sample

irisdingbj:


irisdingbj:
@ostromart @ericvn @sdake can we get this PR in 1.4? As this is a pre-req for https://github.com/istio/istio.io/pull/5652 as we need update all docs to use `istioctl` for installation.

@frankbu FYI

data/examples/vm/values-istio-meshexpansion-gateways.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: install.istio.io/v1alpha2
+kind: IstioControlPlane
+spec:
+  values:
+    global:
+      # Provides dns resolution for global services
+      podDNSSearchNamespaces:
+        - global
```
sdake: line 8 won't work (routing doesn't work properly for service.global).  Additionally DNS is limited to 6 entries, and we may overload the DNS cache.

sdake (commented):
I have set this to backport to 1.4.  The earliest this release will make the public will be 1.4.1.  Once you resolve the problem on line 8, we can approve.

data/examples/vm/values-istio-meshexpansion-gateways.yaml
```diff
@@ -0,0 +1,27 @@
...
+        - "{{ valueOrDefault .DeploymentMeta.Namespace \"default\" }}.global"
```
richardwxn: how does this template part works?

irisdingbj: Just copy from https://github.com/istio/operator/blob/master/data/examples/multicluster/values-istio-multicluster-gateways.yaml and verified this works for vm mesh expansion scenario.

irisdingbj:
@sdake any suggestions for `line 8 won't work (routing doesn't work properly for service.global). Additionally DNS is limited to 6 entries, and we may overload the DNS cache.` 

One thing is the current version actually works as I have verified this.

data/examples/vm/values-istio-meshexpansion-gateways.yaml
morvencao: This will be added to sidecar injection template and rendered by the sider injector, see: https://github.com/istio/installer/blob/master/istio-control/istio-autoinject/files/injection-template.yaml#L390-L392

data/examples/vm/values-istio-meshexpansion-gateways.yaml
irisdingbj: removed the whole section here and verifies it works

morvencao (approved):

data/examples/vm/values-istio-meshexpansion-gateways.yaml
irisdingbj: I have removed the podDNSSearchNamespaces section and verified it works

sdake:
> @sdake any suggestions for `line 8 won't work (routing doesn't work properly for service.global). Additionally DNS is limited to 6 entries, and we may overload the DNS cache.`
> 
> One thing is the current version actually works as I have verified this.

@irisdingbj the DNS overrides don't work in some cloud environments where there are more then 4 DNS entries in /etc/resollv.conf (inside the container on instantiation).  Sometiimes it will work sometimes it wont.  I think enabling multicluster turns on line 8 and 9 in the prior PRs.

For corrective action, I'd recommend deleting the "global" line but keeping the "namespace.global" line.   This probably needs to be done in the istio/installer repo.

Making service.global routing work is very difficult and requires major work to pilot.

For now, this PR looks good.

PR #608, opened 2019-11-13
Use gobuild.sh to build operator

ostromart:
This is required for version info to be populated during build time.


ostromart:
@sdake could you ptal - I removed a line which appears to be redundant now. DNM until you approve.

howardjohn (approved):

howardjohn:
should vfsgen change?

ostromart:
you mean the generated assets? no, that is unexpected. all the generation should run before the controller binary is built and should not cause any change.

ostromart:
interesting... after i run make clean gen again, the changes to assets.gen.go disappear.

ostromart:
somehow this is caused by the deleted line. not sure why it has this effect - maybe @sdake can shed some light.

sdake (approved):
@ostromart I am not sure why a deleted line would imipact the vfsgen assets, but this PR LGTM.

PR #610, opened 2019-11-14
Use gobuild.sh to build operator (#608)

ostromart:
Manual cherry pick due to conflict. 

* Use gobuild.sh to build operator

* Make gen

* Restore deleted line

* Add comment for posterity, convert mesh target also


howardjohn (approved):
fyi @jwendell

PR #611, opened 2019-11-15
Swallow ns error if the istio ns pre-exists

irisdingbj:
https://github.com/istio/istio/issues/18948

### DIFF
--- a/pkg/manifest/installer.go
+++ b/pkg/manifest/installer.go
@@ -321,13 +321,13 @@ func applyManifest(componentName name.ComponentName, manifestStr string, version
 
 		stdoutNs, stderrNs, err := kubectl.Apply(opts.DryRun, opts.Verbose, opts.Kubeconfig, opts.Context, namespace, mns, extraArgs...)
 		stdout += "\n" + stdoutNs
-		stderr += "\n" + stderrNs
+		outerr := stderr + "\n" + stderrNs
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/manifest/installer.go
```diff
@@ -321,13 +321,13 @@ func applyManifest(componentName name.ComponentName, manifestStr string, version
 
 		stdoutNs, stderrNs, err := kubectl.Apply(opts.DryRun, opts.Verbose, opts.Kubeconfig, opts.Context, namespace, mns, extraArgs...)
 		stdout += "\n" + stdoutNs
-		stderr += "\n" + stderrNs
+		outerr := stderr + "\n" + stderrNs
```
ostromart: I may be misreading this, but doesn't this lead to the exact same output as before?

irisdingbj: Now we will not the error  error like  `kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply (repeated 1 times)` in the output if the there is no `err` in the namespace creation. We will only report this error if we fail to create the namespace and there is `err` in the namespace creation.

ostromart: I guess what confuses me here is you are replacing:
a += "foo"
with:
b := a + "foo"
Isn't that the same value? I seem to be missing something obvious here.

irisdingbj: @ostromart  Thanks for the review. Seems this already been fixed. So let me close this pr.

PR #618, opened 2019-11-17
sync values types

richardwxn:
Try to reduce some pains of having to use --force flag by syncing the values_types.proto


pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -295,6 +316,12 @@ message GalleyConfig {
 
   // Enable analysis and status update in Galley
   google.protobuf.BoolValue enableAnalysis = 8;
+
+  // K8s rolling update strategy
+  TypeIntOrStringForPB rollingMaxSurge = 9;
```
howardjohn: Why is this in Galley config and not the general K8s?

pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -142,6 +151,18 @@ message CoreDNSConfig {
...
   // See https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#resource-requests-and-limits-of-pod-and-container
   Resources resources = 8;
+
+  // K8s rolling update strategy
+  TypeIntOrStringForPB rollingMaxSurge = 9;
+
+  // K8s rolling update strategy
+  TypeIntOrStringForPB rollingMaxUnavailable = 10;
+
+  TypeSliceOfMapStringInterface podAntiAffinityLabelSelector = 11;
```
howardjohn: Part of K8s config?

pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -96,15 +96,24 @@ message CertManagerConfig {
   // Image name for the CertManager Deployment.
   string image = 4;
 
+  uint32 replicaCount = 5;
```
howardjohn: same comment

richardwxn:
@howardjohn the reason is that we are still exposing these options(k8s settings, enablement) in values.yaml, although they should only reside in IstioControlPlane API later. Also once we remove those settings from values_types.proto, we should add corresponding error messages in validation to let user know that these should be set under IstioControlPlane API section instead of values.yaml pass through

howardjohn:
ok, is it possible to mark as deprecated now?

richardwxn:
@howardjohn Yep I think we need to do that but gradually. Existing users are already familiar with helm installation so they may tend to put all overlay into values part, they may have to apply force flag every time if we remove all overlapped IstioControlPlane fields now. Instead, I think we can add some warning message in the validation logic now to let user know that things they are overlaying would be deprecated soon and what would be the correct path to overlay. Then in 1.5 maybe? we can deprecate them. WDYT?

pkg/apis/istio/v1alpha1/values_types.proto
ostromart: This is the values schema not IstioControlPlane so it's just a statement of what there is today, for better or worse.

pkg/apis/istio/v1alpha1/values_types.proto
ostromart: Because values.yaml was poorly designed :-)

howardjohn:
By deprecate I don't mean remove, I mean mark as deprecated. For example in the proto:
```proto
 bool java_generate_equals_and_hash = 20 [deprecated=true];
```

richardwxn:
@howardjohn I push new commit to mark k8s related settings as deprecated, but that servers mainly the documentation purpose for developer. We may still need to add some message in the validation function for user to know about it.

richardwxn:
@howardjohn @ostromart PTAL again

ostromart (approved):

PR #623, opened 2019-11-20
Ignore a warning in kubectl apply when namespace is already created.

### LINKED ISSUE #18948: `mesh manifest apply` error
Run the command below to install istio. Before this, I have the namespace already created.

```
istioctl manifest apply
```

I got this error in the end:

```
*** Errors were logged during apply operation. Please check component installation logs above. ***

```

which is caused by:

```
Component Base installed successfully:
======================================
Error detail:

Warning: kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply (repeated 1 times)





namespace/istio-system configured

```

Related issue in k8s: https://github.com/kubernetes/kubernetes/issues/82154

elfinhe:
Fix: https://github.com/istio/istio/issues/18948

### DIFF
--- a/cmd/mesh/manifest.go
+++ b/cmd/mesh/manifest.go
@@ -37,6 +37,13 @@ import (
 	binversion "istio.io/operator/version"
 )
 
+var (
+	ignoreStdErrList = []string{
+		// TODO: remove when https://github.com/kubernetes/kubernetes/issues/82154 is fixed.
+		"Warning: kubectl apply should be used on resource created by either kubectl create --save-config or kubectl apply",
+	}
+)
+
 // ManifestCmd is a group of commands related to manifest generation, installation, diffing and migration.
 func ManifestCmd() *cobra.Command {
 	mc := &cobra.Command{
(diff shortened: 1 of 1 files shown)


cmd/mesh/manifest.go
```diff
@@ -146,7 +146,7 @@ func genApplyManifests(setOverlay []string, inFilename string, force bool, dryRu
 			l.logAndPrintf("\n%s\n%s", cs, strings.Repeat("=", len(cs)))
 		}
 
-		if strings.TrimSpace(out[cn].Stderr) != "" {
+		if !isStderrOkay(out[cn].Stderr) {
```
ostromart: Consider renaming function to ignoreError, which is more meaningful.

cmd/mesh/manifest.go
```diff
@@ -162,6 +162,20 @@ func genApplyManifests(setOverlay []string, inFilename string, force bool, dryRu
 	return nil
 }
 
+func isStderrOkay(stderr string) bool {
+	trimmedStdErr := strings.TrimSpace(stderr)
+	ignoreList := []string{
```
ostromart: better to put this list in var() block at the top to give it more visibility.

cmd/mesh/manifest.go
elfinhe: Fixed.

cmd/mesh/manifest.go
elfinhe: Fixed.

ostromart (approved):

PR #628, opened 2019-11-22
Clean up code

ostromart:
Cosmetic changes only - moving functions to appropriate packages and adding missing comments. 
These disruptive changes need to be done for a while, this seems to be a good quiet time to finally get them in.


pkg/translate/translate.go
```diff
@@ -526,3 +576,77 @@ func mergeK8sObject(base *object.K8sObject, overlayNode interface{}, path util.P
...
+//     will constructs the following patch object:
+//       spec:
+//         template:
+//           spec:
+//             containers:
+//             - name: discovery
+//               env:
+//               - name: NEW_VAR
+//                 value: new_value
+func createPatchObjectFromPath(node interface{}, path util.Path) (map[string]interface{}, error) {
```
richardwxn: would it be  better to move this to tpath or some other util file?

cmd/mesh/upgrade.go
```diff
@@ -177,6 +181,8 @@ func upgrade(rootArgs *rootArgs, args *upgradeArgs, l *logger) (err error) {
 	l.logAndPrintf("Upgrade version check passed: %v -> %v.\n", currentVersion, targetVersion)
 
 	// Read the overridden values from args.inFilename
+	// TODO: Is this correct? Seems to be checking only the overlays under global. Other parts in ICPS can be
+	// overlaid too.
```
richardwxn: @elfinhe Can you create an issue and check this part?

ostromart: I already did: https://github.com/istio/istio/issues/19135

pkg/translate/translate.go
ostromart: I considered based on the function name but on closer inspection this is only used in this file and doesn't contain any functionality that would be useful to export at this time.

richardwxn (approved):

PR #639, opened 2019-11-27
Add unit test and fix for ICP.DeepCopy()

luksa:
DeepCopy() always failed with an error, which was only logged and wasn't
handled properly. The error occured when unmarshalling CreatedTimestamp,
which meant that all remaining fields were never unmarshalled.


pkg/apis/istio/v1alpha2/deepcopy.go
```diff
@@ -15,22 +15,22 @@
...
+	"encoding/json"
 	"github.com/gogo/protobuf/proto"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
-	"istio.io/operator/pkg/util"
 	"istio.io/pkg/log"
 )
 
 // DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
```
howardjohn: If this is autogenerated do we need to change the generation code instead of directly here? although it doesn't mention where its generated from..

luksa: Autogenerated deepcopy is usually found in `zz_generated.deepcopy.gen.go`. Looks like the code (and this comment) here was copied from one of those generated files, but is now maintained manually. I may be wrong, as I've only just begun looking at the code here.

howardjohn:
That sounds right. lgtm but will let someone more familiar approve. Thanks for adding tests

pkg/apis/istio/v1alpha2/deepcopy.go
```diff
@@ -15,22 +15,22 @@
...
 func (m *IstioControlPlane) DeepCopyInto(out proto.Message) {
-	inj, err := util.MarshalWithJSONPB(m)
```
ostromart: I doubt we can get a correct copy without jsonpb, because the struct tags are not treated correctly. You could test it out by trying a copy with data/profiles/default.yaml and see what happens.

pkg/apis/istio/v1alpha2/deepcopy.go
ostromart: It's not autogenerated, but I don't think it will work - see comment below.

sdake:
@luksa welcome to the Istio project with your first commit!  Please note the check for linting is failing:

```
Building with your local toolchain.
pkg/apis/istio/v1alpha2/deepcopy.go:18: File is not `goimports`-ed with -local istio.io/ (goimports)
	"encoding/json"
pkg/apis/istio/v1alpha2/deepcopy_test.go:18: File is not `goimports`-ed with -local istio.io/ (goimports)
	"github.com/gogo/protobuf/types"
	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/utils/pointer"
common/Makefile.common.mk:42: recipe for target 'lint-go' failed
make: *** [lint-go] Error 123
```

If you are not familiar with Istio's CI process, you can determine what failed by clicking "Details" wherever there is a red X on a required test.

See screenshot below:
Screen Shot 2019-11-27 at 10 10 43 AM

pkg/apis/istio/v1alpha2/deepcopy.go
luksa: I've added another test case, which reads the default.yaml file and it works fine. I could simply use UnmarshalWithJSONPB to parse the whole file, since it breaks when unmarshalling the metadata field, so the test parses only the spec, just like the existing code does.

luksa (commented):
@sdake The `make lint-go` command is what I was trying to find. Thanks for the pointer :). I've now configured my IDE to format imports correctly.

pkg/apis/istio/v1alpha2/deepcopy.go
ostromart: Ok, cool, thanks. Could you please also check that operator controller is not broken? We don't have an e2e yet for that unfortunately.

luksa: I ran the controller locally, created the example-icp and it appeared to deploy everything properly. All the pods were up and ready. Deleting the ICP also worked properly. I haven't tested any further than that.

ostromart (approved):

ostromart:
Thanks @luksa !

PR #645, opened 2019-11-28
Export Logger and NewLogger

irisdingbj:


morvencao (approved):

elfinhe:
@irisdingbj should we also add this new Logger to 1.4, considering future PRs may depend on it?

PR #646, opened 2019-11-28
Disable leader election when LEADER_ELECTION_NAMESPACE is not set

luksa:
Debugging the controller is practically impossible, since it loses
the leader election lock when it's suspended for too long. This causes
the process to be terminated.

### DIFF
--- a/cmd/manager/server.go
+++ b/cmd/manager/server.go
@@ -81,12 +81,8 @@ func getWatchNamespace() (string, error) {
 }
 
 // getLeaderElectionNamespace returns the namespace in which the leader election configmap will be created
-func getLeaderElectionNamespace() (string, error) {
-	ns, found := os.LookupEnv("LEADER_ELECTION_NAMESPACE")
-	if !found {
-		return "", fmt.Errorf("LEADER_ELECTION_NAMESPACE must be set")
-	}
-	return ns, nil
+func getLeaderElectionNamespace() (string, bool) {
+	return os.LookupEnv("LEADER_ELECTION_NAMESPACE")
 }
 
 func run() {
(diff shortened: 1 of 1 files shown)


morvencao (commented):
/ok-to-test

cmd/manager/server.go
```diff
@@ -95,9 +91,9 @@ func run() {
 		log.Fatalf("Failed to get watch namespace: %v", err)
 	}
 
-	leaderElectionNS, err := getLeaderElectionNamespace()
-	if err != nil {
-		log.Fatalf("Failed to get leader election namespace: %v", err)
+	leaderElectionNS, leaderElectionEnabled := getLeaderElectionNamespace()
+	if !leaderElectionEnabled {
+		log.Warn("Leader election namespace not set. Leader election is disabled.")
```
richardwxn: Would it be better to add more description to tell user that disabling leader election is not encouraged, and might be only used for debugging purpose.

cmd/manager/server.go
```diff
@@ -95,9 +91,9 @@ func run() {
 		log.Fatalf("Failed to get watch namespace: %v", err)
 	}
 
-	leaderElectionNS, err := getLeaderElectionNamespace()
-	if err != nil {
-		log.Fatalf("Failed to get leader election namespace: %v", err)
+	leaderElectionNS, leaderElectionEnabled := getLeaderElectionNamespace()
+	if !leaderElectionEnabled {
```
richardwxn: Also for your original concern, would it solve the issue if you attach the debug process after it acquires the lock

luksa: It wouldn't. The problem is that the leader election thread needs to run constantly so it can keep updating the lock config map. If it doesn't do that before the lock expires, it's no longer the leader and must terminate immediately. And that's exactly what happens when you allow the thread to run after being suspended for too long.

cmd/manager/server.go
luksa: I've added a warning that this isn't appropriate for production use.

howardjohn (approved):

PR #650, opened 2019-12-02
This introduces the latest installer.sha

### LINKED ISSUE #19227: Istio 1.4.0 tracing service moved from port 80 to 9411
**Bug description**
In istio 1.4.0, `tracing.istio-system` has moved from port `80` to `9411`.  There is already a `zipkin.istio-system` service in the namespace on `9411`.  Both services point to the same workload (Jaeger).

This is a breaking change for us so just wondering why it was made and if it was intentional, it should have been in the upgrade notes (as we have upstream systems talking to tracing.istio-system).  

**Expected behavior**
Consistent service ports between releases or details in release notes explaining the changes.

OR if you're really honouring semver releases, 1.4.0 shouldn't have breaking changes from 1.3.x and a different service port could probably be classed as breaking.

1.4.0

**How was Istio installed?**

sdake:
This should resolve the jaeger regression around ports. Additionally
other features have been added to the installer.

Fixes: https://github.com/istio/istio/issues/19227


sdake:
Verified (manually) this resolves the issue with the upstream charts.

```
tracing                  ClusterIP      10.97.200.88     <none>          80/TCP                                                                                                                       45s
zipkin                   ClusterIP      10.99.33.142     <none>          9411/TCP                                                                                                                     46s
```
which is correct. In the 1.4 release both tracing and zipkin pointed to port 9411.

howardjohn (approved):

PR #652, opened 2019-12-02
Read ICP only from the local cache

luksa:
Previously, the controller read the ICP object twice - first as an
Unstructured object, and then from the local cache. This was a
workaround around the issue where ICP.DeepCopy(), which is called when
you get the object from the cache, didn't work properly. With DeepCopy
now fixed, we can get the ICP from the cache and there's no need to also
get it from the API server.


ostromart (approved):
Thanks Marko! This has been a pain in the neck for quite some time.

morvencao (commented):
Thanks @luksa

PR #655, opened 2019-12-02
Add option to omit base in manifest generate

ostromart:
Generating user gateways with istioctl is quite painful. 
This change allows users to generate YAML for a custom gateway using the following command:
istioctl manifest generate --no-base -f examples/user-gateway/ingress-gateway-only.yaml


cmd/mesh/manifest-generate.go
```diff
@@ -35,13 +35,16 @@ type manifestGenerateArgs struct {
...
+	// noBase suppresses the generation of CRDs for the output manifest.
+	noBase bool
 }
 
 func addManifestGenerateFlags(cmd *cobra.Command, args *manifestGenerateArgs) {
 	cmd.PersistentFlags().StringVarP(&args.inFilename, "filename", "f", "", filenameFlagHelpStr)
 	cmd.PersistentFlags().StringVarP(&args.outFilename, "output", "o", "", "Manifest output directory path")
 	cmd.PersistentFlags().StringSliceVarP(&args.set, "set", "s", nil, SetFlagHelpStr)
 	cmd.PersistentFlags().BoolVar(&args.force, "force", false, "Proceed even with validation errors")
+	cmd.PersistentFlags().BoolVar(&args.noBase, "no-base", false, "Omit generation of base resources like namespaces and CRDs")
```
howardjohn: Why a special flag and not just a component like how everything else works?

ostromart: It could be confusing to users to have an explicit base component. Keep in mind that the custom gateway is the 10%. 
But I'm open, wdyt @sdake?

howardjohn: I am not sure how its any less confusing? Why is base a special case? Why would a user know to pass the `--no-base` to get rid of base but `--set pilot.enabled=false` to get rid of pilot?

ostromart: Ok, working on the alternative PR - let's see how that looks and we can make a final decision.

ostromart:
See alternative implementation: https://github.com/istio/operator/pull/658

PR #658, opened 2019-12-02
Add base feature and component and empty profile

ostromart:
A different way to help with generating gateways, instead of https://github.com/istio/operator/pull/655.
Here we're making base feature and component explicit in the proto and setting in default profile.


richardwxn (commented):
I think this approach is clearer and less confusing than the previous one

richardwxn (approved):

PR #664, opened 2019-12-05
Return error to reconciler when process manifest fails

irisdingbj:
https://github.com/istio/istio/issues/19125

### DIFF
--- a/pkg/helmreconciler/reconciler.go
+++ b/pkg/helmreconciler/reconciler.go
@@ -15,6 +15,7 @@
 package helmreconciler
 
 import (
+	"fmt"
 	"sync"
 
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


morvencao (commented):
not sure if this makes sense.
Once these errors are returned to the main reconcile loop, the controller will re-trigger new reconcile loop. 
However, some kinds of manifest process errors will emerge, no matter how many times the reconcile loop are triggerred, in these cases, we'd better keep the process in the ICP status.

pkg/helmreconciler/reconciler.go
```diff
@@ -115,7 +116,11 @@ func (h *HelmReconciler) Reconcile() error {
 	errs = util.AppendErr(errs, h.customizer.Listener().EndPrune())
 
 	errs = util.AppendErr(errs, h.customizer.Listener().EndReconcile(h.instance, status))
-
+	for _, v := range status.Status {
+		if v.Error != "" {
```
ostromart: It would be better to check against the enum status field (!=ERROR).

PR #667, opened 2019-12-05
Cache objects to reduce updates, simplify pruning

ostromart:
Currently, we write every single object in the Istio manifest to the API server each time the ICP CR is touched. This is very inefficient and slows down updates. This PR introduces a simple caching layer that only writes changed objects. 
It also simplifies pruning, which currently relies on an epoch annotation (which defeats caching). The new scheme follow the kubectl --prune approach, which deletes any selected objects that are not in the passed in manifest.


deploy/crds/istio_v1alpha2_istiocontrolplane_cr.yaml
```diff
@@ -2,8 +2,17 @@
 apiVersion: install.istio.io/v1alpha2
 kind: IstioControlPlane
 metadata:
-  namespace: istio-operator
+  namespace: istio-system
```
morvencao: Is this intent? Do we also need to update the namespace in operator chart: https://github.com/istio/operator/tree/master/data/operator

pkg/helmreconciler/pruning.go
```diff
@@ -28,27 +29,20 @@ import (
...
+func (h *HelmReconciler) Prune(excluded map[string]bool, all bool) error {
 	allErrors := []error{}
-	namespacedResources, nonNamespacedResources := h.customizer.PruningDetails().GetResourceTypes()
+	namespacedResources, _ := h.customizer.PruningDetails().GetResourceTypes()
 	targetNamespace := h.customizer.Input().GetTargetNamespace()
-	err := h.PruneResources(namespacedResources, all, targetNamespace)
-	if err != nil {
-		allErrors = append(allErrors, err)
-	}
-	err = h.PruneResources(nonNamespacedResources, all, "")
```
morvencao: we still need to prune all the `nonNamespacedResources`,  isn't it?

ostromart: No, we decided to leak the common resources so this was actually incorrect. The reason is if you need to rollback, the deleted CRDs will delete the CRs and this will make it impossible to restore the previous state.

deploy/crds/istio_v1alpha2_istiocontrolplane_cr.yaml
ostromart: Yes, it's intentional. The ICP should be in istio-system and the operator should be watching that. The operator resources should remain in istio-operator though.

pkg/helmreconciler/pruning.go
rcernich: We shouldn't be deleting CRDs, but should be deleting the other cluster scoped resources (e.g. ClusterRole, ClusterRoleBinding, etc.).

rcernich (changes requested):
I don't think you can rely on caching.  If the operator is bounced, you lose the cache.  This is the reason state information is not stored in the status field.  Also, the resources may change outside the scope of the operator, which means you still need to hit the api server to get an up to date copy of the resource.

As for the generation annotation, it shouldn't be removed.  There is no way to determine what objects weren't created.  The prune logic looks for all resources that are related to this ICP that are not current and deletes them.

I believe this change will diminish the stability of the operator.

ostromart:
You're right that if user edits the resource we'd have a problem but we don't want users to do that, there's no reason to edit resources directly on the API server if they are managed by the operator, it should always be done through the CR. 
Restarting the operator flushes the cache but it does not cause any inconsistency. 
We need convergence time to be faster than it is right now. Also, the merge logic causes lots of errors which are extremely difficult to debug.

rcernich:
It's more than the user editing the resources.  ResourceVersion seems to change all the time.  That's just one example.  Kubernetes will also be updating status fields.  If the operator can be bounced without negative impacts, then that's okay.

As for the generation annotation, that's pretty key to the prune logic.  As I mentioned above, there is no way to discern what objects weren't created.

Could you point me to the merge logic?  I'm not sure what that is.

Also, regarding performance, have you done any testing to see where the bottlenecks are?  In our maistra operator, the only performance bottleneck is in the prune logic, but that still only takes approximately 15-20s to reconcile.  A complete install, which waits for all deployments to become available only takes 2-3 minutes, which includes pulling images, scheduling, and initialization (i.e. most of that time is spent waiting for the deployments to become ready).  An incremental update (i.e. change to icp) takes seconds.  @sdake was mentioned that rendering charts through the cli took lots of time as well (I haven't confirmed), is the issue with the rendering?

ostromart:
time istioctl manifest generate > /dev/null
real	0m0.931s
user	0m1.278s
sys	0m0.098s

It's not the rendering, it really looks like it's because we're touching every object in the full manifest each time that any change happens to ICP. I'm seeing 20-30s typical convergence time for little changes and that really negatively impacts the user experience because the k8s update expectation is more like <5s. I'm open to other approaches to achieve that but imo we need to land that improvement in 1.5.
The pruning logic change I made in this PR is just to delete everything that isn't in the generated manifest (excluding CRDs, which we need to keep for upgrade rollback-ability reasons) - it's exactly what kubectl --prune does. Could you give a concrete example where this would lead to a bad outcome?

rcernich:
I think I'm starting to understand.

Do we filter events coming through the reconcile loop, i.e. are we reconcile every resource change or only changes to the generation?  If we're not doing this, we should be.  This means the render/apply/prune logic only executes when the spec actually changes.

I think I understand what you're talking about with the prune logic.  It should be simple enough, in that case, to cache the resource kind and name.  My only concern here would be that once the operator is modified to execute incrementally, that the cache could get out of sync.  That said, I don't think it would.

As stated above, the prune logic should be what takes the most time during any update, which my unscientific testing is around 15-20s, which is in line with your numbers for an update.  @luksa is investigating using typed objects in the prune logic, as this uses the cache in the client, which should greatly increase performance in the prune logic.  IMO, this is the proper solution.  (FWIW, the unstructured client doesn't use any caching, so every call hits the api server.)

rcernich:
Also, FWIW, I like the idea of removing the generation annotation, as it really saves updating resources during an update.  (Usually, the only thing that changes for most objects during an update is the generation label.)  That said, having the annotation does identify which version of the ICP resource any particular resource was created from.

ostromart:
If we can make go-client caching work for writes then that would be a great solution for this. In that case at least we'd have to remove the generation annotation because that would force a write every time. It's also possible that the patching we're doing (which this PR also removes) would force the write. 
I'd like to hear an opinion on the merging vs. overwriting of objects from @ayj who's much more of a k8s expert than I am. TBH I'm not sure what negative effects just overwriting could have in practice.

rcernich:
I haven't experienced any issues with patching objects, at least not in maistra.  That said, there are some resources that can be patched, and I've since added delete/create if the update fails (basically force apply).  There may also have been some bugs in the patch logic, e.g. if strategic merge fails, it should revert to standard json merge, and finally delete/create as a last resort.  The idea behind the patching logic was to not have to write customized merge logic for every resource type.

We haven't noticed any performance issues with the update logic though.  Our issues have been with the prune logic. (This was highlighted by bug when deleting the namespace containing the CR, as the deletion timestamp was being updated every second, which was causing a conflict error when going to remove the finalizer.  The initial delay was 15s and if the operator took longer than that, you'd hit this issue, where the namespace controller began continuously updating the deletion timestamp on all remaining custom resources.)

pkg/helmreconciler/rendering.go
```diff
@@ -38,6 +37,14 @@ import (
...
 	binversion "istio.io/operator/version"
 	"istio.io/pkg/log"
+	"istio.io/pkg/version"
+)
+
+var (
+	// objectCache holds the latest copy of each object applied by the controller, keyed by its Hash() function.
+	// Must be locked since ProcessManifest can be run concurrently.
+	objectCache   = make(map[string]*object.K8sObject)
+	objectCacheMu sync.RWMutex
```
ayj: How does this global cache work with multi-tenant controllers? I think it needs to be key'd by ICP (namespace). Alternately, the cache is attached to the reconciler itself which seems to be per-ICP instance already. The latter option would be preferred.

pkg/helmreconciler/reconciler.go
```diff
@@ -223,6 +213,23 @@ func (h *HelmReconciler) Delete() error {
 	return err
 }
 
+// allObjectHashes returns a map with object hashes in mm as the keys.
```
ayj: > in mm as the keys

I don't know what 'mm' means without reading the function body. It looks like its collecting the hashes of all objects across all manifests.

pkg/helmreconciler/pruning.go
ayj: Are non-CRD cluster scoped resources shared between ICP resources? Despite the name, I think our ClusterRole/ClusterRoleBindings are instantiated per-namespace (the namespace is encoded in the name). Removing them might be okay, but it does grant the operator broader cluster-wide privileges vs. adding/removing out of band.

pkg/helmreconciler/rendering.go
rcernich: Yes, it should be keyed by ICP resource.  For multi-tenancy, it's possible a single controller will be managing all installs (i.e. watching all namespaces).

pkg/helmreconciler/pruning.go
rcernich: The namespaced counter parts are Role and RoleBinding.  Also for multi-tenancy, where user namespaces opt-in to the mesh, we (maistra) still use ClusterRole resources as it simplifies management of the RoleBinding resources that get copied into the member namespaces.

That said, while they are not shared, we should be cleaning them up.  (FWIW, I don't consider CRDs to be part of a control plane install, but rather a bootstrapping operation done by the operator, as those are shared/required by all installations.)

pkg/helmreconciler/rendering.go
ostromart: Even with multi-tenant, would we expect to see overlapping objects coming from different ICPs? That seems like it shouldn't be allowed. Even with no caching it would lead to non-deterministic behavior.

pkg/helmreconciler/reconciler.go
ostromart: That was out of date - fixed.

ostromart:
@ayj yes they are shared. They are currently installed but not removed by the controller. I think that's a reasonable tradeoff because we don't have and out of band mechanisms in all scenarios we want to support.

pkg/helmreconciler/rendering.go
ayj: Oh, I see namespace is part of the key and presumably there is only one ICP install per namespace.

ayj (commented):
General question about this approach, would it be better to use the client-go informer+store than rolling our own?

ostromart:
Yes @rcernich suggested the same above and if we could make it work it would be a good solution. I'm not sure what it would take to convince the client to cache writes. I suspect it will be quite difficult to update the cache in such a way that the write to the server doesn't take place every time.

pkg/helmreconciler/rendering.go
ostromart: Right.

rcernich:
My understanding is that the informer will keep things up to date, which will save calls to the api server on the list calls, which is where the performance improvement comes from.

ostromart:
I think the informer will solve part of the problem but my gut feeling is that the bulk of the work is not the list calls but rather the writes for every object and I think it will be really hard to convince the informer not to do those. 
But if you (Rob) or someone else can commit to resolving the performance issue through whatever means by Jan 11th then I'm ok with pausing this PR to wait and see how that works out. If we can solve this with informer I will be very happy.

rcernich:
I'm not sure I follow, wrt the writes.  Are there more performance issues than just pruning?

I'm 100% okay with removing the generation annotation on the resources.  I believe this will cause only resources that were truly modified to be written during a reconcile.

For the prune operation, you'll always have to perform a list call on each resource type, as you don't know what wasn't created during an update.

Please correct me if I've misunderstood something.

ostromart:
Agree about the list operation and informer will help with that for sure. 
The writes are what I'm not sure about. First of all I'm not sure if writes are cached at all, I suspect not, so that would be the first thing to verify. Have you seen the informer doing write caches when an object is updated?

rcernich:
I'm not sure what you mean by the writes being cached, nor why that impacts performance.  Could you provide a concrete example?

Also could you confirm whether or not the poor performance is caused by the updates or the pruning?  As I mentioned previously, my experience is that pruning takes ~80% of the time used during an update, with no resources actually getting pruned (i.e. just time to query the apiserver).

Regardless, I'll see if we can't look at using typed resources instead of unstructured.  (Unstructured makes generic processing easier, but maybe something like RuntimeObject can replace it in the interfaces.)

sdake:
@ostromart the operator appears to be much faster (before it apprently was busted prior to 1.6 alpha2. Is this PR still relevant?

ostromart:
I'm not sure why this was still open since it was incorporated into istio/istio a long time ago. Closing.

PR #670, opened 2019-12-06
sync values for telemetry v2 update

richardwxn:
1. update sha
2. update values schema
3. update gen and goldens


richardwxn:
close as another PR to update the sha is merged

PR #671, opened 2019-12-06
upgrade cmd: using ICPS to compare changes between current install and target install.

### LINKED ISSUE #19135: upgrade uses only the global portion of config for upgrade
genOverlayICPS returns globalValues separately (btw. this is missing unvalidatedValues, which should be merged) and this is the only block which is used for ignoring diffs for upgrade warnings. 
Instead, the entire ICPS object should be used (for example if user overlays a k8s resource setting and the default changes, it should be ignored).

elfinhe:
Fix: https://github.com/istio/istio/issues/19135

Before this change, we used the values in Values API from the configmap of istio-sidecar-injector and the generated ICPS Values section, as the source of old and new installation configurations, which can cause some problems described in https://github.com/istio/istio/issues/19135.

This PR switches the comparison from Values-based to ICPS-based.

------
Related: https://github.com/istio/istio/issues/19451
I also get the kubectl outputs back at cmd/mesh/manifest-common.go:L85, which was removed by mistake, I think.

### DIFF
--- a/cmd/mesh/upgrade.go
+++ b/cmd/mesh/upgrade.go
@@ -181,8 +181,8 @@ func upgrade(rootArgs *rootArgs, args *upgradeArgs, l *Logger) (err error) {
 
 	// Run pre-upgrade hooks
 	hparams := &hooks.HookCommonParams{
-		SourceVer:    currentVersion,
-		TargetVer:    targetVersion,
+		SourceVer:  currentVersion,
+		TargetVer:  targetVersion,
 		SourceICPS: targetICPS,
 		TargetICPS: targetICPS,
 	}
--- a/cmd/mesh/upgrade.go
+++ b/cmd/mesh/upgrade.go
@@ -21,9 +21,6 @@ import (
 	"strings"
 	"time"
 
-	"istio.io/operator/pkg/apis/istio/v1alpha2"
-	"istio.io/operator/pkg/tpath"
-
 	"github.com/ghodss/yaml"
 	goversion "github.com/hashicorp/go-version"
 	"github.com/spf13/cobra"
@@ -31,6 +28,7 @@ import (
 	"istio.io/operator/pkg/compare"
 	"istio.io/operator/pkg/hooks"
 	"istio.io/operator/pkg/manifest"
+	"istio.io/operator/pkg/tpath"
 	opversion "istio.io/operator/version"
 	"istio.io/pkg/log"
 )
@@ -131,7 +129,7 @@ func upgrade(rootArgs *rootArgs, args *upgradeArgs, l *Logger) (err error) {
 	}
 
 	// Generate ICPS objects
-	_, targetICPS, err := genICPS(args.inFilename, "", "", args.force, l)
+	targetICPSYaml, targetICPS, err := genICPS(args.inFilename, "", "", "", args.force, l)
 	if err != nil {
 		return fmt.Errorf("failed to generate ICPS from file %s, error: %s", args.inFilename, err)
 	}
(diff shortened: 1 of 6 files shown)


nick4fake:
Relates to this one:
https://github.com/istio/istio/issues/19398

BTW, is there any ETA for merging? Is there anything I can help with?

elfinhe:
> Relates to this one:
> [istio/istio#19398](https://github.com/istio/istio/issues/19398)
> 
> BTW, is there any ETA for merging? Is there anything I can help with?

Hey Bohdan, can you elaborate why this is related to https://github.com/istio/istio/issues/19398?

nick4fake:
Because you can't use exiting IstioControlPlane API to control telemetry components (like istio) **AND** some parameters from Helm are not supported (I've provided values.kiali.tolerations as an example).
Setting them through unvalidatedValues also impossible as it is not used when merging manifest definitions on istioctl manifest apply. 
This PR doesn't address original issue completely, but at least provides a dirty way to specify Helm values.

I might be wrong though, checking this PR right now.

elfinhe:
> @elfinhe Because you can't use exiting IstioControlPlane API to control telemetry components (like istio) **AND** some parameters from Helm are not supported (I've provided values.kiali.tolerations as an example).
> Setting them through unvalidatedValues also impossible as it is not used when merging manifest definitions on istioctl manifest apply.
> This PR doesn't address original issue completely, but at least provides a dirty way to specify Helm values.
> 
> I might be wrong though, checking this PR right now.

@nick4fake thanks, this PR is addressing some upgrade-specific validation issue, i.e., choosing which part in ICPS to validate for upgrade. so I think it is not related. In #19398, I think Xinnan will do something to fix the helm values issue.

nick4fake:
@elfinhe Understood, I've created a separate issue
https://github.com/istio/istio/issues/19480

cmd/mesh/upgrade.go
```diff
@@ -239,11 +229,11 @@ func upgrade(rootArgs *rootArgs, args *upgradeArgs, l *Logger) (err error) {
...
-	diff := compare.YAMLCmpWithIgnore(curValues, tarValues, nil, ignoreValues)
+// checkUpgradeICPS checks the upgrade eligibility by comparing the current ICPS with the target ICPS
+func checkUpgradeICPS(curICPS, tarICPS, ignoreICPS string, l *Logger) {
+	diff := compare.YAMLCmpWithIgnore(curICPS, tarICPS, nil, ignoreICPS)
 	if diff == "" {
-		l.logAndPrintf("Upgrade check: Values unchanged. The target values are identical to the current values.\n")
+		l.logAndPrintf("Upgrade check: ICPS unchanged. The target values are identical to the current values.\n")
 	} else {
 		l.logAndPrintf("Upgrade check: Warning!!! The following values will be changed as part of upgrade. "+
 			"If you have not overridden these values, they will change in your cluster. Please double check they are correct:\n%s", diff)
```
ostromart: "If you have not overridden these values" doesn't seem to be applicable anymore because this is exactly what we're checking for.

cmd/mesh/upgrade.go
```diff
@@ -397,33 +357,16 @@ func identicalVersions(cv []manifest.ComponentVersion) bool {
 	return true
 }
 
-// genOverlayICPS reads an ICP from filename and returns an unmarshaled and validated ICPS from the spec field of ICP.
-// It separately returns a string which represents just the overlay values in the returned ICPS.
-func genOverlayICPS(filename string, force bool) (string, *v1alpha2.IstioControlPlaneSpec, error) {
+// genOverlayICPS reads an ICPS from filename and returns the YAML of it
+func genOverlayICPS(filename string) (string, error) {
```
ostromart: Is this function still necessary? We seem to be just reading a file and returning the contents, not generating anything.

elfinhe: Fixed.

cmd/mesh/upgrade.go
elfinhe: Fixed.

ostromart (approved):

PR #673, opened 2019-12-08
[release-1.4]sync with installer(telemetry v2 change)

richardwxn:


cmd/mesh/testdata/profile-dump/output/all_off.yaml
```diff
@@ -629,6 +629,14 @@ values:
     enabled: true
     v2:
       enabled: false
+      prometheus:
+        enabled: false
```
howardjohn: This needs to be true I think?

If I have the config today: `telemetry.v2.enabled=true` and upgraded to this, I would lose my telemetry.

howardjohn (changes requested):

cmd/mesh/testdata/profile-dump/output/all_off.yaml
richardwxn: @howardjohn yes you are right, for backward compatibility, it would be better to keep it as true.  Initially I am planning to add some note in the doc to say that we need to add another override with --set values.telemetry.v2.prometheus.enabled=true, but that might lead to confusion though. I would change it.

howardjohn: If we want to make this change it should be done in 1.5.x, we can't change things like this in a patch release. That being said, I don't think we should. We should have simple defaults. In 1.5 hopefully this means out of the box you get v2+prometheus, and don't need to configure anything at all

richardwxn: yep that makes sense

howardjohn (approved):

PR #677, opened 2019-12-10
Update the version of operator for 1.4.3

### LINKED ISSUE #19475: istioctl experimental upgrade does not work
**Bug description**
Cannot upgrade Istio from 1.4.0 to 1.4.1 via istioctl:

I downloaded new versio of istioctl and try to upgrade by running upgrade command:

$istioctl manifest versions

Operator version is 1.4.0.

The following installation package versions are recommended for use with this version of the operator:
  1.4.0

The following installation package versions are supported for upgrade by this version of the operator:
  >=1.3.3
   <1.6

$istioctl experimental upgrade -f ${overlay_file}
Client - istioctl version: 1.4.0

Error: the target version 1.4.1 is not supported by istioctl 1.4.0, please download istioctl 1.4.1 and run upgrade again

Please advise.

**Expected behavior**

istioctl version
client version: 1.4.1
control plane version: 1.4.0
data plane version: 1.4.0 (4 proxies), 1.3.2 (2 proxies)

**How was Istio installed?**
istioctl manifest apply -f ${overlay_file}

AWS

elfinhe:
Update the version of operator for 1.4.3.

Related: https://github.com/istio/istio/issues/18521
Fix: https://github.com/istio/istio/issues/19475


ostromart (approved):

ostromart:
We may as well add 1.4.3 at the same time. Alternatively, let's refactor the table so the operator version is a range rather than enumerating the patch versions if this is not too much work.

cmd/mesh/manifest-versions_test.go
```diff
@@ -113,9 +118,8 @@ func TestGetVersionCompatibleMap(t *testing.T) {
 				binVersion:  goVerNonexistent,
 				l:           l,
 			},
-			want: nil,
-			wantErr: fmt.Errorf("this operator version %s was not found in the version map",
-				goVerNonexistent.String()),
+			want:    latestCm,
+			wantErr: nil,
```
ostromart: There's a logic problem here. If we specify a version like 0.1, we don't want to return the latest version but rather return an error.

cmd/mesh/manifest-versions_test.go
```diff
@@ -53,14 +53,19 @@ func TestGetVersionCompatibleMap(t *testing.T) {
...
 			curCm = &vs[i]
 		}
 		if goVer133.Equal(vs[i].OperatorVersion) {
 			ver133Cm = &vs[i]
 		}
+		if latestCm == nil ||
+			(vs[i].OperatorVersion.GreaterThan(latestCm.OperatorVersion) &&
+				vs[i].OperatorVersion.LessThanOrEqual(binversion.OperatorBinaryGoVersion)) {
+			latestCm = &vs[i]
+		}
```
ostromart: It seems like we're copying logic here that should be exercised in the code being tested.

cmd/mesh/manifest-versions_test.go
elfinhe: fixed, thanks

cmd/mesh/manifest-versions_test.go
elfinhe: fixed.

PR #680, opened 2019-12-10
add tolerations to schema

richardwxn:
To address:https://github.com/istio/istio/issues/19480 and https://github.com/istio/istio/issues/19398. There are some components missing toleration parts in the values schema, because we still support user to set overlay through values pass through part, so we need to add them back.  But ideally k8s settings should be set under operator API instead


XanderStrike (approved):
LGTM for tolerations

Did you check any other fields or is this for tolerations alone?

pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -114,6 +114,8 @@ message CertManagerConfig {
   string podDnsPolicy = 10;
 
   TypeMapStringInterface podDnsConfig = 11;
+
+  TypeSliceOfMapStringInterface tolerations = 12 [deprecated=true];
```
ostromart: Can we use the k8s type here?

pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -1848,6 +1862,8 @@ message KialiConfig {
 
   // See PodAntiAffinityLabelSelector.
   TypeSliceOfMapStringInterface podAntiAffinityTermLabelSelector = 13 [deprecated=true];
+
+  TypeSliceOfMapStringInterface tolerations = 14 [deprecated=true];
```
ostromart: Is this actually supported in the kiali chart?

richardwxn: added another PR in installer to support that

pkg/apis/istio/v1alpha1/values_types.proto
richardwxn: Do you mean the IstioControlPlaneSpec? we use TypeSliceOfMapStringInterface for tolerations in values schema here

ostromart: We may as well use it here too. That's true for other fields that are currently untyped too. I've created https://github.com/istio/istio/issues/19709 to track this, you can skip it for this PR.

ostromart (approved):

PR #681, opened 2019-12-10
Profile set with flag should have highest priority

ostromart:


howardjohn (approved):
Confirmed this fixes https://github.com/istio/istio/issues/19503

richardwxn (approved):

PR #696, opened 2019-12-15
migrate logic and output update

richardwxn:
The change is mainly about

1. The existing migration logic would translate whatever it is and assume other parts, if not explicitly, are disabled. That's not correct, because values.yaml is used as overlay as well, so we should not implicitly set any value if not provided in values.yaml


howardjohn (commented):
should migrate just return the minimal required settings? for example if I set foo.bar in helm it creates an ICP with foo.bar only and not all the default as well

richardwxn:
@howardjohn yes that's the expected case and it should act as overlay to selected profile. I added some extra code in the PR now for testing how the combined output looks like only, will remove them later

howardjohn:
Is that how it is expected to behave today? If so I think its not working
quite right, I can file some issues

richardwxn:
@howardjohn there is an existing discussion already https://github.com/istio/istio/issues/19500, and this PR is created to resolve the concern.

cmd/mesh/manifest-migrate.go
```diff
@@ -112,6 +131,57 @@ func translateFunc(values []byte, l *Logger) error {
 	return nil
 }
 
+func genICPSFromProfile(profile string, ver string) (string, error) {
```
ostromart: This function looks duplicated?

cmd/mesh/manifest-migrate.go
```diff
@@ -142,6 +212,6 @@ func migrateFromClusterConfig(rootArgs *rootArgs, mmArgs *manifestMigrateArgs, l
 	if err != nil {
 		return fmt.Errorf("error marshaling untyped map to YAML: %s", err)
 	}
-
-	return translateFunc(res, l)
+	// no concept of profile when translating in-cluster configs
```
ostromart: why not? in cluster CR still has a profile field.

pkg/translate/translate_value.go
```diff
@@ -244,18 +247,6 @@ func (t *ReverseTranslator) setEnablementAndNamespacesFromValue(valueSpec map[st
 		}
 	}
 
-	// set namespace
```
ostromart: Is this moved somewhere? I'm not sure if we can just remove it.

richardwxn: it is done during the API mapping already

cmd/mesh/manifest-migrate.go
richardwxn: but I think the incluster values.yaml is a concatenation already? we can just translate from whatever it is

cmd/mesh/manifest-migrate.go
richardwxn: this extracts some part from the existing function and delete some parts that are not applicable here

cmd/mesh/manifest-migrate.go
ostromart: Per our discussion, let's make everything relative to default profile / built-in values defaults in charts. This should be simpler and less complex for users to understand.

cmd/mesh/manifest-migrate.go
ostromart: Are they close enough to restructure the code to reuse it? Normally a bit of duplication is not a big deal for simple functions but the logic in here is not obvious and it will be hard to understand in the future how these two versions differ.

richardwxn: I think we don't need this part of logic anymore based on our discussion.

cmd/mesh/testdata/manifest-migrate/output/values.yaml
```diff
@@ -16,11 +16,6 @@ spec:
               maxUnavailable: 25%
           tolerations: []
     enabled: true
-  cni:
```
ostromart: Looks like we only have one golden test file - could you please also add at least Alex's config from the bug to the golden test set?

richardwxn: I think this case is already covered in existing test input, for example there is no base component enablement  defined in input values.yaml thus no enablement for that defined in output ICPS.

richardwxn: and also the unit test covered it

ostromart: I know, but this would be more of a sanity check so that we can be sure a small input file produces a small output file.

richardwxn: added now

ostromart (approved):

PR #699, opened 2019-12-17
Determine default values based on Kubernetes env

howardjohn:
Partially implements https://github.com/istio/istio/issues/19182

This PR adds the ability for certain values to be configured based on
the currently Kubernetes environment. This should be perfectly reliable
for apply and operator, the only problems may arise are with generate.
For now, this just skips these custom modifications. In the future, if
needed, we can add a flag for generate to read from in cluster config as
well, although its a bit suspect because we may generate with one
cluster config and apply to another.

As an initial implementation, this adds support for detecting whether or
not trustworthy jwt tokens are enabled

### DIFF
--- a/cmd/mesh/profile-common.go
+++ b/cmd/mesh/profile-common.go
@@ -20,6 +20,8 @@ import (
 	"path/filepath"
 
 	"github.com/ghodss/yaml"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
 
 	"istio.io/operator/pkg/apis/istio/v1alpha2"
 	"istio.io/operator/pkg/helm"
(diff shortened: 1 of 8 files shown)


cmd/mesh/profile-common.go
```diff
@@ -119,6 +121,13 @@ func genICPS(inFilename, profile, setOverlayYAML, ver string, force bool, l *Log
 		}
 	}
 
+	if kubeConfig != nil {
+		baseYAML, err = util.OverlayYAML(baseYAML, getClusterSpecificValues(kubeConfig))
```
ostromart: Would be good to add some logging here.

cmd/mesh/profile-common.go
```diff
@@ -141,8 +150,30 @@ func genICPS(inFilename, profile, setOverlayYAML, ver string, force bool, l *Log
 	return finalYAML, finalICPS, nil
 }
 
-func genProfile(helmValues bool, inFilename, profile, setOverlayYAML, configPath string, force bool, l *Logger) (string, error) {
-	finalYAML, finalICPS, err := genICPS(inFilename, profile, setOverlayYAML, "", force, l)
+func getClusterSpecificValues(config *rest.Config) string {
```
ostromart: Optional: I suggest restructuring this function a bit:
- create a function to test for JWT support e.g. j, err := clusterSupportsThirdPartyJWT()
- use MakeTreeFromSetList([]string{fmt.Sprintf("values.global.sds.useTrustworthyJwt=%t", j)})

PR #707, opened 2019-12-19
Fallback to previous versions when failing to create translator

elfinhe:
Fallback to previous versions when failing to create translator and reversed translator, in case we may forget to update the version specific files (e.g., TranslateConfig and ReverseTranslators Map) when releasing a new minor version.

### DIFF
--- a/pkg/translate/translate.go
+++ b/pkg/translate/translate.go
@@ -45,9 +45,11 @@ const (
 	HelmValuesEnabledSubpath = "enabled"
 	// HelmValuesNamespaceSubpath is the subpath from the component root to the namespace parameter.
 	HelmValuesNamespaceSubpath = "namespace"
-
 	// devDbg generates lots of output useful in development.
 	devDbg = false
+	// maxFallbackNum is the max number of previous versions that NewTranslator()
+	// or NewReverseTranslator() will attempt by giving configs for that version.
+	maxFallbackNum = 1
 )
 
 var (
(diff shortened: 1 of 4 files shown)


pkg/translate/translate.go
```diff
@@ -109,10 +109,22 @@ type Translation struct {
...
 	f := "translateConfig/translateConfig-" + v + ".yaml"
 	b, err := vfs.ReadFile(f)
 	if err != nil {
+		if fallbackNum > 0 && minorVersion.Minor > 0 {
+			major := minorVersion.Major
+			minor := minorVersion.Minor - 1
+			previous := version.NewMinorVersion(major, minor)
+			log.Warnf("could not read translateConfig file for version %s, fallback to version %s",
+				minorVersion, previous)
+			return newTranslator(previous, fallbackNum-1)
```
richardwxn: I don't quite get the usage of fallback num here, so looks like it is hardcoded to 1 at the exported method, then in this recursion, next call would always have fallbacknum equals 0 so it has only two states. If that's the case why we still need to call newTranslator function again

elfinhe: I can rename the '1' to a constant `maxFallbackNum`. Sorry, I do not get your idea. What's your suggestion?

elfinhe: Fixed.

richardwxn (approved):

PR #708, opened 2019-12-19
Fix paths in proto docs

ostromart:


ostromart:
This is no longer relevant in the new API, closing.

PR #711, opened 2019-12-20
add service to k8s spec

richardwxn:
the service part was added to k8s proto under ISCP spec before. however it is not added to tranlationConfig so it is not doing anything if we add the settings under component.k8s.service


cmd/mesh/testdata/manifest-generate/output/ingressgateway_k8s_settings.yaml
```diff
@@ -0,0 +1,6225 @@
+# Resources for Base component
```
ostromart: Could you delete anything that's not being compared from the output? It looks like only Deployment and Service are tested. Just keep those to make the output readable.

cmd/mesh/testdata/manifest-generate/input/ingressgateway_k8s_settings.yaml
```diff
@@ -0,0 +1,27 @@
...
+  autoInjection:
+    enabled: false
+  gateways:
+    enabled: true
+    components:
+      ingressGateway:
+        enabled: true
+        k8s:
+          service:
+            externalTrafficPolicy: Local
```
ostromart: Some more complex example would be good. For example, adding to list of ports, which I'm actually not sure is going to work with this change...

cmd/mesh/testdata/manifest-generate/output/ingressgateway_k8s_settings.yaml
richardwxn: yep, removed

elfinhe (commented):
LGTM if the test mentioned by Martin is added.

cmd/mesh/testdata/manifest-generate/input/ingressgateway_k8s_settings.yaml
richardwxn: @ostromart I added the mirror of k8s service type so that unmarshalling can work with serviceport. PTAL

ostromart: lgtm, but could you hold off merging till https://github.com/istio/operator/pull/713 goes in? it will be easier to address any merge changes here rather than there.

richardwxn:
need to wait for the change in api repo: https://github.com/istio/api/pull/1229

ostromart:
@richardwxn please move this pr to istio/istio.

PR #712, opened 2019-12-20
Add warning when applying a different version of Istio.

### LINKED ISSUE #18487: istioctl manifest apply command should issue warning if the installed Istio version is different
It's quite possible that a user would have multiple versions of istioctl floating around - we want to prevent them running (say) istioctl manifest apply using the wrong version of istioctl, since this may lead to wrong charts being used.

elfinhe:
Fix: https://github.com/istio/istio/issues/18487

### DIFF
--- a/cmd/mesh/manifest-common.go
+++ b/cmd/mesh/manifest-common.go
@@ -55,6 +55,12 @@ func genApplyManifests(setOverlay []string, inFilename string, force bool, dryRu
 	if err != nil {
 		return fmt.Errorf("failed to generate manifest: %v", err)
 	}
+
+	if icps.Tag != version.OperatorBinaryGoVersion.String() {
+		l.logAndPrintf("Warning: istioctl version %s and the to-be-installed Istio version %s are different. ",
+			version.OperatorBinaryGoVersion.String(), icps.Tag)
+	}
+
 	opts := &kubectlcmd.Options{
 		DryRun:      dryRun,
 		Verbose:     verbose,


howardjohn (changes requested):
tag is not the same as version in all cases. how is version.OperatorBinaryGoVersion set? I couldn't find it

howardjohn (changes requested):
nevermind I found it, this definitely will not work in a large number of cases

elfinhe:
> nevermind I found it, this definitely will not work in a large number of cases

can we say it will work for the official released image in gcr.io and docker.io?

howardjohn:
Well I guess let me verify I understand this right. We have a manual updated hardcoded version (1.4.3 right now) and are comparing that to the tag of the image? or am I missing something?

elfinhe:
> Well I guess let me verify I understand this right. We have a manual updated hardcoded version (1.4.3 right now) and are comparing that to the tag of the image? or am I missing something?

Correct.

elfinhe:
Assumptions

1. the tag of ICPS will represent the image tag of Istio components, which is the version of that component.

1. version.OperatorBinaryGoVersion is the version of this isitoctl binary.

howardjohn:
We discussed at length how it's not feasible to do this. I didn't see
anything suggesting otherwise. Was there some reason we decided we could do
this?

For a direct example, go look at the 1.5.0-alpha.0 release. I'm pretty sure
the tag does not equal the version in the binary

PR #710, opened 2019-12-19
Update operator version to 1.5.0; support setting version by BuildInfo Tag.

elfinhe:
Fix: https://github.com/istio/istio/issues/19710

Fix: https://github.com/istio/istio/issues/19753

PR #716, opened 2019-12-23
fix invalid version format from docker build info.

### LINKED ISSUE #19753: [Operator] operator crash due to version panic
**Bug description**

The operator for the latest master failed to work because of version panic:
```
# out/linux_amd64/mesh manifest generate
panic: Malformed version: a3703b76cf4745f3d56bf653ed751509be116351-dirty

goroutine 1 [running]:
istio.io/operator/version.init.0()
	istio.io/operator@/version/version.go:46 +0x199
```

[ ] Docs
[ X ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Expected behavior**

**How was Istio installed?**

morvencao:
resolve: https://github.com/istio/istio/issues/19753

The version from the docker build info is not valid Semantic Version, that will cause version unmarshal panic.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -18,7 +18,6 @@ import (
 	goversion "github.com/hashicorp/go-version"
 
 	pkgversion "istio.io/operator/pkg/version"
-	buildversion "istio.io/pkg/version"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


howardjohn:
why is it set to a Sha? where did you download from?

version/version.go
```diff
@@ -36,11 +35,6 @@ var (
 func init() {
 	var err error
 	operatorVer := OperatorVersionString
-	// If dockerinfo has a tag (e.g., specified by LDFlags), we will use it as the version of operator
-	tag := buildversion.DockerInfo.Tag
-	if tag != "" && tag != "unknown" {
```
elfinhe: instead we can just check version format here for those dev images: 
```
	if tag != "" && buildversion.IsVersionString(tag) {
```

howardjohn: we should never get a SHA anyways afaik, how is this built?

howardjohn: nevermind, I see how it can happen

elfinhe:
This issue is also fixed in https://github.com/istio/operator/pull/710.

elfinhe (approved):

PR #717, opened 2019-12-23
Sync installer

howardjohn:


howardjohn:
blocking https://github.com/istio/istio/issues/19746

pkg/apis/istio/v1alpha1/values_types.proto
```diff
@@ -562,7 +562,14 @@ message GlobalConfig {
   // Specifies the Istio control plane’s telemetry Pod IP address or remote cluster DNS resolvable hostname
   string remoteTelemetryAddress = 49;
 
-  // The next available key is 54
+  IstiodConfig istiod = 54;
```
elfinhe: can we also add comments here?

howardjohn: done

elfinhe (approved):
LGTM. However, there is some relation between the enablements of istiod and pilot/galley (e.g., pilot is set to disabled but istiod is set to enabled). we need add more logic or docs to handle that.

howardjohn:
Lets go with the minimal approach for now and make it "smarter" later.

richardwxn (approved):

PR #724, opened 2019-12-26
mesh profile dump generates ICPS CR format output.

### LINKED ISSUE #19692: istioctl profile diff generate unexpected outputs
I was following this doc:
https://preliminary.istio.io/docs/setup/install/istioctl/#show-differences-in-profiles

But the commands gave me different outputs:
`istioctl profile diff <(istioctl profile dump default) <(istioctl profile dump demo)`
```
 taohe@taohe:~/istio.io/operator$ istioctl profile diff <(istioctl profile dump default) <(istioctl profile dump sds)
Difference of profiles are:
 autoInjection:
   components:
     injector:
       enabled: true
       k8s:
         replicaCount: 1
         strategy:
           rollingUpdate:
             maxSurge: 100%
             maxUnavailable: 25%
   enabled: true
 cni:
...
...
```

`istioctl profile diff default demo`

```
 taohe@taohe:~/istio.io/operator$ istioctl profile diff default demo
Difference of profiles are:
 kind: IstioControlPlane
 spec:
-  autoInjection:
-    components:
-      injector:
-        enabled: true
-        k8s:
-          replicaCount: 1
-          strategy:
-            rollingUpdate:
-              maxSurge: 100%
-              maxUnavailable: 25%
-    enabled: true
-  cni:
-    enabled: false
-  configManagement:
-    components:
-      galley:
-        enabled: true
...
...
```

### LINKED ISSUE #19804: mesh profile dump not generating ICPS CR format output
We should consider generating a ICPS which can be piped into `kubectl apply` cmd. `profile diff` should also get updated to support ICPS as input.

Current output of `mesh profile dump `
```
base:
  components:
    base:
      enabled: true
  enabled: true
...
```

Expected output
```yaml
kind: IstioControlPlane
spec:
  base:
    enabled: true
    components:
      base:
```

elfinhe:
Fix: https://github.com/istio/istio/issues/19804
Fix: https://github.com/istio/istio/issues/19692

### DIFF
--- a/cmd/mesh/manifest-migrate.go
+++ b/cmd/mesh/manifest-migrate.go
@@ -20,10 +20,8 @@ import (
 	"path/filepath"
 
 	"github.com/ghodss/yaml"
-	"github.com/gogo/protobuf/jsonpb"
 	"github.com/spf13/cobra"
 
-	"istio.io/operator/pkg/apis/istio/v1alpha2"
 	"istio.io/operator/pkg/kubectlcmd"
 	"istio.io/operator/pkg/translate"
 	"istio.io/operator/pkg/util"
@@ -96,26 +94,12 @@ func translateFunc(values []byte, l *Logger) error {
 		return fmt.Errorf("error translating values.yaml: %s", err)
 	}
 
-	translatedICPS := &v1alpha2.IstioControlPlaneSpec{}
-	err = util.UnmarshalWithJSONPB(translatedYAML, translatedICPS)
+	isCPYaml, err := icpsToIcp(translatedYAML)
 	if err != nil {
 		return err
 	}
 
-	isCP := &v1alpha2.IstioControlPlane{Spec: translatedICPS, Kind: "IstioControlPlane", ApiVersion: "install.istio.io/v1alpha2"}
-
-	ms := jsonpb.Marshaler{}
-	gotString, err := ms.MarshalToString(isCP)
-	if err != nil {
-		return fmt.Errorf("error marshaling translated IstioControlPlane: %s", err)
-	}
-
-	isCPYaml, err := yaml.JSONToYAML([]byte(gotString))
-	if err != nil {
-		return fmt.Errorf("error converting JSON: %s\n%s", gotString, err)
-	}
-
-	l.print(string(isCPYaml) + "\n")
+	l.print(isCPYaml + "\n")
 	return nil
 }
 
--- a/cmd/mesh/profile-common.go
+++ b/cmd/mesh/profile-common.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 
 	"github.com/ghodss/yaml"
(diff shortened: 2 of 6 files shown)


richardwxn (commented):
I remembered we had a discussion about this before? I think one problem is that if someone want to check the profile and apply then he can just simply run manifest apply <profile_name> right? Also it makes sense that dump only generates the config part value. Maybe add an extra option to choose to dump to ICPS CR format instead of changing the default now if we believe this is a UX improvement.

howardjohn:
> Also it makes sense that dump only generates the config part value

Why? I never understood why that would be useful

elfinhe:
> I remembered we had a discussion about this before? I think one problem is that if someone want to check the profile and apply then he can just simply run manifest apply <profile_name> right? Also it makes sense that dump only generates the config part value. Maybe add an extra option to choose to dump to ICPS CR format instead of changing the default now if we believe this is a UX improvement.

@richardwxn I have the same question as John. The specs only format is for internal use.

ostromart:
Use of config-path flag will lead to invalid ICPs. That was one reason for not dumping the header. You'd have to ensure at least that you don't emit the header if that flag is used. For me this would seem an unintuitive and inconsistent UX.
Why is omitting the header a problem for users?

howardjohn:
Because if you omit the header I cannot actually use the output without going copying the 1000s of lines of output, opening a text editor, trying to remember the header, indent all the output, and now maybe I can use it. I think in general the input format and output format should be the same as a general rule for tools, otherwise things get confusing/complicated

howardjohn:
Although I didn't know about the config-path flag. Not sure how to reconcile that

elfinhe:
> Because if you omit the header I cannot actually use the output without going copying the 1000s of lines of output, opening a text editor, trying to remember the header, indent all the output, and now maybe I can use it. I think in general the input format and output format should be the same as a general rule for tools, otherwise things get confusing/complicated

+1, now there are 2 versions of `istioctl profile`, e.g, https://github.com/istio/istio/issues/19692:

1.  `istioctl profile diff default demo` has header;
1. `istioctl profile dump` has no header;

for a complete ICPS file (i.e., dump/diff), we should print the header by default, which can be easily piped to `kubectl -f -` to create a CR.

as for --config-path flag and --set flag, the `spec` header can be omitted as input. Not sure if we need to output `spec` in this case.

ostromart:
> Because if you omit the header I cannot actually use the output without going copying the 1000s of lines of output, opening a text editor...

It should be easy to prepend a header - is this just an indentation issue? 
If so, I propose adjusting the indentation in the output so you can just paste blocks directly without re-indenting. 
For --config-path indentation to be correct you can take the length of the path and use that to get the add the correct indentation.

ostromart:
ping @howardjohn, is fixing the indentation acceptable here? see my comment above.

PR #713, opened 2019-12-21
Migrate code to use IstioOperator API in istio/api

ostromart:
See the corresponding API change:
https://github.com/istio/api/pull/1223

PR #727, opened 2019-12-30
Check enablement from values part as well

### LINKED ISSUE #19799: Gateway component not enabled by .values
`--set values.gateways.istio-egressgateway.enabled=true` does nothing -- egress is still disabled. `--set values.gateways.istio-egressgateway.enabled=true --set gateways.components.egressGateway.enabled=true` must be set, which is breaking the helm passthrough API

richardwxn:
should help resolve https://github.com/istio/istio/issues/19799

### DIFF
--- a/pkg/name/name.go
+++ b/pkg/name/name.go
@@ -56,6 +56,8 @@ const (
 	// Operator components
 	IstioOperatorComponentName      ComponentName = "IstioOperator"
 	IstioOperatorCustomResourceName ComponentName = "IstioOperatorCustomResource"
+
+
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/name/name.go
```diff
@@ -103,6 +103,22 @@ var (
 		KialiComponentName:              ThirdPartyFeatureName,
 		TracingComponentName:            ThirdPartyFeatureName,
 	}
+	// ComponentNameToHelmComponentPath defines mapping from component name to helm component root path.
```
ostromart: should be fairly easy to create this programmatically, maybe when you create a translator/reverse translator?

richardwxn: The problem is that the checkenablement part does not depend on translator so I don’t want to get this mapping from the translator, not sure whether there is better way possibly?

ostromart: It probably should depend on it, or actually should go into Translator. Name is a very low level pkg and shouldn't know anything about things like helm so anyway this function probably doesn't belong there now.

ostromart:
Let's hold off merging this till https://github.com/istio/operator/pull/713 goes in - I'm concerned I may break something in this change if I merge into that PR.

pkg/name/name.go
richardwxn: @ostromart translation is the process to convert between operator spec and helm values spec, however to get enablement values or some other paths from one of them does not reply on the other, right? I think it is ok to put them in names now as it does not need a process of translation. I add a TODO to refactor this part later but hope to get this merged soon to fix related bugs first.

ostromart:
@richardwxn please move this over to istio/istio since istio/operator has now been merged there.

PR #728, opened 2019-12-31
fix ICP list unmarshal issue.

morvencao:
Add proto tag for standard list meta, to align with k8s API, eg: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3538-L3542


howardjohn:
Can you link an issue or describe what is fixed or add a test?

elfinhe:
@morvencao

elfinhe (commented):
LGTM. But I think this will be reverted in IstioOperatorList PR#713. Let's hold for a while and apply this change to IstioOperatorList.

PR #735, opened 2020-01-08
add issue link to the ObjectMeta marshalling.

### LINKED ISSUE #19735: `istioctl manifest migrate|apply|generate` command always prints `proto: tag has too few fields: "-"` error
**Bug description**

[X] Installation

**Expected behavior**

`proto: tag has too few fields: "-"` should not be printed.

#### 1. Prepare values.yaml (any values can produce the error)

```
$ cat <<EOF > values.yaml
gateways:
  enabled: true
EOF
```

#### 2. Run `istioctl manifest migrate`

```
$ istioctl manifest migrate values.yaml
proto: tag has too few fields: "-"      # <= HERE
kind: IstioControlPlane
  ... snip ...
```

```
$ istioctl version --remote=false --short=false
version.BuildInfo{Version:"1.4.2", GitRevision:"35eb9dc7c6e78dac5bd8c3d142bc2a4601616932-dirty", GolangVersion:"go1.13.4", BuildStatus:"Modified", GitTag:"1.4.2"}
```

The error is caused when this line is called.

https://github.com/istio/operator/blob/f623b65d63f2b3fa68efcea9e89090b4b3fb65b3/cmd/mesh/manifest-migrate.go#L108

And ObjectMeta here causes the error.

https://github.com/istio/operator/blob/f623b65d63f2b3fa68efcea9e89090b4b3fb65b3/pkg/apis/istio/v1alpha2/istiocontrolplane_types.proto#L175

morvencao:
resolve: https://github.com/istio/istio/issues/19735

### DIFF
--- a/cmd/mesh/manifest-migrate.go
+++ b/cmd/mesh/manifest-migrate.go
@@ -106,6 +106,7 @@ func translateFunc(values []byte, l *Logger) error {
 
 	ms := jsonpb.Marshaler{}
 	gotString, err := ms.MarshalToString(isCP)
+	l.print("there is a known issue about the proto tag above, check https://github.com/istio/istio/issues/19735 for more details.\n\n")
 	if err != nil {
 		return fmt.Errorf("error marshaling translated IstioControlPlane: %s", err)
 	}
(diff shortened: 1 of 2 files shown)


richardwxn (approved):

howardjohn:
logging a bug is weird.. if we really need to do something is there no way to filter the message out? or just do nothing

ostromart:
it's not usual but not unprecedented either (it's fairly common practice in google3 logs for example). i think this is actually the least heavy-handed solution - filtering the message leads to weird code. 
it's one of these situations where every solution has a drawback...

PR #737, opened 2020-01-08
remove file watcher for template renderer

richardwxn:
the recursive file watcher is not necessary


howardjohn (approved):

romil-punetha:
When does this get released? I'm blocked on this. Or is there an alternative to this? The chart method is convenient since we can define values in isolated charts. One big values.yaml for the IstioControlPlane is a bit difficult to manage since I need to create too many entries in it.

PR #741, opened 2020-01-09
fix iop status update issue.

### LINKED ISSUE #20021: Operator reconciler always gets CR not found reconciling error
**Bug description**

Deploy the Istio operator from master and create IstioOperator resourcem, find that the following error `istiooperators.install.istio.io "example-istiocontrolplane" not found` in operator logs, and the reconciler will not stop because of the reconciling error:
```
2020-01-09T07:26:38.831807Z	info	end pruning
2020-01-09T07:26:38.834331Z	info	end reconciling resources
2020-01-09T07:26:38.834402Z	error	reconciling err: istiooperators.install.istio.io "example-istiocontrolplane" not found
2020-01-09T07:26:39.834785Z	info	Reconciling IstioOperator
```

[ ] Docs
[ X ] Installation
[ ] Networking
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[ ] User Experience
[ ] Developer Infrastructure

**Expected behavior**

**How was Istio installed?**

morvencao:
resolve: https://github.com/istio/istio/issues/20021

also fix `e2e-operator` test.

### DIFF
--- a/pkg/controller/istiocontrolplane/listeners.go
+++ b/pkg/controller/istiocontrolplane/listeners.go
@@ -80,7 +80,7 @@ func NewIstioStatusUpdater(instance *iop.IstioOperator) helmreconciler.Rendering
 }
 
 // EndReconcile updates the status field on the IstioOperator instance based on the resulting err parameter.
-func (u *IstioStatusUpdater) EndReconcile(_ runtime.Object, status map[string]*v1alpha1.IstioOperatorSpec_VersionStatus) error {
+func (u *IstioStatusUpdater) EndReconcile(_ runtime.Object, status *v1alpha1.InstallStatus) error {
 	iop := &iop.IstioOperator{}
 	namespacedName := types.NamespacedName{
 		Name:      u.instance.Name,
@@ -89,8 +89,8 @@ func (u *IstioStatusUpdater) EndReconcile(_ runtime.Object, status map[string]*v
 	if err := u.reconciler.GetClient().Get(context.TODO(), namespacedName, iop); err != nil {
 		return fmt.Errorf("failed to get IstioOperator before updating status due to %v", err)
 	}
-	iop.Spec.ComponentStatus = status
-	return u.reconciler.GetClient().Update(context.TODO(), iop)
+	iop.Status = status
+	return u.reconciler.GetClient().Status().Update(context.TODO(), iop)
 }
 
 // RegisterReconciler registers the HelmReconciler with this object
--- a/pkg/controller/istiocontrolplane/listeners.go
+++ b/pkg/controller/istiocontrolplane/listeners.go
@@ -90,7 +90,7 @@ func (u *IstioStatusUpdater) EndReconcile(_ runtime.Object, status map[string]*v
 		return fmt.Errorf("failed to get IstioOperator before updating status due to %v", err)
 	}
 	iop.Spec.ComponentStatus = status
-	return u.reconciler.GetClient().Status().Update(context.TODO(), iop)
+	return u.reconciler.GetClient().Update(context.TODO(), iop)
 }
 
 // RegisterReconciler registers the HelmReconciler with this object
--- a/pkg/helmreconciler/reconciler.go
+++ b/pkg/helmreconciler/reconciler.go
@@ -123,11 +123,11 @@ func (h *HelmReconciler) Reconcile() error {
 
 // processRecursive processes the given manifests in an order of dependencies defined in h. Dependencies are a tree,
 // where a child must wait for the parent to complete before starting.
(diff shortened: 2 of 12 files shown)


pkg/controller/istiocontrolplane/listeners.go
```diff
@@ -90,7 +90,7 @@ func (u *IstioStatusUpdater) EndReconcile(_ runtime.Object, status map[string]*v
 		return fmt.Errorf("failed to get IstioOperator before updating status due to %v", err)
 	}
 	iop.Spec.ComponentStatus = status
```
dgn: I think this line here is the issue. An operator should never update the `spec` of its input CRs

dgn (dismissed):
I know this is not caused by this PR but rather the CRD, but I don't think this is the way to go. Status should always go into the `Status` part of the CR, never into the spec

pkg/controller/istiocontrolplane/listeners.go
dgn: Reason being that you will run into conflicts much more often if you update the spec. In general, `spec` should only be updated by the user, `status` only by the operator. This way, conflicts are much less likely

morvencao: See my comments: https://github.com/istio/istio/issues/20021#issuecomment-572466483

morvencao: The PR https://github.com/istio/operator/pull/713 has move status to spec, need to get the reason that change.

dgn: Yes- sorry I didn't realize the new CRD was a complete rewrite. Will investigate this a bit. I think we should move the status out of the spec again, but that will now require a PR to istio/api

howardjohn:
> I know this is not caused by this PR but rather the CRD, but I don't think this is the way to go. Status should always go into the Status part of the CR, never into the spec

+1

pkg/controller/istiocontrolplane/listeners.go
irisdingbj: > The PR #713 moved status to spec. Moving status to spec is not a good idea, but I need to get the reasons for that change.

@ostromart @howardjohn Can we get a change in istio/api to get the status filed outside spec?

morvencao:
This also fixes `e2e_operator`test.

morvencao:
Raised a PR to move status fields out of IstioOperator spec: https://github.com/istio/api/pull/1231 @ostromart PTAL

morvencao:
Moved status out of spec:
```
# kubectl get iop -o yaml
apiVersion: v1
items:
- apiVersion: install.istio.io/v1alpha1
  kind: IstioOperator
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"install.istio.io/v1alpha1","kind":"IstioOperator","metadata":{"annotations":{},"name":"example-istiocontrolplane","namespace":"default"},"spec":{"profile":"demo"}}
    creationTimestamp: "2020-01-11T01:51:06Z"
    finalizers:
    - istio-finalizer.install.istio.io
    generation: 1
    namespace: default
    resourceVersion: "1540"
    selfLink: /apis/install.istio.io/v1alpha1/namespaces/default/istiooperators/example-istiocontrolplane
    uid: f52c3b52-53ea-43bd-8c87-bc84abfa4525
  spec:
    profile: demo
  status:
    component_status:
      Addon:
        status: 3
        status_string: HEALTHY
      Base:
        status: 3
        status_string: HEALTHY
      Citadel:
        status: 3
        status_string: HEALTHY
      EgressGateways:
        status: 3
        status_string: HEALTHY
      Galley:
        status: 3
        status_string: HEALTHY
      IngressGateways:
        status: 3
        status_string: HEALTHY
      Pilot:
        status: 3
        status_string: HEALTHY
      Policy:
        status: 3
        status_string: HEALTHY
      SidecarInjector:
        status: 3
        status_string: HEALTHY
      Telemetry:
        status: 3
        status_string: HEALTHY
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

pkg/controller/istiocontrolplane/listeners.go
morvencao: Done https://github.com/istio/api/pull/1231

pkg/helmreconciler/reconciler.go
```diff
@@ -189,6 +189,11 @@ func (h *HelmReconciler) processRecursive(manifests ChartManifestsMap) map[strin
 	}
 	wg.Wait()
 
+	out := &v1alpha1.InstallStatus{
+		//TODO: add overall status logic
```
elfinhe: maybe we need an issue to track this.

elfinhe (approved):
Generally LGTM. This unblocks `e2e-operator` test, I would let it merged first. @ostromart ptal if you have a second thought.

PR #745, opened 2020-01-10
Update Help String to be more informative

blhagadorn:
This helps the user better understand how to run `istioctl manifest apply` through the help section with three examples and link to documentation, issue here: https://github.com/istio/istio/issues/19999

### DIFF
--- a/cmd/mesh/root.go
+++ b/cmd/mesh/root.go
@@ -24,7 +24,10 @@ import (
 )
 
 const (
-	SetFlagHelpStr = `Can be used to override a path in the IstioOperator API spec, e.g. to choose a profile (--set profile=demo), enable or disable components (--set policy.enabled=true), or override Istio settings(--set values.global.mtls.enabled=true).  See documentation for more info: https://istio.io/docs/reference/config/istio.operator.v1alpha12.pb/#IstioControlPlaneSpec`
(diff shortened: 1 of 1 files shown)


cmd/mesh/root.go
```diff
@@ -24,9 +24,7 @@ import (
 )
 
 const (
-	SetFlagHelpStr = `Set a value in IstioOperator CustomResource. e.g. --set policy.enabled=true.
-Overrides the corresponding path value in the selected profile or passed through IstioOperator CR
-customization file`
+	SetFlagHelpStr = `Can be used to choose a profile (--set profile=demo), enable or disable components (--set policy.enabled=true), or override Istio settings(--set values.global.mtls.enabled=true).  See documentation for more info: https://istio.io/docs/reference/config/istio.operator.v1alpha12.pb/#IstioControlPlaneSpec`
```
elfinhe: Generally --set is used to override a path in IstioOperator API's spec, inlcluding profile, component, and values. We need one sentence to summarize that first.

blhagadorn: Good point, I've added some verbiage to reflect that in a single sentence before going into the examples

morvencao (commented):
/ok-to-test

ostromart:
istio/operator has been merged to istio/istio, could you please resubmit the PR there.

PR #747, opened 2020-01-10
Update the binary version to 1.4.4

elfinhe:
Update the binary version to 1.4.4

```
$ mesh version
version.BuildInfo{Version:"1.4.4", GitRevision:"unknown", GolangVersion:"go1.13.4", BuildStatus:"unknown", GitTag:"unknown"}
```


howardjohn (approved):
I will approve but re-iterate my stance that this is the wrong way to handle versioning

elfinhe:
> I will approve but re-iterate my stance that this is the wrong way to handle versioning

@howardjohn Now the binary will also read version from BuildInfo, if there is one it will override the hardcoded one.

PR #748, opened 2020-01-11
Package for translating IstioControlPlane to IstioOperator schemas

ostromart:


pkg/patch/patch.go
```diff
@@ -198,7 +198,7 @@ func applyPatches(base *object.K8sObject, patches []*v1alpha1.K8SObjectOverlay_P
 			errs = util.AppendErr(errs, err)
 			continue
 		}
-		errs = util.AppendErr(errs, tpath.WritePathContext(inc, p.Value))
+		errs = util.AppendErr(errs, tpath.WritePathContext(inc, p.Value, false))
```
elfinhe: why do we need merge==false?

ostromart:
Moved to https://github.com/istio/istio/pull/20182.

PR #749, opened 2020-01-12
only prune the managed resources.

### LINKED ISSUE #19815: Operator controller deletes user CRs cluster wide
Operator pruning is removing CRDs. This is REALLY bad - it deletes all user config cluster wide. See logs:

```
2019-12-27T17:01:05.512194Z     info    resource deleted: apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition
2019-12-27T17:01:05.514779Z     info    resource deleted: apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition
2019-12-27T17:01:05.517824Z     info    resource deleted: apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition
2019-12-27T17:01:05.520963Z     info    resource deleted: apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition
```

I put this at P1 since controller is still alpha, but this is a release blocker for promotion to beta

morvencao:
resolve: https://github.com/istio/istio/issues/19815 and https://github.com/istio/istio/issues/19816

- make the pruning list be dynamic map instead of static slice
- add the created/updated resources to the pruning list
- only prune the resource in the dynamic prune list
- don't clean the istio CRDs


elfinhe:
Nice!

pkg/controller/istiocontrolplane/pruningdetails.go
```diff
@@ -70,62 +70,62 @@ var (
 	}
 
 	// ordered by which types should be deleted, first to last
```
elfinhe: also update the comment of how to use the boolean? like `false` means xxx?

morvencao: done

pkg/helmreconciler/rendering.go
```diff
@@ -220,13 +220,14 @@ func (h *HelmReconciler) ProcessObject(chartName string, obj *unstructured.Unstr
 	gvk := mutatedObj.GetObjectKind().GroupVersionKind()
 	receiver.SetGroupVersionKind(gvk)
 	objectKey, _ := client.ObjectKeyFromObject(mutatedObj)
-	namespacedResourceMap, nonNamespacedResourceMap := h.customizer.PruningDetails().GetResourceTypes()
-	if val, ok := namespacedResourceMap[gvk]; !val && ok {
+	namespacedResourceMap, nonNamespacedResourceMap, pruningDetailsMU := h.customizer.PruningDetails().GetResourceTypes()
+	pruningDetailsMU.Lock()
```
irisdingbj: We only need to update the map when h.needUpdateAndPrune == true.

morvencao: Thanks, done.

irisdingbj (approved):

rcernich:
You should verify that this works when the operator is restarted between install and delete.  I suspect this will result in nothing being pruned, as the new instance of the operator won't have performed an install, thus it won't think there are any resource types that need pruning.

@ostromart FYI

PR #753, opened 2020-01-15
Fix addition and removal of finalizers

luksa:
Conflict errors during the addition or removal of finalizers were treated and logged as errors. Running into a conflict is a perfectly normal event - controllers should expect this and handle it appropriately (which this controller does). Given how this is not an error, it shouldn't be logged as one. 

Also, if the controller is unable to clean up properly, it should retry the cleanup instead of giving up and removing the finalizer (and leaving the owned resources in place).


rcernich:
Hey @luksa, the operator has been moved to istio/istio.  Could you please retarget this to that repo?

luksa:
Thanks @rcernich. Moving to istio/istio.

PR #759, opened 2020-02-06
Meshpolicy is a clusterwide resource

btoonk:
Meshpolicy is a clusterwide resource and therefor it does not have a namespace.


morvencao (commented):
Can you move your PR to https://github.com/istio/istio/tree/master/operator ? this repo will sync from that repo.

btoonk:
Will do

PR #768, opened 2020-03-06
Manual backport to remove possibility of data loss

sdake:
Fixes: https://github.com/istio/istio/issues/19815

### DIFF
--- a/pkg/controller/istiocontrolplane/pruningdetails.go
+++ b/pkg/controller/istiocontrolplane/pruningdetails.go
@@ -99,7 +99,8 @@ var (
 		{Group: "rbac.authorization.k8s.io", Version: "v1", Kind: "ClusterRole"},
 		{Group: "rbac.authorization.k8s.io", Version: "v1", Kind: "ClusterRoleBinding"},
 		{Group: "authentication.istio.io", Version: "v1alpha1", Kind: "MeshPolicy"},
-		{Group: "apiextensions.k8s.io", Version: "v1beta1", Kind: "CustomResourceDefinition"},
+		// Cannot currently prune CRDs because this will also wipe out user config.
+		// {Group: "apiextensions.k8s.io", Version: "v1beta1", Kind: "CustomResourceDefinition"},
 	}
 )
 


linsun:
@ostromart can you review/approve?

linsun (approved):

linsun:
cc @ericvn @howardjohn @jwendell we need this in 1.4 else 1.4's operator is broken when user moves to 1.5 - details - https://github.com/istio/istio/issues/21827

linsun:
contingent approval pending testing of this.

ostromart:
Approved, though have we actually verified this happens? What actually should happen is that when operator manifest is updated with new image tag, the controller is simply replaced with the new image which will ignore ICP. The only thing that should trigger deletion is if the ICP CR or CRD is removed, which would not happen if 1.5 operator manifest is kubectl applied or 1.5 is installed.
However - this is theory so if you're actually observing deletion or believe there's a risk this could happen I have no objection to backporting this.

linsun:
@ostromart see details in istio/istio#21827.  @kramvan1 has tested this pretty thoroughly.  If he doesn't delete the ICP resource, old deployments don't get cleaned up.  If he did the ICP resource, then CRDs are removed.  He is basically stuck and can't update to 1.5 using controller side operator.

kramvan1 (approved):
lgtm

sdake:
@ostromart Istio has had a no dats loss policy since at least 1.0. This was triggered by the mess that the special helm annotation caused deletion of CRs during upgrade. As you may know CRs are GCed when a CRD is deleted. The case you mentioned (deletion of the ICP resource) happens with some operator management frameworks (such as OLM).

I have not yet had a chance to properly test this. I should have placed a hold on this PR until I can test this actually implements the desired behavior of leaking all CRDs during an uninstall reinstall upgrade scenario.

sdake:
I have verified the CRDs leak with this PR as they should.

ostromart:
@sdake @linsun sgtm, please go ahead with this.

howardjohn:
meta comment: as a release manage I really don't want to be dealing with fixes for experimental features when we are ~4 months left. I don't mind here since its a trivial fix, but I strongly encourage you too push to move features you depend on out of experimental as quickly as possible moving forward

howardjohn (approved):

sdake:
@howardjohn The goal here is to provide anyone using this experimental feature an upgrade path. I agree we, being Environments, need to move the operator forward to a non-experimental stage.

Any time we, being Istio, have data loss in a maintenance release (and this can only really happen AFIAK with CRDs that are deleted) - we should fix them. We spent all kinds of time in environments avoiding data loss problems from 1.1->1.3, which was one of the reasons for the desire to remove the original helm charts. We did end up finding a suitable workaround for individuals on Istio 1.3, but it was a very specific solution.

Data loss, even in an experimental feature, can be deadly.

Cheers
-steve

PR #776, opened 2020-04-22
Fixing the multicluster gateways example

josephpeacock:
Ref: https://github.com/istio/istio/issues/23155

Looking around this repo, I also see a 
coreDNS entry in the translateConfig, so not sure if that's what isn't working as expected?

https://github.com/istio/operator/blob/master/data/translateConfig/translateConfig-1.5.yaml#L117


morvencao:
@josephpeacock This repo doesn't accept new PR, please raise PR to https://github.com/istio/istio

PR #779, opened 2020-07-13
Use "allowlist" instead of "whitelist"

tbarrella:
As part of https://github.com/istio/istio/issues/25381


dgn (approved):
I'm okay with merging this, but please note that this repository is only used in Istio 1.4.x versions and is going to be archived soon

PR #780, opened 2020-09-02
add in kustomization in root directory to allow for ease of use with …

Kampe:
…kustomize remote base

This allows people to utilize this repository as a base and patch overlays on top.

ie 
```
Bases:
- github.com/istio/operator
```

### DIFF
--- a/kustomization.yaml
+++ b/kustomization.yaml
@@ -1,0 +1,4 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- deploy/


ostromart:
This repo is inactive and out of date - see istio/istio/manifests instead.

Kampe:
Interesting, thats sad because I don't think this is useful all in one place - kustomize remote base assumes the root dir, of which in the istio proper repository, I don't believe they're even present there anymore and unfortunately this would live in the root of the repo and there's no logic there around the operator aside from mention in the .MD

Kampe:
Also the operator no longer provides manifests it seems unfortunately, so this is no longer possible to achieve either.

howardjohn:
Does kustomize require a root dir? I have used `kustomize build "
https://github.com/kubernetes-sigs/service-apis/config/crd` which isn't the
root unless I am understanding incorrectly.

Kampe:
You can reference resources by the raw files in a kustomization.yaml - however I'm pretty sure the remote base kustomization requires a kustomization to live in the root, and point to either resources itself, or another kustomization.yaml in a different directory in which points to the resources.  

Edit: seems your are right and you can specify a URI (invalid to github) and it'll traverse the tree to the kustomize assets, do wonder if this is a feature or a bug.

